PR #27, opened 2017-05-17
Adding Travis config [triage-skip]

hiranya911:


hiranya911:
This was implemented as part of #29

PR #29, opened 2017-05-18
Running Unit Tests without gcloud SDK in Automated Environments

hiranya911:
The unit test framework requires a local installation of the gcloud SDK, and also to authorize it by running `gcloud auth application-default login`. This is an interactive process which cannot be fully automated. As a result we cannot run it in an automated environment like TravisCI. 

This patch updates application default credential tests to run by setting the `GOOGLE_APPLICATION_CREDENTIALS` environment variable. For a number of other auth tests, we introduce a new `MockCredential` (these tests simply require some non-cert credential).

### DIFF
--- a/test/unit/firebase.spec.ts
+++ b/test/unit/firebase.spec.ts
@@ -144,7 +144,7 @@ describe('Firebase', () => {
       });
 
       return firebaseAdmin.app().INTERNAL.getToken().then(token => {
-        if (credPath === 'undefined') {
+        if (typeof credPath === 'undefined') {
           delete process.env.GOOGLE_APPLICATION_CREDENTIALS;
         } else {
           process.env.GOOGLE_APPLICATION_CREDENTIALS = credPath;
--- a/test/unit/firebase.spec.ts
+++ b/test/unit/firebase.spec.ts
@@ -136,12 +136,21 @@ describe('Firebase', () => {
     });
 
     it('should initialize SDK given an application default credential', () => {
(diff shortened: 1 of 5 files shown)


hiranya911:
And now it works on Travis as seen in the PR check results above :)

test/unit/firebase.spec.ts
```diff
@@ -136,12 +136,21 @@ describe('Firebase', () => {
...
+      credPath = process.env.GOOGLE_APPLICATION_CREDENTIALS;
+      process.env.GOOGLE_APPLICATION_CREDENTIALS = path.resolve(__dirname, '../resources/mock.key.json');
       firebaseAdmin.initializeApp({
         credential: firebaseAdmin.credential.applicationDefault(),
       });
 
-      return firebaseAdmin.app().INTERNAL.getToken()
-        .should.eventually.have.keys(['accessToken', 'expirationTime']);
+      return firebaseAdmin.app().INTERNAL.getToken().then(token => {
+        if (credPath === 'undefined') {
```
jshcrowthe: Is `credPath` actually equal to the string `undefined`?

I'm imagining you meant either:

```javascript
if (typeof credPath === 'undefined') { ... }
```

**or**

```javascript
if (credPath === undefined) { ... }
```

hiranya911: Fixed by adding `typeof`. Thanks for pointing it out.

Issue #48, opened 2017-06-30
Integration Test Failure Caused by Database Rules


hiranya911:
Integration test suite may sometimes fail with the following error:

```
FIREBASE WARNING: set at /blocked failed: permission_denied 
[FAILURE] nonNullApp.database().ref("blocked").set(): Error: PERMISSION_DENIED: Permission denied
```

This happens when there are custom rules in the realtime database. Changing the rules back to defaults usually fixes the issue:

```
{
  "rules": {
    ".read": "auth != null",
    ".write": "auth != null",
  }
}
```

The integration test suite should not depend on a specific rules configuration. If such a dependency is needed, the test suite should update the rules config during the setup phase (Java and Python SDKs do this already).

hiranya911:
This has been implemented. Integration tests overwrite the database rules with the defaults now. If this is deemed too invasive, we can implement a warning+prompt mechanism in the future.

PR #51, opened 2017-07-07
Updating the DB rules to the defaults before running integration tests

hiranya911:
Integration test suite expects default security rules. Therefore it should update the rules accordingly before running tests to avoid false positives.

See #48

### DIFF
--- a/test/integration/index.js
+++ b/test/integration/index.js
@@ -26,6 +26,9 @@ var auth = require('./auth');
 var database = require('./database');
 var messaging = require('./messaging');
 
+var apiRequest = require('../../lib/utils/api-request');
+var url = require('url');
+
 var serviceAccount = utils.getCredential();
 var databaseURL = 'https://' + utils.getProjectId() + '.firebaseio.com';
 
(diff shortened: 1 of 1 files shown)


test/integration/index.js
```diff
@@ -75,8 +78,26 @@ utils.assert(
...
+  // Update database rules to the defaults. Rest of the test suite
+  // expects it.
+  const client = new apiRequest.SignedApiRequestHandler(defaultApp);
+  const dbUrl =  url.parse(defaultApp.options.databaseURL);
+  const defaultRules = {
+    rules : {
+      '.read': 'auth != null',
+      '.write': 'auth != null',
+    }
+  }
```
bojeil-google: Add semicolon.

test/integration/index.js
```diff
@@ -75,8 +78,26 @@ utils.assert(
...
+  const headers = {
+    'Content-Type': 'application/json',
+  }
```
bojeil-google: Add semicolon.

test/integration/index.js
```diff
@@ -75,8 +78,26 @@ utils.assert(
...
+function updateRules() {
+  // Update database rules to the defaults. Rest of the test suite
+  // expects it.
+  const client = new apiRequest.SignedApiRequestHandler(defaultApp);
+  const dbUrl =  url.parse(defaultApp.options.databaseURL);
+  const defaultRules = {
+    rules : {
+      '.read': 'auth != null',
+      '.write': 'auth != null',
+    }
```
bojeil-google: Add comma.

hiranya911: Done

test/integration/index.js
hiranya911: Done

test/integration/index.js
hiranya911: Done

test/integration/index.js
```diff
@@ -75,8 +78,26 @@ utils.assert(
...
+    rules : {
+      '.read': 'auth != null',
+      '.write': 'auth != null',
+    },
+  };
+  const headers = {
+    'Content-Type': 'application/json',
+  };
+  return client.sendRequest(dbUrl.host, 443, '/.settings/rules.json', 
+    'PUT', defaultRules, headers, 10000);
```
bojeil-google: A thought came to me. Wouldn't this effectively overwrite a developer's existing rules? If so, this could be problematic. Please confirm.

hiranya911: It does, but that is ok. Other SDKs do the same during integration tests. We also advise developers to use a separate, dedicated project for integration testing since the test suite modifies both the content and the rules of the database.

bojeil-google: Hmm, is it possible to prompt the developer with a yes or no question using something like https://nodejs.org/api/readline.html warning them of this issue before proceeding? A flag can be passed to bypass that if needed. What do you think?

PR #52, opened 2017-07-11
Update index.d.ts

Crazometer:
I'm fairly sure this is correct. Kind of hard to work out with the database.js minified.

Issue #53

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -203,7 +203,7 @@ declare namespace admin.database {
       successCallback?: (a: admin.database.DataSnapshot, b?: string) => any,
       failureCallbackOrContext?: Object|null,
       context?: Object|null
-    ): Promise<any>;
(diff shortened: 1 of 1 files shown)


Crazometer:
I signed it!

hiranya911:
This need to be fixed in https://github.com/firebase/firebase-js-sdk. The Admin SDK simply pulls the minified database code from the JS SDK repo. 

However, I believe the issue you're referring to has been fixed in that repo recently:

https://github.com/firebase/firebase-js-sdk/blob/00f9c37543b6250f601f524bb50408ccf1c56bd9/src/database/api/Query.ts#L241
https://github.com/firebase/firebase-js-sdk/commit/a8919cdd16f3a4811efcc8736d2f167955f54047

When we sync the database code again, this fix will make its way to the Admin SDK.

PR #58, opened 2017-07-18
Implements phone auth support for Admin node.js SDK.

bojeil-google:
Adds phoneNumber to UserRecord and UserInfo.
Implements new API for looking up users by phone number: getUserByPhoneNumber
Adds the ability to pass the phoneNumber property when creating a new user,
or updating an existing one via createUser and updateUser.
This includes the ability to delete the phone number for a specified user.
Defines the new related auth errors:
auth/invalid-phone-number and auth/phone-number-already-exists.

### DIFF
--- a/test/integration/auth.js
+++ b/test/integration/auth.js
@@ -82,7 +82,7 @@ function test(utils) {
     ];
     return Promise.all(promises)
       .catch(function(error) {
-        utils.logFailure('setUp()', error);
+        utils.logFailure('before()', error);
       });
   }
 
--- a/test/integration/auth.js
+++ b/test/integration/auth.js
@@ -33,23 +33,75 @@ function test(utils) {
 
   var newUserUid = utils.generateRandomString(20);
   var nonexistentUid = utils.generateRandomString(20);
+  var testPhoneNumber = '+11234567890';
+  var testPhoneNumber2 = '+16505550101';
+  var nonexistentPhoneNumber = '+18888888888';
+  var updatedEmail = utils.generateRandomString(20) + '@example.com';
+  var updatedPhone = '+16505550102';
 
   var mockUserData = {
     email: newUserUid + '@example.com',
     emailVerified: false,
+    phoneNumber: testPhoneNumber,
     password: 'password',
     displayName: 'Random User ' + newUserUid,
     photoURL: 'http://www.example.com/' + newUserUid + '/photo.png',
     disabled: false,
   };
 
+  /**
+   * Helper function that deletes the user with the specified phone number
+   * if it exists.
+   * @param {string} phoneNumber The phone number of the user to delete.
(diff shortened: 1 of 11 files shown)


test/unit/auth/auth-api-request.spec.ts
```diff
@@ -534,6 +564,75 @@ describe('FirebaseAuthRequestHandler', () => {
...
+  describe('getAccountInfoByPhoneNumber', () => {
+    const httpMethod = 'POST';
+    const host = 'www.googleapis.com';
+    const port = 443;
+    const path = '/identitytoolkit/v3/relyingparty/getAccountInfo';
+    const timeout = 10000;
+    it('should be fulfilled given a valid phoneNumber', () => {
+      const expectedResult = {
+        users : [
+          {phoneNumber: '+11234567890'},
```
hiranya911: Should we also add the `providerUserInfo` field to the expected response, and see if get parsed correctly?

hiranya911:
I believe we also need to update `index.d.ts` right?

test/unit/auth/auth-api-request.spec.ts
bojeil-google: I added a more realistic response and I test the expected response is returned from this operation which is enough for this API. This API returns a response that resolves with the same server response. UserRecord parses that content. I have tests for that in that file. auth.ts passes the response to UserRecord. I have a check for the whole flow in auth.spec.ts. I stub the response of getAccountInfoByPhoneNumber and then check that it is parsed correctly into the expected UserRecord.

bojeil-google:
Thanks for catching the missing definitions in index.d.ts. I added those. I also added detailed response for getAccountInfoByPhoneNumber. More detailed checks on parsing is adding for getUserByPhoneNumber.

hiranya911:
LGTM

test/integration/auth.js
```diff
@@ -33,23 +33,75 @@ function test(utils) {
...
+    // Delete any existing users that could affect the test outcome.
+    var promises = [
+      deletePhoneNumberUser(testPhoneNumber),
+      deletePhoneNumberUser(testPhoneNumber2),
+      deletePhoneNumberUser(nonexistentPhoneNumber),
+      deletePhoneNumberUser(updatedPhone)
+    ];
+    return Promise.all(promises)
+      .catch(function(error) {
+        utils.logFailure('setUp()', error);
```
hiranya911: Perhaps log this as `before()` to be clear?

hiranya911:
LGTM with one minor suggestion.

test/integration/auth.js
bojeil-google: Thanks for catching this mislabel. I originally had named this setUp() but then renamed it to before().

PR #61, opened 2017-07-25
Provides contributors with a prompt to choose whether to overwrite, s…

bojeil-google:
…kip or abort the rules overwrite necessary for running integration tests.

### DIFF
--- a/test/integration/index.js
+++ b/test/integration/index.js
@@ -16,6 +16,7 @@
 
 var _ = require('lodash');
 var chalk = require('chalk');
+var readline = require('readline');
 
 var admin = require('../../lib/index');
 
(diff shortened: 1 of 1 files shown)


bojeil-google:
Will add a flag to force overwrite for automated scripts.

test/integration/index.js
```diff
@@ -16,6 +16,7 @@
 
 var _ = require('lodash');
 var chalk = require('chalk');
+var readline = require('readline');
```
hiranya911: Should we add this to `package.json`?

test/integration/index.js
```diff
@@ -78,6 +79,58 @@ utils.assert(
...
+ * Prompts the developer whether the Database rules should be
+ * overwritten with the relevant rules for the integration tests.
+ * The developer has 3 options:
+ * yes/y to agree to the rules overwrite.
+ * skip to skip the overwrite (rules already manually configured) and continue
+ * with the tests.
+ * no/n or other to abort.
+ * @return {Promise} A promise that resolves when the rules change is processed.
+ */
+function promptForUpdateRules() {
```
hiranya911: Lets also support an `--overwrite yes|no|skip` flag as discussed offline. For example:

`node test/integration --overwrite yes` should simply skip the prompt, and proceed to overwrite the rules.

test/integration/index.js
bojeil-google: I believe this module comes with node.js: https://nodejs.org/api/readline.html, similar to http module.

test/integration/index.js
bojeil-google: Yes, I will do so in my next commit. Thanks.

bojeil-google:
You're welcome, @hiranya911 !

Issue #62, opened 2017-07-26
FR: Provide Stronger Type Checking for createUser() and updateUser() Methods in Auth


hiranya911:
These methods currently accept arbitrary collections of properties. It is better to support a more strongly typed signature. For example:

`UserRecord createUser(req?)`

where `req` is a type defined as follows:

```
export type CreateRequest {
  public uid: string
  public displayName: string
  public emailVerified: bool
  ...
}
```

hiranya911:
Fixed.

PR #64, opened 2017-07-27
Type definitions for createUser() and updateUser() parameters

hiranya911:
Implementing type definitions for arguments of `createUser()` and `updateUser()` methods in `auth`. This improvement was suggested for the Node SDK during a recent API review discussion. 

@bojeil-google Do we need to send this through the API review process?

See #62.

### DIFF
--- a/src/auth/user-record.ts
+++ b/src/auth/user-record.ts
@@ -35,6 +35,29 @@ function parseDate(time: any): string {
   return null;
 }
 
+/** Parameters for create user operation */
+export type CreateRequest = {
+  uid?: string;
+  displayName?: string;
+  email?: string;
+  emailVerified?: boolean;
+  phoneNumber?: string;
+  photoURL?: string;
+  disabled?: boolean;
+  password?: string;
+}
+
+/** Parameters for update user operation */
+export type UpdateRequest = {
+  displayName?: string;
+  email?: string;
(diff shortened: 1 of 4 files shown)


src/auth/user-record.ts
```diff
@@ -35,6 +35,29 @@ function parseDate(time: any): string {
   return null;
 }
 
+/** Parameters for create user operation */
+export type CreateRequest = {
```
bojeil-google: Wouldn't it be better to use interfaces:
export interface UpdateRequest {
  displayName?: string;
  email?: string;
  ...
}
export interface CreateRequest extends UpdateRequest {
   uid?: string;
}

hiranya911: Done

bojeil-google:
Hmm, I don't think we need to go through the API review, but we can quickly run it by someone on the council to confirm.

hiranya911:
Got the confirmation that API review process is not required for this change. It was suggested to use anonymous types if possible, but I believe named types make the code simpler (since each type has many fields), and it will surely make the documentation easier to navigate as well. The type names used here (`CreateRequest` and `UpdateRequest`) were already approved for the Java SDK.

hiranya911:
Reassigning to @bojeil-google for one more review, since `index.d.ts` was updated.

PR #79, opened 2017-08-31
Updating the build/test/lint scripts to use npm instead of gulp

hiranya911:
This PR changes how we lint, test and build the SDK.

* Linting: `npm run lint`
* Linting and unit tests: `npm test`
* Unit tests with code coverage: `npm run test:coverage`
* Integration tests: `npm run integration`

With this we will only use gulp when building release packages. Even in that case we can invoke it via npm as `npm run build`. The `createReleaseTarball.sh` script has been updated to use the new npm build targets.

There are couple of advantages to this move:
* The old gulp script was very complicated. This simplifies it down to bare bones.
* Stack traces of failed tests are impossible to comprehend when invoked via gulp (due to compilation and in-memory evaluation of gulp). With this we run mocha directly over TypeScript files, so the line numbers directly map to source code. For example:

```
1) Messaging sendToDevice() should be rejected given an array containing more than 1,000 registration tokens:
     Error: An unknown server error was returned.
      at FirebaseMessagingError.Error (native)
      at FirebaseMessagingError.FirebaseError [as constructor] (src/utils/error.ts:47:5)
      at new FirebaseMessagingError (src/utils/error.ts:177:5)
      at Function.FirebaseMessagingError.fromServerError (src/utils/error.ts:172:12)
      at src/messaging/messaging-api-request.ts:142:36
      at process._tickCallback (internal/process/next_tick.js:109:7)
```

* Unit tests no longer require an explicit build step (i.e. no need to copy compiled files into a lib directory). This speeds up the tests slightly. 

Huge thank you to @jshcrowthe who was instrumental in figuring out most of the things needed to make this work.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -104,37 +104,31 @@ follow the prompts:
 $ gcloud beta auth application-default login
 ```
 
-### Lint, Build, and Test
+### Running the Linter
 
 Source files are written in [TypeScript](https://www.typescriptlang.org/) and linted using
-[TSLint](https://palantir.github.io/tslint/). Tests are run on the compiled JavaScript files.
-
-The lint, build, and test process is kicked off via the default `gulp` task, although you can also
-run each task individually:
+[TSLint](https://palantir.github.io/tslint/). Run the following command to kick off the linter:
 
 ```bash
-$ gulp         # Lint, build, and test
-$ gulp lint    # Just lint
-$ gulp build   # Just build
-$ gulp test    # Just test
+$ npm run lint
 ```
 
 ### Running Tests
 
 There are two test suites: unit and integration. The unit test suite is intended to be run during
-development and the integration test suite is intended to be run before packaging up release
+development, and the integration test suite is intended to be run before packaging up release
 candidates.
 
 To run the unit test suite:
 
-```
-$ gulp   # Lint, build, and run unit test suite
+```bash
+$ npm test   # Lint and run unit test suite
 ```
 
-To run the integration test suite:
+If you wish to skip the linter, and only run the unit tests:
 
-```
-$ node test/integration   # Run integration test suite
+```bash
+$ npm run test:unit
 ```
 
 The integration test suite requires a service account JSON key file, and an API key for a Firebase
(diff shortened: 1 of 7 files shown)


justinrosenthal:
This seems to be working great for me, and I'm basing a new PR on it... but I feel like I shouldn't be the one to LGTM this.  Looking forward to seeing it merged, though.

gulpfile.js
```diff
@@ -122,54 +101,9 @@ gulp.task('copyTypings', function() {
...
-          }))
-          .pipe(istanbul.writeReports())
-          .on('finish', function() {
-            return del(paths.testBuild).then(exit);
-          });
-      });
-  })
-});
-
 // Re-runs the linter and regenerates js every time a source file changes
```
bojeil-google: This doesn't seem to run the linter anymore.

test/unit/auth/credential.spec.ts
```diff
@@ -256,7 +257,7 @@ describe('Credential', () => {
       const c = new RefreshTokenCredential(new RefreshToken(TEST_GCLOUD_CREDENTIALS));
       return c.getAccessToken().then((token) => {
         expect(token.access_token).to.be.a('string').and.to.not.be.empty;
-        expect(token.expires_in).to.equal(ONE_HOUR_IN_SECONDS);
+        expect(token.expires_in).to.greaterThan(FIVE_MINUTES_IN_SECONDS);
```
bojeil-google: I noticed this test is flaky. Why did you choose > 5mins?
Another option is:
.to.be.within(ONE_HOUR_IN_SECONDS - 1, ONE_HOUR_IN_SECONDS)

hiranya911: 5 minutes is the skew we take into consideration when scheduling [proactive token refresh](https://github.com/firebase/firebase-admin-node/blob/hkj-build-regime/src/firebase-app.ts#L130) events. Other than that the SDK can handle any reasonable expiry value sent by the server. I'd like to treat the default 3600s as an implementation detail of the token server . WDYT?

gulpfile.js
hiranya911: Updated comment.

CONTRIBUTING.md
```diff
@@ -173,6 +182,10 @@ Here are some highlights of the directory structure and notable source files
   * `resources/` - Provides mocks for several variables as well as mock service account keys.
 * `.github/` - Contribution instructions as well as issue and pull request templates.
 * `createReleaseTarball.sh` - Generates a new release tarball and ensures it passes all tests.
-* `gulpfile.js` - Defines the `gulp` tasks.
+* `verifyReleaseTarball.sh` - Installs and validates the release tarballs created by
+  `createReleaseTarball.sh`.
+* `gulpfile.js` - Defines the `gulp` tasks necessary for building release artifacts.
 * `tslint.json` - TypeScript linting rules.
 * `tsconfig.json` - TypeScript configuration options.
+* `tsconfig-lint.json` - TypeScript configuration options for the linter. This simple widens
```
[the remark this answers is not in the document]
hiranya911: Done

test/unit/auth/credential.spec.ts
bojeil-google: The 5 minute value is an arbitrary one that we have chosen. Don't you think the check should be more precise? I see the counter argument on that being a token server detail. Anyway, I don't mind either way.

Issue #80, opened 2017-09-01
Value of FirebaseAppError.code is unexpected


justinrosenthal:
In the `FirebaseAppError` constructor, it prepends the string `app/` to the error code before calling `super`.  This makes it very difficult to look at an instance of `FirebaseAppError` at runtime and determine which error code it has.

For example:
```js
let err = new FirebaseAppError(AppErrorCodes.NETWORK_ERROR, "ECONNRESET");
if (err.code === AppErrorCodes.NETWORK_ERROR) {
  // ...
} else {
  // Unexpectedly we end up here because err.code === ('app/' + AppErrorCodes.NETWORK_ERROR)
}
```

Is there a suggestion for how to deal with this? I couldn't find an example elsewhere in the codebase.

hiranya911:
Can't we rely on `FirebaseAppError` to always behave this way? If so we can test for equality with `'app/' + code` or at least check for `err.code.endsWith(code)`. I'm not sure why we do this error code prefixing business, but it seems to be the convention followed in the whole SDK (See `FirebaseAuthError` and `FirebaseMessagingError` for example).

justinrosenthal:
Indeed there are ways to check as long as you know the inner workings of `FirebaseAppError` (though you'd prefer not to litter the magic string `/app` all over the codebase), but for the uninitiated developer, it's not obvious why my code above doesn't work.

Before I went ahead and implemented a conditional other than `=== <code>` I wanted to ask what the maintainers suggest.

hiranya911:
I'm going to resolve this since `PrefixedFirebaseError` introduced in #85 makes error code checking easier.

PR #85, opened 2017-09-11
HttpRequestHandler: Retry on Network Error

justinrosenthal:
In an effort to make the admin SDK more resilient to temporary network errors, this PR adds a single retry to *all* requests made via `HttpRequestHandler` that fail due to a network error.  Timeouts and completed HTTP requests (regardless of response code) will never retry.

To work around #80 this PR also introduces `PrefixedFirebaseError`, a subclass of `FirebaseError` which attempts to make dealing with error-code prefixing a little easier by exposing a `hasCode()` method.

### DIFF
--- a/src/utils/api-request.ts
+++ b/src/utils/api-request.ts
@@ -32,6 +32,7 @@ export type ApiCallbackFunction = (data: Object) => void;
 export class HttpRequestHandler {
   /**
    * Sends HTTP requests and returns a promise that resolves with the result.
+   * Will retry once if the first attempt encounters an AppErrorCodes.NETWORK_ERROR.
    *
    * @param {string} host The HTTP host.
    * @param {number} port The port number.
@@ -43,6 +44,43 @@ export class HttpRequestHandler {
    * @return {Promise<Object>} A promise that resolves with the response.
    */
   public sendRequest(
+      host: string,
+      port: number,
+      path: string,
+      httpMethod: HttpMethod,
+      data?: Object,
+      headers?: Object,
+      timeout?: number): Promise<Object> {
+    // Convenience for calling the real _sendRequest() method with the original params.
+    const sendOneRequest = () => {
+      return this._sendRequest(host, port, path, httpMethod, data, headers, timeout);
+    };
+
+    return sendOneRequest()
+      .catch((response: { statusCode: number, error: string|Object }) => {
+        // Retry if the request failed due to a network error.
+        if (response.error instanceof FirebaseAppError) {
+          if ((<FirebaseAppError> response.error).hasCode(AppErrorCodes.NETWORK_ERROR)) {
+            return sendOneRequest();
+          }
+        }
+        return Promise.reject(response);
+      });
+  }
+
+  /**
+   * Sends HTTP requests and returns a promise that resolves with the result.
+   *
+   * @param {string} host The HTTP host.
+   * @param {number} port The port number.
+   * @param {string} path The endpoint path.
+   * @param {HttpMethod} httpMethod The http method.
+   * @param {Object} [data] The request JSON.
+   * @param {Object} [headers] The request headers.
+   * @param {number} [timeout] The request timeout in milliseconds.
+   * @return {Promise<Object>} A promise that resolves with the response.
+   */
+  private _sendRequest(
       host: string,
       port: number,
       path: string,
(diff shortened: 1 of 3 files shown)


bojeil-google:
I personally am not a fan of this. You can can add a retry using public APIs and this should not be done on the low level `HttpRequestHandler`.  Blind successive retries also have high probability of failure with the same error. In general, some exponential backoff is needed or some delay after the first request if you don't plan to keep trying.

inlined:
@bojeil-google This is actually more in line with google-cloud-node, which also retries on connection issues. E.g. See [the ECONNRESET bug](https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2254). Feel free to message me at inlined@ internally if you have any Qs for why we're more focused on these issues now.

bojeil-google:
This is a hack in my opinion, and an ugly one. I will try to follow up offline.

src/utils/error.ts
```diff
@@ -73,30 +73,67 @@ export class FirebaseError extends Error {
...
-      code: 'app/' + code,
+      code: `${codePrefix}/${code}`,
       message,
     });
+
+    /* tslint:disable:max-line-length */
+    // Set the prototype explicitly. See the following link for more details:
+    // https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work
+    /* tslint:enable:max-line-length */
+    (this as any).__proto__ = PrefixedFirebaseError.prototype;
```
hiranya911: Should we use `Object.setPrototypeOf()` here as shown in the doc you've linked?

justinrosenthal: This was just copied from elsewhere in the same file: https://github.com/firebase/firebase-admin-node/blob/master/src/utils/error.ts#L53

I applied it to every subclass, however, because the current implementation is buggy in that it actually breaks the inheritance.

I can't really speak to the difference between `(this as any).__proto__` and `Object.setPrototypeOf()` but happy to make the change if someone who knows better chimes in :)

inlined:
To be fair, there's a very good argument that we should not automatically retry HTTP errors unless we know whether the request was idempotent. For low-level TCP errors, we know that customers are seeing an abnormally high number of errors because infra outside the machine is killing these sockets uncleanly (e.g. NAT cleanups). This issue is becoming more apparent with auto-scaling and highly isolated environments like Cloud Functions, Lambda, Kubeless, etc. Even gRPC has agreed that this category of error will be retried regardless of the transport's retry policy. It would be most ideal if we can verify that no bytes were successfully transmitted as part of the request though.

hiranya911:
@bojeil-google, @justinrosenthal Have you guys reached a consensus about this PR? If I can get answers to the couple of questions I posted above, I can arrive at an LGTM.

justinrosenthal:
I don't believe there's been a high level discussion outside of this PR, yet.  But in the meantime, to address your questions:

1. Should we implement a bit more sophisticated retry logic? (e.g. retry 3 times with some delay in between)

My intention was to do the simplest thing that would improve behavior in the common case we're trying to solve: sockets getting in a bad state in environments like Cloud Functions.  Any implementation that introduces delays should be configurable, and appropriate values chosen for each use of `HttpRequestHandler`.  Unless there's an existing need for configurable retries + backoff elsewhere in the SDK, I would just assume keep this PR straightforward.

2. In the event of a network error, should we do something extra to tear down the existing socket/connection? (call req.abort() perhaps)

I don't see any indication in the documentation that's necessary.  I believe if there's a networking error the request is considered finished without needing to explicitly mark it as such, and the socket wouldn't be reused.  Let me know if you have evidence to the contrary.

hiranya911:
We call `req.abort()` in the event of a socket timeout. I'm just wondering whether we need to do the same in case of an exception.

src/utils/api-request.ts
```diff
@@ -50,6 +51,43 @@ export class HttpRequestHandler {
...
+    // Convenience for calling the real _sendRequest() method with the original params.
+    const sendOneRequest = () => {
+      return this._sendRequest(host, port, path, httpMethod, data, headers, timeout);
+    };
+
+    return sendOneRequest()
+      .catch((response: { statusCode: number, error: string|Object }) => {
+        // Retry if the request failed due to a network error.
+        if (response.error instanceof FirebaseAppError) {
+          if ((<FirebaseAppError> response.error).hasCode(AppErrorCodes.NETWORK_ERROR)) {
```
bojeil-google: By the way, all request errors are currently considered network errors:
`req.on('error', (error) => {...})`
That could be too general. Would it be better to catch ECONNRESET errors and retry only in that case?

bojeil-google:
Sorry for the delay. The implementation as designed is fine. Would be curious to see the change in failure rates after this change.

src/utils/api-request.ts
justinrosenthal: Indeed this retries for any non-HTTP and non-timeout error: connection reset, broken pipe, etc.  @inlined You're most familiar with the ways we've seen this present, do you think it's worth passing up the original error and having a whitelist here?

inlined: ECONNRESET is the one off the top of my head.

inlined:
I agree that affected failure rate stats would be great. Unfortunately our best metric will either be global failures for any reason or customer reports. We took the privacy-safe stance of just aggregating user-space errors as a black box & incrementing a metric. After triaging customer issues we've started to believe that aggregate "user errors" will reveal things that might actually be environment and or SDK issues (whether that means a bugfix or just hardening like this PR). It will take some time to figure out & implement a data pipeline that can give us this insight (common "user errors" that we can help fix) with appropriate sanitization to respect customer privacy.

hiranya911:
@justinrosenthal Feel free to merge this when you're happy with the outcome. I can include it in the next release.

PR #106, opened 2017-10-11
Static Loading of Service Namespaces

hiranya911:
Admin SDK monkeypatches service namespaces (`Auth`, `Messaging` etc.) to the parent `admin` namespace at SDK initialization. The `index.ts` file calls a set of `register*()` functions, which imports the services, and attaches them to `admin`. However, for most services this both unnecessary, and error prone.

This PR gets rid of the monkeypatching for all services but RTDB. There are several benefits for this change:
1. We can get rid of a significant amount of boilerplate code, whose sole purpose was to facilitate dynamically injecting various services into the `admin` namespace.
2. Services can be statically exposed from `admin` in a way that is both simple and type safe. Services will continue to be loaded/initialized on-demand (e.g. `Auth` service will not get initialized until `admin.auth()` is invoked).
3. Integration between services and the `admin` namespace can be verified with unit tests.
4. We can expose arbitrary types from the `admin` namespace without having them implement the `FirebaseServiceInterface`. This is useful when exposing external types such as `Firestore` from the `admin` namespace.


src/firebase-namespace.ts
```diff
@@ -25,6 +25,11 @@ import {
   ApplicationDefaultCredential,
 } from './auth/credential';
 import {Firestore} from '@google-cloud/firestore';
+import * as _ from 'lodash/object';
```
jshcrowthe: AFAICT you are bringing this in to do:

```javascript
_.assign()
```

There is a native JS version of this called `Object.assign`, it may be able to trim a dependency for you.

test/integration/firestore.js
```diff
@@ -47,29 +47,34 @@ function test(utils) {
...
 
   function testFieldValue() {
-    const expected = {
-        name: 'Mountain View',
-        population: 77846,
-        timestamp: admin.firestore.FieldValue.serverTimestamp(),
-    };
     const reference = admin.firestore().collection('cities').doc()
-    return reference.set(expected)
+    return Promise.resolve()
```
jshcrowthe: What was the rationale behind this change?

It seems like you've added an extra step in the chain unnecessarily.

hiranya911: This is to allow any errors thrown by `admin.firestore.FieldValue` to be caught in the `catch` block. I noticed that errors thrown outside the promise chain, cause the test suite to come to a halt.

src/firebase-namespace.ts
hiranya911: Brilliant! Done. Thanks for the tip.

test/integration/firestore.js
jshcrowthe: Ohhh gotcha. Ok, then yeah this is fine then.

src/firebase-namespace.ts
```diff
@@ -273,16 +278,15 @@ export class FirebaseNamespace {
...
+   * Gets the `Auth` service namespace. The returned namespace can be used to get the
+   * `Auth` service for the default app or an explicitly specified app.
    */
-  /* istanbul ignore next */
-  public auth(): FirebaseServiceInterface {
-    throw new FirebaseAppError(
-      AppErrorCodes.INTERNAL_ERROR,
-      'INTERNAL ASSERT FAILED: Firebase auth() service has not been registered.',
-    );
+  get auth(): FirebaseServiceNamespace<Auth> {
```
phra: this line breaks [es6 import mocking](https://railsware.com/blog/2017/01/10/mocking-es6-module-import-without-dependency-injection/)..

Issue #121, opened 2017-10-25
npm update firebase-admin fails with ENOENT: no such file or directory, rename


dschnelldavis:
I think I have the same issue as posted [here](https://github.com/npm/npm/issues/18916) for NPM. However, I have not had this issue with any other packages, so I think the problem is firebase-admin version 5.4.3 (version 5.4.2 works fine), not NPM.

#### System info
Mac OS: 10.12.6
Node: 8.6.0
NPM:  5.5.1
firebase-admin: 5.4.2 -> 5.4.3
Installed NPM Packages:
> @ google-cloud/storage@1.4.0
> @ types/fs-extra@4.0.3
> child-process-promise@2.2.1
> cloudconvert@1.1.2
> firebase-admin@5.4.2
> firebase-functions@0.7.1
> fs-extra@4.0.2
> lodash@4.17.4
> mailgun-js@0.13.1
> mime@1.4.1
> ringcentral@3.1.3
> typescript@2.5.3

#### Install Error
Running `npm update` gives this error:
```shell
npm ERR! path /Users/Username/Sites/Sitename/functions/node_modules/firebase-admin/node_modules/grpc/node_modules/abbrev
npm ERR! code ENOENT
npm ERR! errno -2
npm ERR! syscall rename
npm ERR! enoent ENOENT: no such file or directory, rename '/Users/Username/Sites/Sitename/node_modules/firebase-admin/node_modules/grpc/node_modules/abbrev' -> '/Users/Username/Sites/Sitename/node_modules/firebase-admin/node_modules/grpc/node_modules/.abbrev.DELETE'
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/Username/.npm/_logs/2017-10-25T01_25_07_312Z-debug.log
```
And here's the relevant part of the log file:
```text
8231 verbose unlock done using /Users/Username/.npm/_locks/staging-e6d3ce262201f329.lock for /Users/Username/Sites/Sitename/node_modules/.staging
8232 verbose stack Error: ENOENT: no such file or directory, rename '/Users/Username/Sites/Sitename/node_modules/firebase-admin/node_modules/grpc/node_modules/abbrev' -> '/Users/Username/Sites/Sitename/node_modules/firebase-admin/node_modules/grpc/node_modules/.abbrev.DELETE'
8233 verbose cwd /Users/Username/Sites/Sitename
[...]

PR #124, opened 2017-10-26
Removed skrinkwrap File

hiranya911:
Migrating from `npm-shrinkwrap.json` to `package-lock.json`. This file is far more stable than the shrinkwrap, and it cannot be published with releases. Users who install `firebase-admin` will get the dependencies as stated in our `package.json` file. Package lock also gets automatically updated when we update the `package.json` file.

### DIFF
--- a/createReleaseTarball.sh
+++ b/createReleaseTarball.sh
@@ -89,13 +89,13 @@ sed -i '' -e s/"\"version\": \".*\""/"\"version\": \"${VERSION_WITHOUT_RC}\""/ p
 echo
 
 
-####################################
-#  REGENERATE npm-shrinkwrap.json  #
-####################################
-echo "[INFO] Removing lib/, node_modules/, and npm-shrinkwrap..."
-rm -rf lib/ node_modules/ npm-shrinkwrap.json
+############################
+#  REINSTALL DEPENDENCIES  #
+############################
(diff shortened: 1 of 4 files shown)


createReleaseTarball.sh
```diff
@@ -89,13 +89,13 @@ sed -i '' -e s/"\"version\": \".*\""/"\"version\": \"${VERSION_WITHOUT_RC}\""/ p
...
-####################################
-#  REGENERATE npm-shrinkwrap.json  #
-####################################
-echo "[INFO] Removing lib/, node_modules/, and npm-shrinkwrap..."
-rm -rf lib/ node_modules/ npm-shrinkwrap.json
+############################
+#  REINSTALL DEPENDENCIES  #
+############################
+echo "[INFO] Removing lib/, and node_modules/..."
+rm -rf lib/ node_modules/
```
jshcrowthe: Does it make sense to, since you're trying to clean install here, just do:

```bash
git clean -xdf
```

And just prune everything that doesn't exist by default in the git tree?

hiranya911: This is a good suggestion. But I think we need to add a few more entries to  `.gitignore` before we do this. Will do as a separate PR.

hiranya911:
Related to #121

Issue #136, opened 2017-11-14
Wrong "typings" key in package.json after monorepo refactor.


adamduren:
## Reproduction steps
See [repo](https://github.com/adamduren/firebase-typings-issue) to reproduce.

## Description

The path for typings is missing the `cjs` directory.

```
"typings": "dist/index.d.ts",
```

Should be 
```
"typings": "dist/cjs/index.d.ts",
```

Noticed in the following places:

 * https://github.com/firebase/firebase-js-sdk/blob/master/packages/app/package.json#L50
 * https://github.com/firebase/firebase-js-sdk/blob/master/packages/messaging/package.json#L48
 * https://github.com/firebase/firebase-js-sdk/blob/master/packages/storage/package.json#L45

Results in the following error during compilation with `tsc`

```
node_modules/@firebase/database/dist/cjs/index.d.ts(16,35): error TS7016: Could not find a declaration file for module '@firebase/app'. '/workspace/functions/node_modules/@firebase/app/dist/cjs/index.js' implicitly has an 'any' type.
node_modules/@firebase/database/dist/cjs/src/api/Database.d.ts(3,46): error TS7016: Could not find a declaration file for module '@firebase/app'. '/workspace/functions/node_modules/@firebase/app/dist/cjs/index.js' implicitly has an 'any' type.
  Try `npm install @types/@firebase/app` if it exists or add a new declaration (.d.ts) file containing `declare module '@firebase/app';`
node_modules/@firebase/database/dist/cjs/src/core/AuthTokenProvider.d.ts(16,52): error TS7016: Could not find a declaration file for module '@firebase/app'.'/workspace/functions/node_modules/@firebase/app/dist/cjs/index.js' implicitly has an 'any' type.
  Try `npm install @types/@firebase/app` if it exists or add a new declaration (.d.ts) file containing `declare module '@firebase/app';`
node_modules/@firebase/database/dist/cjs/src/core/Repo.d.ts(3,29): error TS7016: Could not find a declaration file for module '@firebase/app'. '/workspace/functions/node_modules/@firebase/app/dist/cjs/index.js' implicitly has an 'any' type.
  Try `npm install @types/@firebase/app` if it exists or add a new declaration (.d.ts) file containing `declare module '@firebase/app';`
```

## Environment
 * Node Version: v8.9.1 (also tried 6.x, 7.x)
 * NPM Version: v5.5.1
 * TypeScript Version: v2.4.2 / v2.6.1

[...]

PR #140, opened 2017-11-20
Adding the RTDB typings back

hiranya911:
Adding the old RTDB typings back until we figure out a more permanent solution with the JS SDK. Should resolve issues like #136

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -15,7 +15,6 @@
  */
 
 import {Bucket} from '@google-cloud/storage';
-import * as _rtdb from '@firebase/database';
 import * as _firestore from '@google-cloud/firestore';
 
 declare namespace admin {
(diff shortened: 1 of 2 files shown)


src/index.d.ts
```diff
@@ -175,6 +174,118 @@ declare namespace admin.credential {
...
+
+    child(path: string): admin.database.DataSnapshot;
+    exists(): boolean;
+    exportVal(): any;
+    forEach(action: (a: admin.database.DataSnapshot) => boolean): boolean;
+    getPriority(): string|number|null;
+    hasChild(path: string): boolean;
+    hasChildren(): boolean;
+    numChildren(): number;
+    toJSON(): Object;
```
jshcrowthe: JS SDK has this line as `Object | null`

src/index.d.ts
```diff
@@ -175,6 +174,118 @@ declare namespace admin.credential {
...
+    val(): any;
+  }
+
+  interface OnDisconnect {
+    cancel(onComplete?: (a: Error|null) => any): Promise<void>;
```
jshcrowthe: JS SDK does `Promise<any>` here and on the rest of the functions in this interface.

src/index.d.ts
```diff
@@ -175,6 +174,118 @@ declare namespace admin.credential {
...
+    startAt(value: number|string|boolean|null, key?: string): admin.database.Query;
+    toJSON(): Object;
+    toString(): string;
+  }
+
+  interface Reference extends admin.database.Query {
+    key: string|null;
+    parent: admin.database.Reference|null;
+    root: admin.database.Reference;
+
```
jshcrowthe: This is missing the `path` property?

src/index.d.ts
```diff
@@ -175,6 +174,118 @@ declare namespace admin.credential {
...
+    child(path: string): admin.database.Reference;
+    onDisconnect(): admin.database.OnDisconnect;
+    push(value?: any, onComplete?: (a: Error|null) => any): admin.database.ThenableReference;
+    remove(onComplete?: (a: Error|null) => any): Promise<void>;
```
jshcrowthe: `Promise<any>`?

src/index.d.ts
```diff
@@ -175,6 +174,118 @@ declare namespace admin.credential {
...
+      onComplete?: (a: Error|null, b: boolean, c: admin.database.DataSnapshot|null) => any,
+      applyLocally?: boolean
+    ): Promise<{
+      committed: boolean,
+      snapshot: admin.database.DataSnapshot|null
+    }>;
+    update(values: Object, onComplete?: (a: Error|null) => any): Promise<void>;
+  }
+
+  interface ThenableReference extends admin.database.Reference, Promise<any> {}
```
jshcrowthe: JS SDK extends a `PromiseLike<any>`

src/index.d.ts

src/index.d.ts

src/index.d.ts
hiranya911: I will keep this unchanged for now, since we already seem to have some other (non-DB) APIs that return `Promise<void>`

PR #152, opened 2017-12-06
fix(package): drop unused dep google-auth-library

ofrobots:
I was trying to integration-test a semver major update, that includes some [breaking changes](https://github.com/google/google-auth-library-nodejs/releases/tag/1.0.0-alpha.1), being made over at [google-auth-library](https://github.com/google/google-auth-library-nodejs). To my surprise, all tests passed. It turns out the package.json is the only thing that depends on `google-auth-library`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,11 +47,10 @@
   ],
   "types": "./lib/index.d.ts",
   "dependencies": {
-    "@firebase/database": "^0.1.3",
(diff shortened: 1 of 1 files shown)


ofrobots:
No, there are no changes to package-lock.json. This package still transitively depends on google-auth-library.

ofrobots:
Can this be landed now?

Issue #153, opened 2017-12-06
Start testing against Node 8 and 9


ofrobots:
Hi Folks, over at my PR (#265) I noticed that the tests are run against Node 4, 5, 6, 7. Is there a reason for not testing against the actual LTS and Current releases (8.x and 9.x respectively).

BTW, Node 5 and 7 are no longer supported upstream. You can stop testing against those.

hiranya911:
@jshcrowthe WDYT?

hiranya911:
I actually got a test failure on Node 9: https://travis-ci.org/firebase/firebase-admin-node/builds/313726251

hiranya911:
Testing on Node 8 is now in place.

hiranya911:
Closing this since we have Node 8 CI set up now. We can add Node 9 in the future.

ofrobots:
Is there rationale for not starting to test against Node 9 immediately? Node 9 is the "current" version of Node, and as you can see from the [charts here](https://nodesource.com/node-by-numbers), Node 9 is already has significant usage.

Either way, I don't believe you've adequately addressed the original issue I opened that requested testing on both 8.x and 9.x. Either this bug should be reopened, or you should open another bug to track the addition of Node 9.x, if you're not going to address it immediately.

PR #175, opened 2018-01-05
Added Node 8 to the list of environments we test against

hiranya911:
Node 8 is the latest stable release of Node.js. And according to #153 Node 5 is no longer supported upstream. Therefore this PR trades Node 5 for 8 in the set of environments tested in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
+  - "8"
   - "7"
   - "6"
-  - "5"
   - "4"


hiranya911:
I think https://github.com/nodejs/Release provides context. According to this,

* Node 4: Long term support provided until April 2018
* Node 6: Long term support provided until April 2019
* Node 8: Active long term support
* Node 5 and 7: Neither active, nor long term support provided

bojeil-google:
Interesting. Thanks for the clarification.

AdriVanHoudt:
It might be worth adding just `node` to the version, this will make sure that you always test latest as well.

PR #181, opened 2018-01-11
Bumped version to 5.8.0

hiranya911:


package-lock.json
```diff
@@ -4522,12 +4522,14 @@
         "camelcase": {
           "version": "1.2.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
```
avishalom: do we want all these "optional"s in this PR?

PR #188, opened 2018-01-18
Upgraded nock to 9.x

hiranya911:


hiranya911:
@merlinnot good catch. I meant to upgrade both types and implementation. 

@avishalom PTAL.

PR #192, opened 2018-01-23
Updating changelog when building releases

hiranya911:


createReleaseTarball.sh
```diff
@@ -88,6 +88,13 @@ echo "[INFO] Updating version number in package.json to ${VERSION_WITHOUT_RC}...
 sed -i '' -e s/"\"version\": \".*\""/"\"version\": \"${VERSION_WITHOUT_RC}\""/ package.json
 echo
 
+#########################
+#  UPDATE CHANGELOG.md  #
+#########################
+echo "[INFO] Updating version number in CHANGELOG.md to ${VERSION_WITHOUT_RC}..."
+sed -i '' -e s/"# Unreleased"/"# v${VERSION_WITHOUT_RC}"/ CHANGELOG.md
```
avishalom: This seems a bit fragile, What if "# Unreleased" appears anywhere else. 
Also, it seems that the current version doesn't have a blank line in line #2 .
What if it was deleted in the process. 

I suggest, 
- only seding the first line,
- replace "# Unreleased" or ignore it and append if it isn't there
- append to the end of the first line so that anything non congruent stands out 
s.t. like  
> sed -i '' -E '1 s/(# Unreleased|^)/# Unreleased'$'\\\n\\\n-\\\n'"# v${VERSION_WITHOUT_RC}/"

instead of lines 95-97

hiranya911: Updated the `sed` command to only replace the first occurrence. Points 2 and 3 hard to implement in a way that is compatible with both Linux and Mac. So I'm keeping the simple `cat to file, and move` solution for now.

avishalom: My only worry is that if someone accidentally changes that first line, the version will not be added.
How about replacing the first "# Unreleased" with nothing ("") and adding the v$VERSION_WITHOUT_RC in the echo in the next line ?

PR #194, opened 2018-01-24
Upgrading tslint to 5.9.0

hiranya911:
Upgrading to a more recent version of tslint, which performs a wide range of checks on source. I've updated the source to be compliant.


src/auth/auth-api-request.ts
```diff
@@ -320,9 +319,9 @@ export class FirebaseAuthRequestHandler {
    * Looks up a user by uid.
    *
    * @param {string} uid The uid of the user to lookup.
-   * @return {Promise<Object>} A promise that resolves with the user information.
+   * @return {Promise<object>} A promise that resolves with the user information.
```
bojeil-google: Does the linter complain about javadoc types too?

src/auth/auth-api-request.ts
```diff
@@ -597,48 +596,41 @@ export class FirebaseAuthRequestHandler {
...
-        let error;
-        if (typeof response === 'object' && 'statusCode' in response) {
-          // response came directly from a non-200 response.
-          error = response.error;
-        } else {
-          // response came from a thrown error on a 200 response.
-          error = response;
-        }
-
+        const error = (typeof response === 'object' && 'statusCode' in response)
```
bojeil-google: Better for readability to have the ? on the previous line:
```
const error = (typeof response === 'object' && 'statusCode' in response) ?
  response.error : response;
```
Here and elsewhere

src/auth/credential.ts
```diff
@@ -237,12 +237,9 @@ function requestAccessToken(transit, options: Object, data?: Object): Promise<Go
...
 
-  constructor(serviceAccountPathOrObject: string|Object) {
-    if (typeof serviceAccountPathOrObject === 'string') {
-      this.certificate_ = Certificate.fromPath(serviceAccountPathOrObject);
-    } else {
-      this.certificate_ = new Certificate(serviceAccountPathOrObject);
-    }
+  constructor(serviceAccountPathOrObject: string | object) {
+    this.certificate_ = (typeof serviceAccountPathOrObject === 'string')
+      ? Certificate.fromPath(serviceAccountPathOrObject) : new Certificate(serviceAccountPathOrObject);
```
bojeil-google: Better for readability to have the ? on the previous line

src/auth/credential.ts
```diff
@@ -302,12 +299,9 @@ export interface Credential {
...
 
-  constructor(refreshTokenPathOrObject: string|Object) {
-    if (typeof refreshTokenPathOrObject === 'string') {
-      this.refreshToken_ = RefreshToken.fromPath(refreshTokenPathOrObject);
-    } else {
-      this.refreshToken_ = new RefreshToken(refreshTokenPathOrObject);
-    }
+  constructor(refreshTokenPathOrObject: string | object) {
+    this.refreshToken_ = (typeof refreshTokenPathOrObject === 'string')
+      ? RefreshToken.fromPath(refreshTokenPathOrObject) : new RefreshToken(refreshTokenPathOrObject);
```
bojeil-google: Better for readability to have the ? on the previous line

src/instance-id/instance-id-request.ts
```diff
@@ -94,26 +93,17 @@ export class FirebaseInstanceIdRequestHandler {
...
       .catch((response) => {
-        let error;
-        if (typeof response === 'object' && 'error' in response) {
-          error = response.error;
-        } else {
-          error = response;
-        }
-
+        const error = (typeof response === 'object' && 'error' in response)
+          ? response.error : response;
```
bojeil-google: Better for readability to have the ? on the previous line

src/instance-id/instance-id-request.ts
```diff
@@ -94,26 +93,17 @@ export class FirebaseInstanceIdRequestHandler {
...
-        let template: string = ERROR_CODES[response.statusCode];
-        let message: string;
-        if (template) {
-          message = `Instance ID "${apiSettings.getEndpoint()}": ${template}`;
-        } else {
-          message = JSON.stringify(error);
-        }
+        const template: string = ERROR_CODES[response.statusCode];
+        const message: string = template
+          ? `Instance ID "${apiSettings.getEndpoint()}": ${template}` : JSON.stringify(error);
```
bojeil-google: Better for readability to have the ? on the previous line

src/auth/auth-api-request.ts
hiranya911: Not really, but I thought it makes sense to keep the jsdoc types in sync with the implementation. WDYT?

src/auth/auth-api-request.ts
hiranya911: Done

src/auth/credential.ts

src/auth/credential.ts

src/instance-id/instance-id-request.ts
hiranya911: Done

src/instance-id/instance-id-request.ts
hiranya911: Done

src/auth/auth-api-request.ts
bojeil-google: The downside is that it will not align with our externs format. Could this be an issue?

hiranya911: I think that's fine since the externs are curated by hand anyway. Going forward we can adhere to the distinction:
* Only Typescript in source (everything on Github)
* Only JS in API docs (externs)

PR #198, opened 2018-01-30
Removing '# Unreleased' with sed instead of replacing

hiranya911:
Replacing the pattern leaves an extra newline character around which has to be removed manually. I'm using a stricter pattern match here (with line delimiters), and then deleting the matched line with sed altogether.


hiranya911:
That's overkill. We can curate these files over time. Even if something goes wrong, it's easily correctable.

PR #201, opened 2018-02-01
Applying Linter on Tests

hiranya911:
So far we haven't been linting our test sources. Now that we have upgraded to the latest version of tslint (#194), we should start linting our unit and integration tests. This PR adds the following npm tasks:

```
npm run lint:src # Lints the source files (everything under src/)
npm run lint:unit # Lints unit tests
npm run lint:integration # Lints integration tests
```

The aggregate task 'npm run lint` runs all of the above, which will run as part of `npm test` as well (this causes all lint jobs to run with CI). 

I also fixed all the existing lint errors in test files. I relaxed the linter rules a little bit for tests by introducing a separate `tslint-test.json` file.


package.json
```diff
@@ -7,12 +7,15 @@
   "homepage": "https://firebase.google.com/",
   "scripts": {
     "build": "gulp build",
-    "lint": "tslint --project tsconfig-lint.json --format stylish",
+    "lint": "run-s lint:src lint:unit lint:integration",
```
merlinnot: `npm-run-all` supports glob-like pattern matching, so you can do `run-s lint:*`.

Why did you decide to run these in sequence instead of parallel?

hiranya911:
@merlinnot What should `lint:*:fix` do? Run `tslint` with the `--fix` flag?

Good idea to run the linter tasks in parallel.

merlinnot:
Since `tslint` is a dev dependency (which is a good thing!) one should use `node_modules/tslint/bin/... —config ... —fix` to fix all auto-fixable issues, which is a bit annoying. It’s much easier to `yarn run lint:unit:fix` or `yarn run lint:fix`. You can even leverage existing scripts like so:
```json
{
  ...,
  “lint:unit:fix”: “run-s \”lint:unit —fix\””
  ...
}
```

One day we can even add precommit hooks on top of this, so it fixes everything before committing using [husky](https://github.com/typicode/husky) and [lint-staged](https://github.com/okonet/lint-staged), but that shouldn’t be a part of this PR.

Issue #77, opened 2017-08-25
TypeError: Cannot read property 'navigator' of undefined


moander:
I experience this error when running mocha with --use_strict argument

Version 4.2.1
node v6.11.2
mocha v3.5.0

```
TypeError: Cannot read property 'navigator' of undefined
    at node_modules\firebase-admin\lib\database\database.js:39:492
    at Object.<anonymous> (node_modules\firebase-admin\lib\database\database.js:254:3)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (node_modules\firebase-admin\lib\index.js:11:1)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (node_modules\firebase-functions\lib\apps.js:25:16)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (node_modules\firebase-functions\lib\providers\database.js:25:14)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (node_modules\firebase-functions\lib\index.js:32:16)
```

hiranya911:
What exactly are you trying to achieve? If you want to run Admin SDK test suites just follow the instructions at https://github.com/firebase/firebase-admin-node/blob/master/CONTRIBUTING.md#running-tests

moander:
It's my own test cases that fails to run.

All tests referencing firebase-admin fails with this error.

[...]

PR #221, opened 2018-03-01
Send email verification

### LINKED ISSUE #46: Feature Request: Cloud Function should support user.sendEmailVerification() like client sdk
[Refiling for user @sun3 from https://github.com/firebase/functions-samples/issues/181]

Feature Request:

The ```user.sendEmailVerification()``` needs to be supported by Firebase Cloud Functions. This is available in client side but not server side. When using Cloud Functions to create new users we also need to automatically send Email Verification before new user can use our apps. This is currently as stopping block for our apps.

At this point the ```user.sendEmailVerification()``` gives error that function is not found.

```
    admin.auth().createUser({
        email: emailAddress,
        emailVerified: false,
        password: password,
        displayName: '', //name,
        disabled: false
    })
        .then(function (user) {
            // A error representation of the newly created user is returned
            console.log("Created Firebase User successfully with id: ", user.uid);
            console.log("user.emailVerified:", user.emailVerified);

            // Send Email Verification
            user.sendEmailVerification()
                .then(function (emailSent) {
                    console.log('emailSent ', emailSent);
                })
                .catch(function (error) {
                    console.log('emailSent error ', error);
                });

    // ... Additional code below
```

Thank you and I am open to any suggestions.

Crazometer:
Fixes #46 

Adds `admin.auth().sendEmailVerification(uid)`. Tested it locally and it works great.

This could be made much nicer if the `/getOobConfirmationCode` endpoint was changed to optionally accept a `localId` (uid) instead of an `idToken`. This is how the other endpoints work. This would make the process use one less REST call and not require the use of `createCustomToken` which isn't supported by all environments. 

Let me know if this is going in the right direction and I'll clean it up a bit and add tests (or you can).

### DIFF
--- a/src/auth/auth-api-request.ts
+++ b/src/auth/auth-api-request.ts
@@ -287,6 +287,19 @@ export const FIREBASE_AUTH_SIGN_UP_NEW_USER = new ApiSettings('signupNewUser', '
         'INTERNAL ASSERT FAILED: Unable to create new user');
     }
   });
+/**
+ * Instantiates the getOobConfirmationCode endpoint settings for sending a verification email.
+ */
+export const FIREBASE_AUTH_GET_OOB_CONFIRMATION_CODE = new ApiSettings('getOobConfirmationCode', 'POST')
+  // Set request validator.
+  .setRequestValidator((request: any) => {
+    // requestType is required
+    if (typeof request.requestType === 'undefined') {
+      throw new FirebaseAuthError(
+        AuthClientErrorCode.INTERNAL_ERROR,
+        'INTERNAL ASSERT FAILED: Server request is missing request type');
+    }
+  });
 
 /**
  * Class that provides mechanism to send requests to the Firebase Auth backend endpoints.
--- a/src/auth/auth-api-request.ts
+++ b/src/auth/auth-api-request.ts
@@ -301,6 +301,17 @@ export const FIREBASE_AUTH_GET_OOB_CONFIRMATION_CODE = new ApiSettings('getOobCo
     }
   });
 
+export const FIREBASE_AUTH_VERIFY_CUSTOM_TOKEN = new ApiSettings('verifyCustomToken', 'POST')
+  // Set request validator.
+  .setRequestValidator((request: any) => {
+    // token is required
+    if (typeof request.token === 'undefined') {
+      throw new FirebaseAuthError(
+        AuthClientErrorCode.INTERNAL_ERROR,
+        'INTERNAL ASSERT FAILED: Server request is missing custom token');
+    }
+  });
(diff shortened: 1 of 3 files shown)


bojeil-google:
Thank you for your interest in this feature. We really appreciate your contribution. For new APIs, you need to file a new feature requests.
We have an extensive internal API review process. This has to go through that review first.
https://firebase.google.com/support/contact/bugs-features/

That said, the API proposed is inefficient and combines admin and client endpoints. You are making 3 round trips for this feature. This is exactly what developers are already doing to support this via Admin SDK.

Thanks again for your effort. We plan to support this API with a dedicated endpoint. It is on our roadmap.

bojeil-google:
Just to clarify, no admin SDK endpoint should rely on an ID token to make an authorized call. That belongs on the client SDK. For Admin SDK a service account credential is enough to send the authorized call to send an email verification (without the same client side limitations). Client side calls are subject to throttling and different checks than admin endpoints. That is why we can't rely on internal implementations in the admin SDK with such restrictions.

Crazometer:
Thanks for the response,

I agree that the implementation is a bit messy but do you think this could be handy as an interim solution because:

1. #77 has been open for a long time
2. The suggested solution involves including the client-sdk (fairly large) and has a number of steps. This strays from the simplicity most firebase services offer.
3. When the endpoint is eventually implemented the transition will be seamless to the developer.

Perhaps you could mark the method as experimental?

> So for example, a developer using this implementation to send multiple email verifications could end up getting rate limited as from the perspective of the Auth backend, this is coming from a client API.

Correct me if I'm wrong but this would only prevent you from sending a large number of verify emails to the same user. This is probably desired behaviour for most use cases.

bojeil-google:
You are free to use this solution on your own. In fact, right now, your options are limited and this is a good workaround.

However, we can't accept a hack in our official SDKs for various reasons.

As I explained earlier, we have a rigorous review process for new APIs. Even if I wanted to roll out a new API, I can't just simply do it.

hiranya911:
I think Auth backend servers will see this as initializing many login sessions from the same IP/device, which is a problem. I kind of like the idea of having this as an interim solution though (clearly marked as experimental in the documentation). But I'll side with @bojeil-google and the auth team's decision on this since they understand the implications of this far better than I do.

bojeil-google:
I can use this as a way to push harder for this feature. I totally understand the importance of having an easy way to send verifications via the admin SDK. We have several features in the pipeline that developers have been asking for for some time. You will be seeing them soon. Once we can get those out of the door, we can focus on this feature.

Once again, I think you can use this solution though don't be surprised if you get throttled.

PR #265, opened 2018-05-13
Lazy loading modules for faster load times

hiranya911:
Lazy loading internal modules and some dependencies for faster load times. This essentially defers loading child modules until they are explicitly required (using the technique described [here](http://www.typescriptlang.org/docs/handbook/modules.html#optional-module-loading-and-other-advanced-loading-scenarios)). For example none of the `auth/*` sources will be loaded until `admin.auth()` is invoked. This also ensures that only the child modules used by the developers get loaded into the Node.js process.

## Experimental results

Test code:
```
const start = Date.now();
require('firebase-admin');
const end = Date.now();
console.log('Load time:', (end-start), 'ms');
```

### MacBook Pro (Core i7 2.2 GHz, 16GB RAM)

Before (ms): 154, 151, 151, 153, 153 (mean = 152.4ms)
After (ms): 18, 18, 17, 18, 18 (mean = 17.8ms)
Speed up of initial load: 88.32%

### Ubuntu 16.04 (Core i7 2.8GHz, 16GB RAM)

Before (ms): 61, 59, 59, 59, 59 (mean = 59.4 ms)
After (ms): 4, 5, 5, 4, 5 (mean = 4.6 ms)
Speed up of initial load: 92.25%

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1100,6 +1100,12 @@
       "integrity": "sha512-1v3/dB01fkW5qTpZH5zjjBLgj97Jhpu/tjJt/g0LHMPTuj2D2a6sJhj9RcwqmXl+TtokCZQyaynxL4buN5G2WA==",
       "dev": true
     },
+    "@types/caseless": {
+      "version": "0.12.1",
+      "resolved": "https://registry.npmjs.org/@types/caseless/-/caseless-0.12.1.tgz",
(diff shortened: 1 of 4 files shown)


hiranya911:
Done. Removed package-lock changes from the PR.

Also FYI @laurenzlong. I think Functions is likely to be the biggest beneficiary of this change.

PR #279, opened 2018-05-24
Fixing a test failure caused by Firestore 0.14.1 release

hiranya911:
`projectId` field is no longer reliably exposed by the Firestore client -- it seems to be lazy initialized.


hiranya911:
FYI @schmidt-sebastian

PR #288, opened 2018-06-06
Changes Auth integration tests for session cookies to use multiple UIDs

bojeil-google:
Changes Auth integration tests for session cookies to use multiple UIDs to minimize the risk of flakiness in integration tests.


bojeil-google:
Thanks for the review.

PR #290, opened 2018-06-09
Adds additional delay at end of revocation session cookie test

bojeil-google:
Adds additional delay at end of revocation session cookie test so this doesn't affect other tests that use the same uid.

### DIFF
--- a/test/integration/auth.spec.ts
+++ b/test/integration/auth.spec.ts
@@ -391,6 +391,11 @@ describe('admin.auth', () => {
         .then(() => {
           return admin.auth().verifySessionCookie(currentSessionCookie, true)
             .should.eventually.be.rejected.and.have.property('code', 'auth/session-cookie-revoked');
+        })
+        .then(() => {
+          // Wait an additional second to avoid token revocation for specified uid
+          // from affecting other tests that use the same uid.
+          return new Promise((resolve) => setTimeout(resolve, 1000));
         });
     });
 


test/integration/auth.spec.ts
```diff
@@ -391,6 +391,11 @@ describe('admin.auth', () => {
         .then(() => {
           return admin.auth().verifySessionCookie(currentSessionCookie, true)
             .should.eventually.be.rejected.and.have.property('code', 'auth/session-cookie-revoked');
+        })
+        .then(() => {
+          // Wait an additional second to avoid token revocation for specified uid
+          // from affecting other tests that use the same uid.
+          return new Promise((resolve) => setTimeout(resolve, 1000));
```
hiranya911: I'm not entirely sure what this is expected to achieve. We are not using this UID in any other tests as far as I can see.  Plus we have a 1s delay before each revocation in the test suite. Can you please elaborate why this is needed?

bojeil-google:
This is needed because we re-use the first session cookie uid for this test:
https://github.com/firebase/firebase-admin-node/blob/master/test/integration/auth.spec.ts#L415

It is possible by the time the test reached the `createCustomToken` call, we are still within revocation (1 second) range.

I have 2 options: add the delay or create a new uid for this test.

hiranya911:
We are already using `uid2` in this test case. The test case at L415 uses `uid`. Isn't that enough?

bojeil-google:
Ah you are right, we use `uid2` too here. Ok, then in that case, there is no need for this. Will close this.

PR #280, opened 2018-05-24
Re-implementing the HTTP Client Abstraction

hiranya911:
The existing HTTP client abstractions (`HttpRequestHandler` and `SignedApiRequestHandler`) have a number of caveats:

1. Do not support HTTP endpoints (HTTPS only)
2. Do not expose response headers
3. Complex and often confusing error handling semantics (e.g. even low-level network errors are reported as errors with HTTP status codes)
4. Support for handling different payload types is broken (e.g. cannot handle anything other than `application/json`, `text/html` and `text/plain`)
5. Not easy to extend or test (public API takes 5 arguments already)

I ran into pretty much all the above problems while prototyping the IAM support for token minting. 

To resolve these problems I propose a new set of HTTP client abstractions (`HttpClient` and `AuthorizedHttpClient`). My original plan was to use the [`axios`](https://www.npmjs.com/package/axios) library under the hood, but its support for handling timeouts [needs a bit of work](https://github.com/axios/axios/issues/459). So I came up with an implementation which borrows from both `axios` and our already existing `HttpRequestHandler` class. The resulting implementation:

1. Supports both HTTPS and HTTP
2. Exposes response headers alongside status and payload
3. Simplifies error handling -- HTTP errors are reported by throwing an `HttpError` which provides access to the response object. All other low-level errors result in a `FirebaseAppError`.
4. Supports any payload type
5. Fairly easy to test; new arguments can be added by extending the `HttpRequestConfig` type

This PR implements the new abstractions and provides unit tests. I will update the rest of the codebase to use them in a series of future PRs (work already underway in a separate branch).

PR #291, opened 2018-06-11
Migrating the auth module to use the new HTTP Client

hiranya911:
#280 introduced a new set of HTTP client abstractions. This PR migrates the auth module to use them.The resulting code is shorter (about 150 lines less) and easier to test.

### DIFF
--- a/src/utils/api-request.ts
+++ b/src/utils/api-request.ts
@@ -18,53 +18,63 @@ import {deepCopy} from './deep-copy';
 import {FirebaseApp} from '../firebase-app';
 import {AppErrorCodes, FirebaseAppError} from './error';
 import axios, {AxiosInstance, AxiosTransformer, AxiosResponse, AxiosError} from 'axios';
-
 import {OutgoingHttpHeaders} from 'http';
+
 import https = require('https');
 
 /** Http method type definition. */
-export type HttpMethod = 'GET' | 'POST' | 'DELETE';
+export type HttpMethod = 'GET' | 'POST' | 'PUT' | 'DELETE';
 /** API callback function type definition. */
 export type ApiCallbackFunction = (data: object) => void;
 
 export interface HttpRequest {
-  method: ('get' | 'post' | 'put' | 'delete');
+  method: HttpMethod;
   url: string;
   headers?: {[key: string]: string};
   data?: any;
   timeout?: number;
 }
 
-export class HttpResponse {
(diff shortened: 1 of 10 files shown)


src/utils/api-request.ts
```diff
@@ -111,12 +111,9 @@ class DefaultHttpResponse implements HttpResponse {
...
 export class HttpError extends Error {
-
-  public readonly response: HttpResponse;
-
-  constructor(resp: LowLevelResponse) {
-    super(`Server responded with status ${resp.status}.`);
-    this.response = new DefaultHttpResponse(resp);
+  constructor(public readonly response: HttpResponse) {
+    super(`Server responded with status ${response.status}.`);
+    Object.setPrototypeOf(this, HttpError.prototype);
```
bojeil-google: What does this actually do? You are setting prototype of HttpError to prototype of HttpError?

test/unit/utils.ts
```diff
@@ -84,3 +84,7 @@ export function responseFrom(data: any, status: number = 200, headers: any = {})
     text: JSON.stringify(data),
   };
 }
+
+export function errorFrom(data: any, status: number = 500): HttpError {
```
bojeil-google: If I recall, the default error http status for majority of errors above is 400. It is 500 here.

src/utils/api-request.ts
hiranya911: This is needed for the `instanceof` checks to work correctly: https://github.com/Microsoft/TypeScript/issues/13965

test/unit/utils.ts
hiranya911: I don't think it matters that much. The old test cases did not use any status codes, which implies that the auth module is agnostic to the actual HTTP status codes.

With this change we now have test cases that return both 500 and 400 errors. WDYT?

src/utils/api-request.ts
bojeil-google: Can you add a comment on this. There is no way I would have guessed this.

test/unit/utils.ts
bojeil-google: In the past there was no indication of http status code during test simulation. With this change, it is implied all simulated status codes are 500s. It is true Auth doesn't care in this case, but other services may. Note this could also change in the future. 500 could be handled differently.
Anyway, this is probably fine for now.

PR #301, opened 2018-07-06
Using the new HttpClient in messaging module

hiranya911:
Using the new `AuthorizedHttpClient` implementation in the FCM code.

This is a follow up of #280 and #291.

Also introducing the `isJson` field to the HTTP response interface, to make it easier to handle JSON-formatted responses.

### DIFF
--- a/src/messaging/messaging-api-request.ts
+++ b/src/messaging/messaging-api-request.ts
@@ -15,17 +15,15 @@
  */
 
 import {FirebaseApp} from '../firebase-app';
-import {HttpMethod, SignedApiRequestHandler} from '../utils/api-request';
-import {FirebaseError, FirebaseMessagingError, MessagingClientErrorCode} from '../utils/error';
+import {HttpMethod, AuthorizedHttpClient, HttpRequestConfig, HttpError} from '../utils/api-request';
+import {FirebaseMessagingError, MessagingClientErrorCode} from '../utils/error';
 
 import * as validator from '../utils/validator';
 
 // FCM backend constants
-const FIREBASE_MESSAGING_PORT = 443;
 const FIREBASE_MESSAGING_TIMEOUT = 10000;
 const FIREBASE_MESSAGING_HTTP_METHOD: HttpMethod = 'POST';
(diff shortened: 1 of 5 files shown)


src/messaging/messaging-api-request.ts
```diff
@@ -96,76 +94,70 @@ export class FirebaseMessagingRequestHandler {
...
+      timeout: FIREBASE_MESSAGING_TIMEOUT,
+    };
+    return this.httpClient.send(request).then((response) => {
       // Send non-JSON responses to the catch() below where they will be treated as errors.
-      if (typeof response === 'string') {
-        return Promise.reject({
-          error: response,
-          statusCode: 200,
-        });
+      if (!response.isJson) {
```
bojeil-google: I think it would be better to create a property or method to determine whether the response is valid, something like `isValid()` or `valid`.

src/utils/api-request.ts
```diff
@@ -108,6 +109,10 @@ class DefaultHttpResponse implements HttpResponse {
       `request: "${ this.request }."`,
     );
   }
+
+  get isJson(): boolean {
```
bojeil-google: If you need a method to check if a response is valid, I think it is better to give it another name. `isValid()`. or something like that. Also a `isSomething()` should be a method. If you need a property for this, call it something like `valid`.
Note JSCore style guide:
`Avoid using the 'is' prefix for boolean properties unless it's unclear from context that the property is a boolean.`

hiranya911: It's not a matter of validity. It's about whether the response is json or not. We have endpoints (e.g. Metadata server), where the valid response is plain text. It also seems at least some of our backend endpoints return html error messages. So we need a way to easily differentiate between json and non-json responses, so the client code can handle them easily.

src/messaging/messaging-api-request.ts
hiranya911: Renamed to `contentJson` and added some documentation. Didn't feel like naming this just `json`, since that seems to imply this returns a json object. wdyt?

PR #309, opened 2018-07-18
Upgraded to Firestore 0.15.4

hiranya911:
* Re-exporting the Firestore Timestamp type
* Updating affected tests

### DIFF
--- a/test/integration/firestore.spec.ts
+++ b/test/integration/firestore.spec.ts
@@ -78,7 +78,7 @@ describe('admin.firestore', () => {
       .then((snapshot) => {
         const data = snapshot.data();
         expect(data.timestamp).is.not.null;
-        expect(data.timestamp instanceof admin.firestore.Timestamp).is.true;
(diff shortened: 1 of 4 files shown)


test/integration/firestore.spec.ts
```diff
@@ -76,7 +78,7 @@ describe('admin.firestore', () => {
       .then((snapshot) => {
         const data = snapshot.data();
         expect(data.timestamp).is.not.null;
-        expect(data.timestamp instanceof Date).is.true;
+        expect(data.timestamp instanceof admin.firestore.Timestamp).is.true;
```
schmidt-sebastian: This is: `expect(data.timestamp).to.be.instanceOf(admin.firestore.Timestamp)`

hiranya911: Done. 

I believe the package-lock.json changes are because we define a lot of our dependencies loosely with wildcards. So everytime we regen the file, there are new versions to pick up.

Issue #313, opened 2018-07-22
FR: Add option for hash input order when importing users


jwngr:
#### Feature Request Description

Add a new `hash.inputOrder` option for the [`importUsers()`](https://firebase.google.com/docs/reference/admin/node/admin.auth.Auth#importUsers) auth method, inspired by the [`--hash-input-order` parameter of the Firebase CLI](https://firebase.google.com/docs/cli/auth).

It will accept a string value of `"SALT_FIRST"` or `"PASSWORD_FIRST"` and it will only apply to the `SHA512`, `SHA256`, `SHA1`, `MD5`, `HMAC_SHA512`, `HMAC_SHA256`, `HMAC_SHA1`, and `HMAC_MD5` algorithms.

#### Usage Example

```js
const importOptions = {
  hash: {
    algorithm: 'SHA256',
    rounds: 80,
    inputOrder: 'PASSWORD_FIRST',  // or 'SALT_FIRST'
  }
};

return admin.auth().importUsers(users, importOptions)
  .then((results) => {
    console.log('Users successfully imported!', results);
  })
  .catch((error) => {
    console.log('Failed to import users:', error);
  });
```

PR #314, opened 2018-07-22
Added hash input order option when importing users

jwngr:
Implements #313.

This PR includes the implementation of a new `hash.inputOrder` option for the `importUsers()` auth methods, with the relevant unit and integration tests and type definitions. See the issue for more information and a usage example.

Note that all unit tests pass on my local machine and all integration tests except `admin.auth importUsers() successfully imports users with MD5 to Firebase Auth w/ explicit password-first hash` pass as well. It is my understanding from the [`--hash-input-order` documentation](https://firebase.google.com/docs/cli/auth) that it should apply to the `MD5` algorithm as well. However, this integration test fails although the request being made to the Firebase Auth backend looks correct to me. It would be great if someone from the Firebase Auth team took a look into what is going wrong with that test.


bojeil-google:
Sorry for the delay @jwngr. The API review for your proposed Node.js changes is under way.

PR #318, opened 2018-07-25
Using the new HttpClient API in Credentials

hiranya911:
Using the new `HttpClient` API in Credentials module. With this I believe we are using the new API everywhere in the SDK (I will create a separate PR to remove the old code). I also managed to re-enable an old refresh token unit test that had been disabled for a long time.

This is a follow up to #280, #291, and #301

### DIFF
--- a/src/auth/credential.ts
+++ b/src/auth/credential.ts
@@ -216,11 +216,11 @@ function requestAccessToken(client: HttpClient, request: HttpRequestConfig): Pro
  * Implementation of Credential that uses a service account certificate.
  */
 export class CertCredential implements Credential {
-  private readonly certificate_: Certificate;
+  private readonly certificate: Certificate;
   private readonly httpClient: HttpClient;
 
   constructor(serviceAccountPathOrObject: string | object) {
-    this.certificate_ = (typeof serviceAccountPathOrObject === 'string') ?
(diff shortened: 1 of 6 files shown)


src/auth/credential.ts
```diff
@@ -186,75 +185,59 @@ export interface GoogleOAuthAccessToken {
...
+    );
   });
 }
 
 /**
  * Implementation of Credential that uses a service account certificate.
  */
 export class CertCredential implements Credential {
-  private certificate_: Certificate;
+  private readonly certificate_: Certificate;
```
bojeil-google: since you're already modifying this, let's remove the trailing underscore.

src/auth/credential.ts
```diff
@@ -295,11 +278,13 @@ export interface Credential {
  * Implementation of Credential that gets access tokens from refresh tokens.
  */
 export class RefreshTokenCredential implements Credential {
-  private refreshToken_: RefreshToken;
+  private readonly refreshToken_: RefreshToken;
```
bojeil-google: Same here. Let's remove trailing underscore.

src/auth/credential.ts

src/auth/credential.ts

PR #322, opened 2018-08-01
Upgraded Firestore version to 0.16.0

hiranya911:
Also dropping Node4 config from CI, as Firestore (and GCP) no longer supports it.


hiranya911:
Node4 related stuff have been sync'ed via #328

PR #328, opened 2018-08-03
Removing Node4 from the list of tested/supported frameworks

hiranya911:
Firestore and GCP have dropped Node4 support. It's time for us to join the club.

Related to #322


schmidt-sebastian:
I am surprised you don't have an "engines" entry in your package.json, similar to https://github.com/googleapis/nodejs-firestore/blob/master/package.json#L8

PR #332, opened 2018-08-07
Made Firestore and GCS Optional Deps

### LINKED ISSUE #295: Install without compiling gRPC module
* Operating System version: Amazon Linux 4.4.4
  * Firebase SDK version: 5.12.1
  * Library version: ?
  * Firebase Product: Messaging

#### Steps to reproduce:

`yarn add firebase-admin` or `npm install firebase-admin` will start to compile some gRPC native module with node-gyp since there aren't any pre-built binaries for this platform.

This requires a lot of memory, making it so we can't run our server on a t2.micro (512MB), and instead need to use t2.medium, just because it's trying to compile this native module.

Is there any way to use firebase messaging without needing to compile native module on install..?

_I worked around this by compiling it inside docker before deploying. It would be nice if these packages was split out as separate things so we didn't need to install everything including this native dependency when I just need the messaging parts_

hiranya911:
Since some users frequently encounter installation issues with GCP libraries (when they are not even using them), this PR re-declares those dependencies as `optional`. 

Partially resolves #295

### DIFF
--- a/src/firestore/firestore.ts
+++ b/src/firestore/firestore.ts
@@ -118,7 +118,8 @@ function initFirestore(app: FirebaseApp): Firestore {
     throw new FirebaseFirestoreError({
       code: 'missing-dependencies',
       message: 'Failed to import the Cloud Firestore client library for Node.js. '
-          + 'Make sure to install the "@google-cloud/firestore" npm package.',
+          + 'Make sure to install the "@google-cloud/firestore" npm package. '
+          + `Original error: ${err}`,
     });
   }
   return new firestoreDatabase(options);
--- a/src/firestore/firestore.ts
+++ b/src/firestore/firestore.ts
@@ -110,7 +110,16 @@ export function getFirestoreOptions(app: FirebaseApp): any {
 
 function initFirestore(app: FirebaseApp): Firestore {
   const options = getFirestoreOptions(app);
-  // Lazy-load the Firestore implementation here, which in turns loads gRPC.
-  const firestoreDatabase: typeof Firestore = require('@google-cloud/firestore');
+  let firestoreDatabase: typeof Firestore;
+  try {
+    // Lazy-load the Firestore implementation here, which in turns loads gRPC.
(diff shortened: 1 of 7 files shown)


src/firestore/firestore.ts
```diff
@@ -110,7 +110,16 @@ export function getFirestoreOptions(app: FirebaseApp): any {
 
 function initFirestore(app: FirebaseApp): Firestore {
   const options = getFirestoreOptions(app);
-  // Lazy-load the Firestore implementation here, which in turns loads gRPC.
-  const firestoreDatabase: typeof Firestore = require('@google-cloud/firestore');
+  let firestoreDatabase: typeof Firestore;
+  try {
+    // Lazy-load the Firestore implementation here, which in turns loads gRPC.
```
schmidt-sebastian: Do we need to do the same error checking for Storage?

hiranya911: We already have it in Storage: https://github.com/firebase/firebase-admin-node/blob/c6a4e222d8c60eeb03e22e997e0cc36d9f25d28d/src/storage/storage.ts#L61-L71

hiranya911: I made an improvement by including the original error in the returned error message.

src/storage/storage.ts
```diff
@@ -60,13 +60,13 @@ export class Storage implements FirebaseServiceInterface {
...
     try {
-      /* tslint:disable-next-line:no-var-requires */
       storage = require('@google-cloud/storage');
-    } catch (e) {
+    } catch (err) {
       throw new FirebaseError({
         code: 'storage/missing-dependencies',
         message: 'Failed to import the Cloud Storage client library for Node.js. '
-          + 'Make sure to install the "@google-cloud/storage" npm package.',
+          + 'Make sure to install the "@google-cloud/storage" npm package.'
```
schmidt-sebastian: Add space at end before merging. Thanks!

hiranya911: Done

PR #353, opened 2018-09-06
Setting noImplicitAny: true

hiranya911:
This has been a long standing open issue for us. I've gone through all sources and tests and fixed the implicit any inferences.


hiranya911:
Putting this on hold until we figure a solution for Node 6 and 7.

PR #368, opened 2018-10-15
Upgrades Google Cloud Storage & Firestore to latest versions

ovaris:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -153,9 +153,9 @@
       }
     },
     "@google-cloud/firestore": {
-      "version": "0.16.0",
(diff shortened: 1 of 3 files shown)


Steven4294:
can we get this merged pls

hiranya911:
Storage 2.x has breaking changes, and therefore we cannot upgrade to it in a patch release. Also it has its own typings, which means we have to drop the usage of `@types/google-cloud__storage` as well. We are discussing of ways to do this upgrade in the near future.

hiranya911:
Closing in favor of #437

PR #377, opened 2018-10-26
pass firebase version to firestore

Feiyang1:
pass firebase version to firestore


src/firestore/firestore.ts
```diff
@@ -63,7 +63,7 @@ export class FirestoreService implements FirebaseServiceInterface {
   }
 }
 
-export function getFirestoreOptions(app: FirebaseApp): any {
+export function getFirestoreOptions(app: FirebaseApp): Settings {
```
hiranya911: Can you add a test case as well? Or at least update the existing test cases.

https://github.com/firebase/firebase-admin-node/blob/master/test/unit/firestore/firestore.spec.ts

Feiyang1: done.

Issue #253, opened 2018-04-08
FR:Support proxy for local testing


perqin:
### [REQUIRED] Step 2: Describe your environment

  * Operating System version: `Linux archlinux 4.15.15-1-ARCH #1 SMP PREEMPT Sat Mar 31 23:59:25 UTC 2018 x86_64 GNU/Linux`
  * Library version: 5.12.0
  * Firebase Product: Cloud Messaging

### [REQUIRED] Step 3: Describe the problem

My local computer is behind proxy so the Node.js Admin SDK can't send the request to FCM server. It will be great if it supports HTTP proxy or proxy agent.

p2k:
You can try this package as a workaround until it is supported officially: https://github.com/np-maintain/global-tunnel

hiranya911:
We can probably support specifying an HTTP/S agent as an AppOption.

```
const tunnel = require('tunnel');

admin.initializeApp({
  httpAgent: tunnel.httpsOverHttps({
    proxy: {
      host: 'my.proxy.host',
      port: 3128,
    },
  }),
});
```

Users will have to install `tunnel` (or an equivalent library) and initialize the Admin SDK as shown above. Does that sound like a reasonable solution?

hiranya911:
#402 implements the support for specifying an `http.Agent` for the SDK. This use case should be supported now.

PR #388, opened 2018-11-12
Allow global http agent for firebase as a setting (#253)

Gyunikuchan:
## Discussion
As discussed in https://github.com/firebase/firebase-admin-node/issues/253

The global http agent is implemented as a static variable under the FirebaseNamespace.
I'm not proud of this solution, but there seems to be no clean way of doing this short of a bigger rewiring around HttpClient constructions.

## Usage
```
import * as admin from "firebase-admin";
const HttpsProxyAgent = require("https-proxy-agent");

export const initializeFirebaseAdmin = (serviceAccount: any) => {
	const proxyUrl = process.env.HTTPS_PROXY || process.env.https_proxy || process.env.HTTP_PROXY || process.env.http_proxy;
	if (proxyUrl) {
		admin.setHttpAgent(new HttpsProxyAgent(proxyUrl));
	}

	admin.initializeApp({
		credential: admin.credential.cert(serviceAccount),
	});
};
```


Gyunikuchan:
thx @googlebot

hiranya911:
I do like the simplicity of this approach. But I'm not sure if we want to expose any statics from the FirebaseNamespace. Will discuss this with the team and get back to you.

danpe:
Thank you @Gyunikuchan, great feature.
[global-tunnel](https://www.npmjs.com/package/global-tunnel-ng) use-case won't work for me, so I wanted to specify an agent for firebase-admin, I tried cloning your repo using `"firebase-admin": "git+https://github.com/Gyunikuchan/firebase-admin-node.git"` in my npm dependencies, but no luck when importing?

Error: Cannot find module 'firebase-admin'

Gyunikuchan:
> Thank you @Gyunikuchan, great feature.
> [global-tunnel](https://www.npmjs.com/package/global-tunnel-ng) use-case won't work for me, so I wanted to specify an agent for firebase-admin, I tried cloning your repo using `"firebase-admin": "git+https://github.com/Gyunikuchan/firebase-admin-node.git"` in my npm dependencies, but no luck when importing?
> 
> `Error: Cannot find module 'firebase-admin'`
> 
> I guess because it's not an official version release?
> @hiranya911 Can you please approve that :)

If you trust my repo or just want a quick insecure build, you may import from 
`"firebase-admin": "git+https://github.com/Gyunikuchan/firebase-admin-node.git#built"`

PR #393, opened 2018-11-21
add subtitle in messaging.ApsAlert payload

chemidy:
Add subtitle field in ApsAlert payload + tests


src/messaging/messaging.ts
```diff
@@ -41,6 +41,8 @@ const CAMELCASED_NOTIFICATION_PAYLOAD_KEYS_MAP = {
   clickAction: 'click_action',
   titleLocArgs: 'title_loc_args',
   titleLocKey: 'title_loc_key',
+  subtitleLocArgs: 'subtitle_loc_args',
+  subtitleLocKey: 'subtitle_loc_key',
```
hiranya911: This is for the legacy send operations. I don't think we should be adding new features to this API. Lets revert these two lines.

chemidy: ok done

PR #402, opened 2018-11-29
Support for specifying an http.Agent

hiranya911:
go/firebase-admin-node-proxy

Tested with `mitmproxy` and `tunnel2`:

```
    const tunnel = require('tunnel2');
    const proxyAgent = tunnel.httpsOverHttp({
      proxy: {
        host: 'localhost',
        port: 8080,
      }
    });
    const options = getOptions();
    options.httpAgent = proxyAgent;
    const app = admin.initializeApp(options, 'proxyApp');
    return admin.auth(app).createUser({}).then((user) => {
      return admin.auth(app).deleteUser(user.uid);
    }).catch((err) => {
      console.log(err);
    });
```

Sample request captured at `mitmproxy`:

```
2018-11-29 16:19:51 POST https://identitytoolkit.googleapis.com/v1/projects/my-project-id/accounts
                         ← 200 OK application/json 114b 497msDetail
X-Client-Version:  Node/Admin/6.3.0                                                                                                                                                         
Authorization:     Bearer xxxxxx
Content-Type:      application/json;charset=utf-8                                                                                                                                           
Content-Length:    2                                                                                                                                                                        
Host:              identitytoolkit.googleapis.com:443                                                                                                                                       
Connection:        close                                                                                                                                                                    
JSON                                                                                                                                                                                  [m:auto]
{}
```


test/unit/utils/api-request.spec.ts
```diff
@@ -450,6 +446,71 @@ describe('AuthorizedHttpClient', () => {
...
+      agentForApp = options.httpAgent;
+    });
+
+    afterEach(() => {
+      transportSpy.restore();
+      transportSpy = null;
+      return mockAppWithAgent.delete();
+    });
+
+    it('should use the HTTP agent set in request', () => {
```
Feiyang1: Is it any different than 'should use the HTTP agent set in request' above. They look pretty similar.

hiranya911: The first one was for the `HttpClient` class. This one's for `AuthorizedHttpClient`, which is a subclass of the former. The child class has the ability to use the agent configured at app-level, but also supports overriding it at request-level. These tests verify this.

danpe:
Hey @hiranya911, @Feiyang1 
Is there a planned merge/release date for that issue?

hiranya911:
@danpe nothing concrete. As soon as the code review is done.

PR #409, opened 2018-12-06
Implemented APNs Critical Sound Support

hiranya911:
go/apns-critical-sound

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -447,7 +447,7 @@ declare namespace admin.messaging {
   type Aps = {
     alert?: string | ApsAlert;
     badge?: number;
-    sound?: string;
+    sound?: string | CriticalSound;
     contentAvailable?: boolean;
     mutableContent?: boolean;
     category?: string;
@@ -468,6 +468,12 @@ declare namespace admin.messaging {
     actionLocKey?: string;
     launchImage?: string;
   };
+
+  type CriticalSound = {
+    critical?: boolean;
+    name?: string;
+    volume?: number;
+  }
 
   type Notification = {
     title?: string;
--- a/src/messaging/messaging.ts
+++ b/src/messaging/messaging.ts
@@ -348,7 +348,7 @@ function validateApsSound(sound: string | CriticalSound) {
       'apns.payload.aps.sound must be a string or a non-null object');
   }
 
-  const volume = (sound as CriticalSound).volume;
+  const volume = sound.volume;
   if (typeof volume !== 'undefined') {
     if (!validator.isNumber(volume)) {
       throw new FirebaseMessagingError(
--- a/src/messaging/messaging.ts
+++ b/src/messaging/messaging.ts
@@ -170,12 +170,18 @@ export interface ApnsPayload {
 export interface Aps {
   alert?: string | ApsAlert;
   badge?: number;
-  sound?: string;
+  sound?: string | CriticalSound;
   contentAvailable?: boolean;
   category?: string;
   threadId?: string;
   mutableContent?: boolean;
   [customData: string]: any;
+}
+
+export interface CriticalSound {
+  critical?: boolean;
+  name?: string;
+  volume?: number;
 }
 
 export interface ApsAlert {
@@ -299,6 +305,7 @@ function validateAps(aps: Aps) {
       MessagingClientErrorCode.INVALID_PAYLOAD, 'apns.payload.aps must be a non-null object');
   }
   validateApsAlert(aps.alert);
+  validateApsSound(aps.sound);
 
   const propertyMappings: {[key: string]: string} = {
     contentAvailable: 'content-available',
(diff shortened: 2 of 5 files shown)


src/index.d.ts
```diff
@@ -469,6 +469,12 @@ declare namespace admin.messaging {
     launchImage?: string;
   };
 
+  type CriticalSound = {
+    critical?: boolean;
```
Feiyang1: Isn't it redundant since it's of type CriticalSound and it should be always true?

src/index.d.ts
```diff
@@ -469,6 +469,12 @@ declare namespace admin.messaging {
...
+    name?: string;
```
Feiyang1: is it optional? What happens if you don't provide a name?

src/messaging/messaging.ts
```diff
@@ -332,6 +339,40 @@ function validateAps(aps: Aps) {
...
+function validateApsSound(sound: string | CriticalSound) {
+  if (typeof sound === 'undefined' || validator.isString(sound)) {
+    return;
+  } else if (!validator.isNonNullObject(sound)) {
+    throw new FirebaseMessagingError(
+      MessagingClientErrorCode.INVALID_PAYLOAD,
+      'apns.payload.aps.sound must be a string or a non-null object');
+  }
+
+  const volume = (sound as CriticalSound).volume;
```
Feiyang1: Does typescript not narrow it down to CriticalSound at this point? 
If it's the case, you can define `validator.isString` as an user defined type guard with signature:
`isString(input: any): input is String`

src/messaging/messaging.ts
```diff
@@ -332,6 +339,40 @@ function validateAps(aps: Aps) {
   }
 }
 
+function validateApsSound(sound: string | CriticalSound) {
```
Feiyang1: return type?

src/messaging/messaging.ts
```diff
@@ -332,6 +339,40 @@ function validateAps(aps: Aps) {
...
+    if (volume < 0 || volume > 1) {
+      throw new FirebaseMessagingError(
+        MessagingClientErrorCode.INVALID_PAYLOAD,
+        'apns.payload.aps.sound.volume must be in the interval [0, 1]');
+    }
+  }
+  const soundObject = sound as {[key: string]: any};
+  const key = 'critical';
+  const critical = soundObject[key];
+  if (typeof critical !== 'undefined' && critical !== 1) {
```
Feiyang1: We should not need to check `critical !== 1` since it's defined as a boolean.

src/messaging/messaging.ts
```diff
@@ -332,6 +339,40 @@ function validateAps(aps: Aps) {
...
+    if (critical === true) {
+      soundObject[key] = 1;
```
Feiyang1: The name of the function indicates it just validates the input and doesn't have a return value (void), but we modify the input directly. It's hard for people to understand the intention and this style should be avoid. I would rename it to something like `convertToApsSoundPayload` and return the modified object. At least I would put some comments to explain what's going on, but I really think we shouldn't modify input directly.

src/messaging/messaging.ts
hiranya911: JS developers often pass 1 for this kind of values, since APNs documentation says so. Our policy so far has been to silently allow it.

src/index.d.ts
hiranya911: APNs docs are not clear on that point: https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/generating_a_remote_notification#2990112

I believe it will default to system sound if none specified.

src/index.d.ts
hiranya911: Technically, following is allowed in APNs:

```
"sound": {
  "name": "default"
}
```

This is same as:

```
"sound": "default"
```

src/messaging/messaging.ts
Feiyang1: hmm, that's interesting. Any reason we deviate from the APNs definition? Can we just define critical as number?

src/index.d.ts
hiranya911: It can be false. We are using the APNs terminology here. So this is fine. Plus the API change has already been approved.

src/index.d.ts
hiranya911: Will make sure this gets documented in the published docs.

src/messaging/messaging.ts
hiranya911: Well, the only acceptable values for this field are 0 and 1. So it's really a bool that APNs exposes as an int for some reason. We have always exposed them as bools in our public API, but been flexible to accept ints as well. See the `Aps.mutableContent` property for a similar field.

src/messaging/messaging.ts
hiranya911: We have a bunch of un-exported helper functions like this in this module. They all directly modify the input. However, we guarantee that these methods receive a copy of the original input, so in-place modification is ok. 

I know it's not ideal, but I want to follow the existing pattern here. We can change them to not mutate the input in a separate PR.

src/messaging/messaging.ts
hiranya911: Done

hiranya911:
@Feiyang1 Made some of the changes, and responded to the others. PTAL.

Issue #411, opened 2018-12-06
Requests don't support localAddress parameter


NicholasMendes:
Describe your environment

  * Operating System version: MacOS Mojave 10.14.2 beta (18C52a)
  * Library version: **All** (6.1.0)
  * Firebase Product: Messaging

Describe the problem

Firebase-admin-node doesn't support **localAddress** parameter on http request. This is useful if you have limited internet connection and want to route where the requests are sent through.

Relevant Code:

The following code can be aded to solve this issue.

file: **api-request.js**
line: **158**

```
var options = {
            hostname: parsed.hostname,
            port: parsed.port,
            path: parsed.path,
            method: config.method,
            headers: headers,
            localAddress: undefined,
        };
if (process.env.LOCAL_ADDRESS) {
            options.localAddress = process.env.LOCAL_ADDRESS;
}
```

hiranya911:
Is this supported by any other GCP libraries? We are somewhat reluctant to support options that are not supported in the GCP ecosystem.

hiranya911:
Closing this as won't fix for now. The exact use case and the value addition is unclear.

PR #416, opened 2018-12-10
Add option to specify local address for http requests.

NickMeloMendes:
This commit allows the user to specify an address to route the http requests, this is useful if for example your network connection is limited and only available through a given address.

Closes: #411


NickMendes:
I signed it!

gilsonjustjr (approved):
Looks good.

PR #418, opened 2018-12-10
Add integration tests for Project Management service.

nbegley:


test/integration/project-management.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+/**
+ * Ensures that an Android app owned by these integration tests exist. If not one will be created.
+ *
+ * @return {Promise<AndroidApp>} Android app owned by these integration tests.
+ */
+function ensureAndroidApp(): Promise<admin.projectManagement.AndroidApp> {
+  return admin.projectManagement().listAndroidApps()
+      .then((apps) => Promise.all(apps.map((app) => app.getMetadata())))
+      .then((metadatas) => {
+        const metadataOwnedByTest =
```
hiranya911:
```
if (metadatas.some((metadata) => isIntegrationTestApp(metadata.packageName))) {

}
```

test/integration/project-management.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+/**
+ * Ensures that an iOS app owned by these integration tests exist. If not one will be created.
+ *
+ * @return {Promise<IosApp>} iOS app owned by these integration tests.
+ */
+function ensureIosApp(): Promise<admin.projectManagement.IosApp> {
+  return admin.projectManagement().listIosApps()
+      .then((apps) => Promise.all(apps.map((app) => app.getMetadata())))
+      .then((metadatas) => {
+        const metadataOwnedByTest =
```
hiranya911: Use `Array.some()`.

test/integration/project-management.spec.ts
nbegley: metadataOwnedByTest is reused inside the if statement, so I'd rather keep it in its current form instead of performing an array scan twice.

test/integration/project-management.spec.ts
nbegley: Same comment as above.

PR #421, opened 2018-12-14
Fixes optional fields in UserRecord types to be optional.

bojeil-google:
Fixes optional fields in UserRecord types to be optional.
References will be updated to reflect that too.


hiranya911:
Also update the CHANGELOG

PR #434, opened 2019-01-02
Running a Node.js environment check at installation

### LINKED ISSUE #41: Update Install Script to Print Warning if Used in a Client Environment
The `npm install` command should warn the user if the admin SDK is installed in a client environment.

hiranya911:
Resolves #41 

Check inspired by https://www.npmjs.com/package/detect-node

### DIFF
--- a/env-check.js
+++ b/env-check.js
@@ -1,0 +1,19 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const isNode = (Object.prototype.toString.call(global.process) === '[object process]');
+if (!isNode) {
+  const message = `
+======== WARNING! ========
+
+firebase-admin appears to have been installed in an abnormal Node.js environment.
+This package should only be used in server-side or backend Node.js environments,
+and should not be used in web browsers or other client-side environments.
+
+Use the Firebase JS SDK for client-side Firebase integrations:
+
+https://firebase.google.com/docs/web/setup
+`;
+  console.log(message);
+}
--- a/package.json
+++ b/package.json
@@ -18,7 +18,8 @@
     "test:coverage": "nyc npm run test:unit",
     "lint:src": "tslint --format stylish -p tsconfig.json",
     "lint:unit": "tslint -c tslint-test.json --format stylish test/unit/*.ts test/unit/**/*.ts",
-    "lint:integration": "tslint -c tslint-test.json --format stylish test/integration/*.ts"
+    "lint:integration": "tslint -c tslint-test.json --format stylish test/integration/*.ts",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -18,7 +18,8 @@
     "test:coverage": "nyc npm run test:unit",
     "lint:src": "tslint --format stylish -p tsconfig.json",
     "lint:unit": "tslint -c tslint-test.json --format stylish test/unit/*.ts test/unit/**/*.ts",
-    "lint:integration": "tslint -c tslint-test.json --format stylish test/integration/*.ts"
+    "lint:integration": "tslint -c tslint-test.json --format stylish test/integration/*.ts",
+    "postinstall": "node ./env-check.js"
```
Feiyang1: We should do a run time check instead of the installation time check. This check will always return true since we are running it with node.

hiranya911: Done

env-check.js
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const isNode = (Object.prototype.toString.call(global.process) === '[object process]');
```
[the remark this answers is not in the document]
hiranya911: Added a comment

env-check.js
```diff
@@ -0,0 +1,19 @@
...
+if (!isNode) {
+  const message = `
+======== WARNING! ========
+
+firebase-admin appears to have been installed in an abnormal Node.js environment.
```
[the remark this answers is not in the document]
hiranya911: After considering feedback from you and @Feiyang1 I've implemented this check as a runtime check. We will log an error if the package is imported into a client-side runtime.

src/index.ts
```diff
@@ -16,11 +16,21 @@
...
 import * as firebase from './default-namespace';
 
-// Register the Database service
-// For historical reasons, the database code is included as minified code and registers itself
-// as a side effect of requiring the file.
-/* tslint:disable:no-var-requires */
-// require('./database/database');
-/* tslint:enable:no-var-requires */
+// Only Node.js has a process variable that is of [[Class]] process
+if (Object.prototype.toString.call(global.process) !== '[object process]') {
```
Feiyang1: Not sure If the package will ever run in an env without `global` global. 
If yes, we should do `if (global && Object.prototype.toString.call(global.process) !== '[object process]')`

hiranya911: Looks like browsers don't have it. Therefore I'm checking for `process` instead of `global` now. Check adapted from https://github.com/iliakan/detect-node/blob/master/index.js

Issue #179, opened 2018-01-10
Differentiating between invalid and expired tokens when verifying


simenbrekken:
### Environment

  * Operating System version: MacOS High Sierra 10.13.1 (17B1003)
  * Firebase SDK version: 5.7.0
  * Library version: 5.7.0
  * Firebase Product: auth

### Description

There's currently no robust way to detect if an id token is invalid or has expired when performing verification via `auth().verifyIdToken` other than checking if the actual `error.message` contains the string "expired". 

The [`verifyIdToken` implementation of `token-generator`](https://github.com/firebase/firebase-admin-node/blob/master/src/auth/token-generator.ts#L227-L236) correctly detects expired tokens by checking for `error.name === 'TokenExpiredError'` but this error code is discarded as it's rejected as an `FirebaseAuthError` with the `AuthClientErrorCode.INVALID_ARGUMENT` error code.

Would it be possible to to add something like `AuthClientErrorCode.TOKEN_EXPIRED` or expose the original `error.name` as `error.errorInfo.name`?

I'd be glad to contribute a PR for any of the above propsals.

hiranya911:
Sounds reasonable to me. @bojeil-google what do you think? Shall we introduce a new `auth/id-token-expired` error code? (we already have `auth/id-token-revoked`, so there's precedent for something like this)

bojeil-google:
In general, I think it is a good idea but I think this could be a breaking changes. Developers may be expecting the current error code for both. If we add a new error code, they would basically miss it (they won't be catching it). I suggest we keep this until the next major version bump.

simenbrekken:
While waiting for the next major version bump would it be possible as per my second suggestion to surface the original `TokenExpiredError` as `error.errorInfo.name` in the meantime?

hiranya911:
It's going to be hard to justify an API change that will become unnecessary a few months later. What if we modify the error message (`errorInfo.message`) to some constant value for the moment?

simenbrekken:
How about `${message} (auth/id-token-expired)` so the same constant can be reused when it's introduced as a separate error type?

hiranya911:
Yes, that sounds reasonable. @simenbrekken would you be able to give us a patch?

simenbrekken:
Certainly. I'll have a patch ready this weekend.

vlapo:
Is this change still on roadmap? I see major release is out and this issue is closed but `auth/id-token-expired` error code is not implemented.

[...]

PR #442, opened 2019-01-15
Implemented errors codes for expired ID tokens and session cookies

hiranya911:
Related to #179

### DIFF
--- a/test/unit/auth/token-verifier.spec.ts
+++ b/test/unit/auth/token-verifier.spec.ts
@@ -40,9 +40,8 @@ chai.use(chaiAsPromised);
 
 const expect = chai.expect;
 
-
 const ONE_HOUR_IN_SECONDS = 60 * 60;
-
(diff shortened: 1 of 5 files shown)


test/unit/auth/token-verifier.spec.ts
```diff
@@ -46,13 +47,18 @@ const ONE_HOUR_IN_SECONDS = 60 * 60;
 /**
  * Returns a mocked out success response from the URL containing the public keys for the Google certs.
  *
+ * @param {string=} path URL path to which the mock request should be made. If not specified, defaults
+ *   to the URL path of ID token public key certificates.
  * @return {Object} A nock response object.
  */
-function mockFetchPublicKeys(): nock.Scope {
+function mockFetchPublicKeys(path?: string): nock.Scope {
```
bojeil-google: You can directly set a default value from here.

hiranya911: Done

PR #444, opened 2019-01-17
Handle special case of application default credentials location

yinzara:
When the GOOGLE_APPLICATION_CREDENTIALS environment variable is pointed to
the refresh token file created by 'gcloud auth application-default login',
Firebase admin would error as it tried to parse it as a certificate.
This fix doesn't attempt to parse the file as a certificate if the
variable points to the refresh token file and instead just attempts
refresh token file parsing

### Discussion

### Testing

### API Changes

    can discuss it together.


hiranya911:
ADC protocol only requires that the variable point to a service account json file: https://cloud.google.com/docs/authentication/production

> First, ADC checks to see if the environment variable GOOGLE_APPLICATION_CREDENTIALS is set. If the variable is set, ADC uses the service account file that the variable points to. The next section describes how to set the environment variable.

However, in practice many implementations are lenient about this, and allow both service accounts and refresh token files. I'm ok with making the Node.js implementation do the same.

But right now there's a CI failure, and we cannot proceed until that is fixed.

yinzara:
Thank you. This is important in environments where you may pass your local application default credentials into a docker container in a dev environment but pass a service account key in production.

I didn't realize you aren't able to test refresh token validity in the CI environment until I read the other tests better.  Since my test requires a valid refresh token, I'm going to add the exclusion that calls  "skipAndLogWarningIfNoGcloud" which should fix the build but that means this test will only run on a local environment.

yinzara:
I changed my mind :) I rewrote the test case so that it doesn't actually validate the credentials with the google API but instead just verifies that the parsed credentials are a RefreshTokenCredentials and uses the test example credentials instead of real ones.

yinzara:
My test fix did not work so I changed them to use Sinon to mock the reading of the file

src/auth/credential.ts
```diff
@@ -355,7 +355,10 @@ export class ApplicationDefaultCredential implements Credential {
   private credential_: Credential;
 
   constructor(httpAgent?: Agent) {
-    if (process.env.GOOGLE_APPLICATION_CREDENTIALS) {
+    if (process.env.GOOGLE_APPLICATION_CREDENTIALS &&
```
hiranya911: If we are going to implement this we should be consistent with other languages. Specifically, there's no reason to special case `GCLOUD_CREDENTIALS_PATH`. Rather, we should do something like:

```
if (process.env.GOOGLE_APPLICATION_CREDENTIALS) {
  const fileContent = readFile(process.env.GOOGLE_APPLICATION_CREDENTIALS);
  if (fileContent.type === 'authorized_user') {
    // create refresh token credential
  } else if (fileContent.type === 'service_account') {
   // create cert credential  
  } else {
   // throw error
  }
}
```

test/unit/auth/credential.spec.ts
```diff
@@ -395,6 +395,36 @@ describe('Credential', () => {
         privateKey: mockCertificateObject.private_key,
       });
     });
+
+    it('should parse valid token if app def creds point to default refresh token loc', () => {
+      process.env.GOOGLE_APPLICATION_CREDENTIALS = GCLOUD_CREDENTIAL_PATH;
+
+      let readFileSyncStub;
```
hiranya911: There's an `fsStub` variable already in scope for these tests. See if we can use that here.

yinzara:
Updated.  Quite a bit more code to handle that and still keep all existing test cases passing but I did it and added a few more to cover the other code paths I added.

src/auth/credential.ts
```diff
@@ -355,20 +355,63 @@ export class ApplicationDefaultCredential implements Credential {
...
-      this.credential_ = new CertCredential(serviceAccount, httpAgent);
-      return;
-    }
 
-    // It is OK to not have this file. If it is present, it must be valid.
-    const refreshToken = RefreshToken.fromPath(GCLOUD_CREDENTIAL_PATH);
-    if (refreshToken) {
-      this.credential_ = new RefreshTokenCredential(refreshToken, httpAgent);
-      return;
+    if (typeof process.env.GOOGLE_APPLICATION_CREDENTIALS === 'string' &&
```
hiranya911: This doesn't need to be this complex. You only have to modify the first if block in the existing constructor. Rest of the constructor can remain unchanged. Something like:

```
constructor(httpAgent?: Agent) {
    if (process.env.GOOGLE_APPLICATION_CREDENTIALS) {
      this.credential_ = credentialFromFile(process.env.GOOGLE_APPLICATION_CREDENTIALS, httpAgent);
      return;
    }

    // Rest is unchanged
}

function credentialFromFile(filePath: string, httpAgent?: Agent): Credential {
  const credentialsFile = readCredentialFile(filePath);
  if (typeof credentialsFile !== 'object') {
    throw new FirebaseAppError(
      AppErrorCodes.INVALID_CREDENTIAL,
      'Failed to parse contents of the credentials file as an object',
    );
  }
  if (credentialsFile.type === 'service_account') {
    return new CertCredential(credentialsFile, httpAgent);
  }
  if (credentialsFile.type === 'authorized_user') {
    return new RefreshTokenCredential(credentialsFile, httpAgent);
  }
  throw new FirebaseAppError(
    AppErrorCodes.INVALID_CREDENTIAL,
    'Invalid contents in the credentials file',
  );
}

function readCredentialFile(filePath: string): {[key: string]: any} {
  if (typeof filePath !== 'string') {
      throw new FirebaseAppError(
        AppErrorCodes.INVALID_CREDENTIAL,
        'Failed to parse credentials file: TypeError: path must be a string',
      );
  }
  let fileText: string;
  try {
    fileText = fs.readFileSync(filePath, 'utf8');
  } catch (error) {
    throw new FirebaseAppError(
      AppErrorCodes.INVALID_CREDENTIAL,
      `Failed to read credentials from file ${filePath}: ` + error,
    );
  }
  try {
    return JSON.parse(fileText);
  } catch (error) {
    throw new FirebaseAppError(
      AppErrorCodes.INVALID_CREDENTIAL,
      'Failed to parse contents of the credentials file as an object: ' + error,
    );
  }
}
```

test/unit/auth/credential.spec.ts
```diff
@@ -395,6 +408,24 @@ describe('Credential', () => {
         privateKey: mockCertificateObject.private_key,
       });
     });
+
+    it('should parse valid token if app def creds point to default refresh token loc', () => {
```
hiranya911: Lets spell out words in full: `should create a RefreshTokenCredential if application default credentials variable point to gcloud credentials path`.

yinzara:
Changes made as requested

test/unit/auth/credential.spec.ts
```diff
@@ -395,6 +408,24 @@ describe('Credential', () => {
...
+      const adc = new ApplicationDefaultCredential();
+      const c = adc.getCredential();
+      expect(c).is.instanceOf(RefreshTokenCredential);
+      expect(c).to.have.property('refreshToken').that.includes({
+        clientId: MOCK_REFRESH_TOKEN_CONFIG.client_id,
+        clientSecret: MOCK_REFRESH_TOKEN_CONFIG.client_secret,
+        refreshToken: MOCK_REFRESH_TOKEN_CONFIG.refresh_token,
+        type: MOCK_REFRESH_TOKEN_CONFIG.type,
+      });
+      // tslint:disable-next-line:no-unused-expression
```
hiranya911: Is this required?

yinzara: Odd. My dev environment gives an error on that line if I don't have it, but tslint command line doesn't complain.  Either way I've removed it, updated the test case name as suggested and updated the CHANGELOG

PR #448, opened 2019-01-24
Fixing error handling in Firebase credentials

hiranya911:
The error handling in Firebase credentials has been broken due to a bug introduced while migrating to the `HttpClient` API. Specifically, when credentials fail to fetch an access token, they no longer include the original server error in the exceptions thrown. This prevents more higher level classes like `FirebaseApp` from reporting verbose errors.

This PR fixes the issue, and adds test cases to verify the behavior.

### DIFF
--- a/src/auth/credential.ts
+++ b/src/auth/credential.ts
@@ -198,16 +198,23 @@ function requestAccessToken(client: HttpClient, request: HttpRequestConfig): Pro
     }
     return json;
   }).catch((err) => {
-    const message = getErrorMessage(err);
-    throw new FirebaseAppError(AppErrorCodes.INVALID_CREDENTIAL, message);
+    throw new FirebaseAppError(AppErrorCodes.INVALID_CREDENTIAL, getErrorMessage(err));
   });
 }
 
+/**
+ * Constructs a human-readable error message from the given Error.
+ */
 function getErrorMessage(err: Error): string {
   const detail: string = (err instanceof HttpError) ? getDetailFromResponse(err.response) : err.message;
   return `Error fetching access token: ${detail}`;
 }
 
+/**
+ * Extracts details from the given HTTP error response, and returns a human-readable description. If
+ * the response is JSON-formatted, looks up the error and error_description fields sent by the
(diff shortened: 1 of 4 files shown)


src/auth/credential.ts
```diff
@@ -190,28 +190,36 @@ export interface GoogleOAuthAccessToken {
...
-    throw new FirebaseAppError(
-      AppErrorCodes.INVALID_CREDENTIAL,
-      `Failed to parse access token response: ${err.toString()}`,
-    );
+    const message = getErrorMessage(err);
+    throw new FirebaseAppError(AppErrorCodes.INVALID_CREDENTIAL, message);
   });
 }
 
+function getErrorMessage(err: Error): string {
```
bojeil-google: Add function javadoc.

src/auth/credential.ts
```diff
@@ -190,28 +190,36 @@ export interface GoogleOAuthAccessToken {
...
+  const detail: string = (err instanceof HttpError) ? getDetailFromResponse(err.response) : err.message;
+  return `Error fetching access token: ${detail}`;
+}
+
+function getDetailFromResponse(response: HttpResponse): string {
```
bojeil-google: Add function javadoc.

src/auth/credential.ts
```diff
@@ -190,28 +190,36 @@ export interface GoogleOAuthAccessToken {
...
-      return json;
     }
+    return json;
   }).catch((err) => {
-    throw new FirebaseAppError(
-      AppErrorCodes.INVALID_CREDENTIAL,
-      `Failed to parse access token response: ${err.toString()}`,
-    );
+    const message = getErrorMessage(err);
+    throw new FirebaseAppError(AppErrorCodes.INVALID_CREDENTIAL, message);
```
bojeil-google: You can directly pass `getErrorMessage(err)` instead of `message` interim variable.

src/auth/credential.ts

src/auth/credential.ts

src/auth/credential.ts

hiranya911:
Made the suggested changes and updated the changelog. Over to @bojeil-google for another look.

PR #460, opened 2019-02-19
Improve typings of `UpdateRequest` interface

### LINKED ISSUE #459: Invalid typings of `auth.updateUser`
### Describe your environment

  * Operating System version: Darwin Kernel Version 18.2.0
  * Firebase SDK version: 7.0.0
  * Library version: ?
  * Firebase Product: auth

### Describe the problem

According to the official [documentation](https://firebase.google.com/docs/reference/admin/node/admin.auth.UpdateRequest) and a common sense some of the fields in `UpdateRequest` are nullable. It's not reflected in typings.

#### Steps to reproduce:

Try to compile a code with the following statement.

#### Relevant Code:

```ts
auth().updateUser(userId, { phoneNumber: null })
```

merlinnot:
Resolves #459

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
--
+- [fixed] Improved typings of `UpdateRequest` interface to support deletion of
+  properties.
 
 # v7.0.0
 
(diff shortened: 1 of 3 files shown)


bojeil-google:
Thanks for the fix!

merlinnot:
Done, seems weird it's not generated, but I guess there is a reason for it :) Also added a changelog entry.

PR #468, opened 2019-03-04
Ability to install package from GitHub

schmidt-sebastian:
Without the "prepare" rule, the TypeScript compile is not invoked when installing this package directly from GitHub.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   },
   "scripts": {
     "build": "gulp build",
+    "prepare": "npm run build",
     "lint": "run-p lint:src lint:unit lint:integration",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -10,6 +10,7 @@
   },
   "scripts": {
     "build": "gulp build",
+    "prepare": "npm run build",
```
hiranya911: It looks like this will get called after `npm install` automatically. Should we therefore change `createReleaseTarball.sh` to not call it explicitly?

schmidt-sebastian: Yes, that's true. The tarball contains the .js file even without the addition `npm run build`.

Issue #504, opened 2019-04-12
FR: [Database] Allow getting and setting the security rules


jsayol:
### [REQUIRED] Step 2: Describe your environment

  * Operating System version: Linux 5.0.7
  * Firebase SDK version: n/a
  * Library version: 7.2.0
  * Firebase Product: database

### [REQUIRED] Step 3: Describe the problem
It would be useful to be able to get and set the database security rules through the Admin SDK.

#### Workaround
I'm currently calling the REST API to read and write the database rules, obtaining the admin access token from `admin.app().INTERNAL.getToken()` (although I could also do it properly by getting an access token through a service account file, for example).

#### Steps to reproduce:
n/a

#### Relevant Code:
n/a

jsayol:
Off the top of my head, a possible solution may be to add a couple of methods `.getRules()` and `.setRules()` to `admin.database()`, which would internally call the REST API to get and set the security rules.

If you're open to adding a couple such methods to the Admin SDK API, I'd be happy to look further into it and submit a PR.

hiranya911:
Hi @jsayol. Interestingly enough we have had some early discussions about such a feature. Are you more interested in RTDB rules or Firestore rules?

hiranya911:
@schmidt-sebastian wdyt?

jsayol:
Hey @hiranya911, that's good to hear! I'm more interested in RTDB rules at the moment, but I would also like to be able to do it for Firestore rules. I just haven't started looking into what the current options are for the latter.

jsayol:
I've been thinking about and my initial suggestion of putting this under `admin.database()` and that might not be the best idea.

The first call to that method instantiates a `Database` instance, which in turn will have several side effects like establishing a network connection to the database backend. In cases where you just want to get or set the rules, you don't need that connection to happen so it might be a waste of resources. Plus, if these methods were to be present in the `Database` object, then they would either need to be implemented in the `@firebase/database` package (which would bloat the client SDK unnecessarily) or they would need to be monkey-patched by the admin SDK [after instantiating the database](https://github.com/firebase/firebase-admin-node/blob/3537d0b227a08c576c34a3097348c662c481db79/src/database/database.ts#L70). None of those options sound very appealing to me.

A better option in my opinion would be to add this as part of the `ProjectManagement` namespace. While rule handling is not technically part of the Project Management API, I think it does fit well in there for the purpose of the admin SDK. So, just an idea:

```js
// For RTDB
[...]

PR #507, opened 2019-04-21
WIP: Implement Rules API

jsayol:
### Discussion

See this issue: https://github.com/firebase/firebase-admin-node/issues/504

### Testing

:heavy_check_mark: All existing tests keep passing.
:heavy_check_mark: Added the necessary tests (unit and integration) for the new features.

### API Changes

As discussed in the linked issue, these new types and methods are introduced to the `admin.projectManagement` namespace:

```ts
type RulesService = 'firestore' | 'storage' | 'database';

interface Ruleset {
  /**
   * The UUID of the ruleset.
   */
  id: string;

  /**
   * Timestamp in ISO 8601 format.
   */
  createTime: string;
}

interface RulesetWithFiles extends Ruleset {
  /**
   * Array of ruleset files.
   */
  files: RulesetFile[];
}

interface RulesetFile {
  /**
   * Name of the file.
   */
  name: string;

  /**
   * The content of the file (the actual rules).
   */
  content: string;
}

/**
 * Possible filters when listing releases. All optional and can be combined.
 */
interface ListRulesReleasesFilter {
  releaseName?: string;
  rulesetName?: string;
  testSuiteName?: string;
}

/**
 * The result of listing rules releases.
 */
interface ListRulesReleasesResult {
  releases: RulesRelease[];
  pageToken?: string;
}

interface RulesRelease {
  /**
   * The name of the release.
   * This will usually be either `firebase.storage` or `cloud.firestore`.
   */
  name: string;

  /**
   * ID of the ruleset associated with the release.
   */
  rulesetId: string;

  /**
   * Timestamp in ISO 8601 format.
   */
  createTime: string;

  /**
   * Timestamp in ISO 8601 format.
   */
  updateTime: string;
}

/**
 * The result of listing rulesets.
 */
interface ListRulesetsResult {
  rulesets: Ruleset[];
  pageToken?: string;
}

interface ProjectManagement {
  /**
   * Gets the current rules for the service:
   *   - For RTDB that's whatever `.settings/rules.json` returns.
   *   - For Firestore/Storage it's the contents of the ruleset for the release
   *     associated with that service.
   */
  getRules(service: admin.projectManagement.RulesService): Promise<string>;

  /**
   * Sets the new rules to be used with the service:
   *   - For RTDB it PUTs them to `.settings/rules.json`.
   *   - For Firestore/Storage it creates a new ruleset with the specified
   *     content and updates/creates the appropriate release for the
   *     service with that ruleset.
   */
  setRules(
    service: admin.projectManagement.RulesService,
    content: string,
  ): Promise<void>;

  /**
   * Like `setRules()` but reads the rules content from a file.
   */
  setRulesFromFile(
    service: admin.projectManagement.RulesService,
    filePath: string,
  ): Promise<void>;

  /**
   * Gets the list of rules releases for the project.
   *
   * It optionally accepts an object specifying filters to use.
   *
   * The maximum number of rulesets to return is determined by the optional
   * `pageSize` argument. Defaults to 10, maximum is 100 (according to API docs).
   */
  listRulesReleases(
    filter?: admin.projectManagement.ListRulesReleasesFilter,
    pageSize?: number,
    pageToken?: string,
  ): Promise<admin.projectManagement.ListRulesReleasesResult>;

  /**
   * Gets the given rules release.
   */
  getRulesRelease(
    name: string,
  ): Promise<admin.projectManagement.RulesRelease>;

  /**
   * Creates a new rules release with the given name and associated to
   * the given ruleset name.
   */
  createRulesRelease(
    name: string,
    rulesetId: string,
  ): Promise<admin.projectManagement.RulesRelease>;

  /**
   * Updates the ruleset name associated with an existing rules release.
   */
  updateRulesRelease(
    name: string,
    rulesetId: string,
  ): Promise<admin.projectManagement.RulesRelease>;

  /**
   * Deletes the given rules release.
   */
  deleteRulesRelease(name: string): Promise<void>;

  /**
   * Gets the list of rulesets for the project. The Rulesets only contain
   * metadata (`name` and `createTime`), not the actual files.
   *
   * The maximum number of rulesets to return is determined by the optional
   * `pageSize` argument. Defaults to 10, maximum is 100 (according to API docs).
   *
   * It optionally accepts a pageToken returned from a previous call, in order
   * to get the next set of results if there's more.
   */
  listRulesets(
    pageSize?: number,
    pageToken?: string,
  ): Promise<admin.projectManagement.ListRulesetsResult>;

  /**
   * Gets the ruleset by its id. The returned Ruleset contains its files.
   */
  getRuleset(
    rulesetId: string,
  ): Promise<admin.projectManagement.RulesetWithFiles>;

  /**
   * Creates a new ruleset with the given files.
   */
  createRuleset(
    files: RulesetFile[],
  ): Promise<admin.projectManagement.RulesetWithFiles>;

  /**
   * Deletes the given rules ruleset.
   */
  deleteRuleset(rulesetId: string): Promise<void>;
}
```

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -796,6 +796,45 @@ declare namespace admin.projectManagement {
     getConfig(): Promise<string>;
   }
 
+  type RulesService = 'firestore' | 'storage' | 'database';
+
+  interface Ruleset {
+    name: string;
+    createTime: string;
+  }
+
+  interface RulesetWithFiles extends Ruleset {
+    files: RulesetFile[];
+  }
+
+  interface RulesetFile {
+    name: string;
+    content: string;
+  }
+
+  interface ListRulesReleasesFilter {
+    releaseName?: string;
+    rulesetName?: string;
+    testSuiteName?: string;
+  }
+
+  interface ListRulesReleasesResult {
+    releases: RulesRelease[];
+    pageToken?: string;
+  }
+
+  interface RulesRelease {
+    name: string;
+    rulesetName: string;
+    createTime: string;
+    updateTime: string;
+  }
+
+  interface ListRulesetsResult {
+    rulesets: Ruleset[];
+    pageToken?: string;
+  }
+
   interface ProjectManagement {
     app: admin.app.App;
 
@@ -807,8 +846,39 @@ declare namespace admin.projectManagement {
     createAndroidApp(
         packageName: string, displayName?: string): Promise<admin.projectManagement.AndroidApp>;
     createIosApp(bundleId: string, displayName?: string): Promise<admin.projectManagement.IosApp>;
-    getDatabaseRules(): Promise<string>;
-    setDatabaseRules(rules: string): Promise<void>;
+    getRules(service: admin.projectManagement.RulesService): Promise<string>;
+    setRules(
+      service: admin.projectManagement.RulesService,
+      content: string,
+    ): Promise<void>;
+    setRulesFromFile(
+      service: admin.projectManagement.RulesService,
+      filePath: string,
+    ): Promise<void>;
+    listRulesReleases(
+      filter?: admin.projectManagement.ListRulesReleasesFilter,
+      maxResults?: number,
+      pageToken?: string,
+    ): Promise<admin.projectManagement.ListRulesReleasesResult>;
+    getRulesRelease(name: string): Promise<admin.projectManagement.RulesRelease>;
+    createRulesRelease(
+      name: string,
+      rulesetName: string,
+    ): Promise<admin.projectManagement.RulesRelease>;
+    updateRulesRelease(
+      name: string,
+      rulesetName: string,
+    ): Promise<admin.projectManagement.RulesRelease>;
+    deleteRulesRelease(name: string): Promise<void>;
+    listRulesets(
+      maxResults?: number,
+      pageToken?: string,
+    ): Promise<admin.projectManagement.ListRulesetsResult>;
+    getRuleset(name: string): Promise<admin.projectManagement.RulesetWithFiles>;
+    createRuleset(
+      files: RulesetFile[],
+    ): Promise<admin.projectManagement.RulesetWithFiles>;
+    deleteRuleset(name: string): Promise<void>;
   }
 }
 
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -857,7 +857,7 @@ declare namespace admin.projectManagement {
     ): Promise<void>;
     listRulesReleases(
       filter?: admin.projectManagement.ListRulesReleasesFilter,
-      maxResults?: number,
+      pageSize?: number,
       pageToken?: string,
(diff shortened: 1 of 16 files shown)


hiranya911:
Hi @jsayol. I haven't looked at the code yet. But to keep the conversation up-to-date:

1. This still needs to go through our API approval process, which will take several weeks.
2. I don't think we want to expose any of the releases stuff in the API (at least not yet). So my vote is for removing all the `*RulesRelease()` APIs.
3. I would also recommend the following name changes:
   - `Ruleset` -> `RulesetInfo` (not a huge fan of this name, but don't have a better one yet)
   - `RulesetWithFile` -> `Ruleset`

jsayol:
> This still needs to go through our API approval process, which will take several weeks.

Got it, no problem! I figured that even if the public-facing API needs to change, all the underlying code will still be the same so I went ahead and implemented it, to give us a starting point.
Even if it's eventually decided not to include this in the SDK, it was only a couple of days of work and it gave me a lot of insight into the world of Firebase security rules, so I learned quite a bit from it :smile: 

> I don't think we want to expose any of the releases stuff in the API (at least not yet). So my vote is for removing all the *RulesRelease() APIs.

I understand. I guess the same is true for some of the Ruleset-related methods, then? `listRulesets()` and `getRuleset()` still make sense on their own, but creating rulesets doesn't make much sense if you can't associate them to a release, and trying to delete them without knowing whether they're currently associated with a release is also a bit pointless, since it might fail.

That being said, removing those methods would leave unsolved the issue that @rachelmyers brought up, where a developer might hit the quota limit for creating rulesets. In that case they would need to head to the console and manually delete them.

A possible solution would be to add a method to delete the X oldest unused rulesets. I ended up implementing [such a method](https://github.com/jsayol/firebase-admin-node/blob/045382b3870062d185dca82f52003f8d35a323f4/test/integration/project-management.spec.ts#L719-L792) for the integration tests (otherwise running them a few times could quickly run into the quota limit) so it should be trivial to add that to the SDK. The only downside is that this method can potentially trigger quite a few network requests.

> I would also recommend the following name changes:
> * `Ruleset` -> `RulesetInfo` (not a huge fan of this name, but don't have a better one yet)
> * `RulesetWithFile` -> `Ruleset`

I like a change along those lines, and I agree that none of these options are ideal. Something I was thinking about earlier is that `Ruleset` could actually be converted into its own class, with `id` and `createTime` as fields and a `getFiles()` method. So a `Ruleset` object would implement this interface:

```ts
interface Ruleset {
  id: string;
  createTime: string;
  getFiles(): Promise<RulesetFile[]>;
}
```

This would completely eliminate the need to duplicate the types with `Ruleset`/`RulesetWithFile` or `RulesetInfo`/`Ruleset`.

hiranya911:
RTDB rules management API was implemented in #595. Firestore and Storage rules management APIs will be implemented in the future.

PR #537, opened 2019-05-23
Add readonly to interfaces properties in index.d.ts

TrCaM:
### Description

- Add readonly to all properties of types and interfaces declared in index.d.ts

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -33,9 +33,9 @@ declare namespace admin {
   }
 
   interface ServiceAccount {
-    readonly projectId?: string;
-    readonly clientEmail?: string;
-    readonly privateKey?: string;
+    projectId?: string;
+    clientEmail?: string;
+    privateKey?: string;
   }
 
   interface GoogleOAuthAccessToken {
@@ -44,13 +44,13 @@ declare namespace admin {
   }
 
   interface AppOptions {
-    readonly credential?: admin.credential.Credential;
-    readonly databaseAuthVariableOverride?: Object;
-    readonly databaseURL?: string;
-    readonly serviceAccountId?: string;
-    readonly storageBucket?: string;
-    readonly projectId?: string;
-    readonly httpAgent?: Agent;
+    credential?: admin.credential.Credential;
+    databaseAuthVariableOverride?: Object;
+    databaseURL?: string;
+    serviceAccountId?: string;
+    storageBucket?: string;
+    projectId?: string;
+    httpAgent?: Agent;
   }
 
   var SDK_VERSION: string;
@@ -121,17 +121,17 @@ declare namespace admin.auth {
   }
 
   interface UpdateRequest {
-    readonly disabled?: boolean;
-    readonly displayName?: string | null;
-    readonly email?: string;
-    readonly emailVerified?: boolean;
-    readonly password?: string;
-    readonly phoneNumber?: string | null;
-    readonly photoURL?: string | null;
+    disabled?: boolean;
+    displayName?: string | null;
+    email?: string;
+    emailVerified?: boolean;
+    password?: string;
+    phoneNumber?: string | null;
+    photoURL?: string | null;
   }
 
   interface CreateRequest extends UpdateRequest {
-    readonly uid?: string;
+    uid?: string;
   }
 
   interface DecodedIdToken {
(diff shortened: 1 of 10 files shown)


src/index.d.ts
```diff
@@ -85,76 +85,76 @@ declare namespace admin.app {
...
+    readonly providerData: admin.auth.UserInfo[];
+    readonly passwordHash?: string;
+    readonly passwordSalt?: string;
+    readonly customClaims?: Object;
+    readonly tokensValidAfterTime?: string;
 
     toJSON(): Object;
   }
 
   interface UpdateRequest {
```
hiranya911: Can't make this readonly. Developers should be able to incrementally construct an UpdateRequest if they wish to.

```
const  req: UpdateRequest = {};
req.uid = 'alice'; // This will fail if uid is readonly
```

src/index.d.ts
```diff
@@ -20,37 +20,37 @@ import {Agent} from 'http';
...
   }
 
   interface GoogleOAuthAccessToken {
-    access_token: string;
-    expires_in: number;
+    readonly access_token: string;
+    readonly expires_in: number;
   }
 
   interface AppOptions {
```
hiranya911: Can't make this readonly.

src/index.d.ts
```diff
@@ -85,76 +85,76 @@ declare namespace admin.app {
...
+    readonly disabled?: boolean;
+    readonly displayName?: string | null;
+    readonly email?: string;
+    readonly emailVerified?: boolean;
+    readonly password?: string;
+    readonly phoneNumber?: string | null;
+    readonly photoURL?: string | null;
   }
 
   interface CreateRequest extends UpdateRequest {
```
hiranya911: Can't make this readonly.

src/index.d.ts
```diff
@@ -163,113 +163,113 @@ declare namespace admin.auth {
 
 
   interface UserImportOptions {
```
hiranya911: Shouldn't be readonly

src/index.d.ts
```diff
@@ -163,113 +163,113 @@ declare namespace admin.auth {
...
+    readonly passwordHash?: Buffer;
+    readonly passwordSalt?: Buffer;
   }
 
   interface SessionCookieOptions {
-    expiresIn: number;
+    readonly expiresIn: number;
   }
 
   interface ActionCodeSettings {
```
hiranya911: Shouldn't be readonly

src/index.d.ts
```diff
@@ -453,220 +453,220 @@ declare namespace admin.database.ServerValue {
 }
 
 type BaseMessage = {
```
hiranya911: Shouldn't be readonly. Same goes for all the types referenced here, and their child types (e.g. AndroidConfig -> AndroidNotification)

TrCaM:
@nbegley, @hiranya911, I reverted the readonly on the types that are used as input arguments for functions, thus allow user to modify those objects if necessary. Please check if my update is reasonable, and if I need to add/remove readonly on some other fields.
Thanks

src/index.d.ts
```diff
@@ -748,29 +748,29 @@ declare namespace admin.firestore {
...
   interface InstanceId {
-    app: admin.app.App;
+    readonly app: admin.app.App;
 
     deleteInstanceId(instanceId: string): Promise<void>;
   }
 }
 
 declare namespace admin.projectManagement {
   interface ShaCertificate {
```
TrCaM: `ShaCertificate` is used as input in `addShaCertificates()` and `deleteShaCertificate()`, but I still keep the readonly here as I don't think user will change the object after initializing it

hiranya911: @nbegley do you concur?

nbegley: I concur - `ShaCertificate`s are intended to be immutable.

nbegley:
We'll probably have to consider this a breaking change. Even though users shouldn't be relying on any of these fields being mutable, it's possible that some of them may be.

hiranya911:
I agree @nbegley. On a second thought, I'm even reluctant to do this change for output types. Somebody may have written code like this:

```
const user = await admin.auth().getUser(uid);
user.customClaims = {foo: 'bar'};
await admin.auth().setCustomUserClaims(uid, user.customClaims);
```

Having said that I do believe this is a good change, and something we ought to do at some point (possibly for the next major release). @TrCaM can you create an issue for this, so we will remember to revisit it in the future?

TrCaM:
Sure, I can add an issue

hiranya911:
Thanks. We will track this requirement via the internal issue you've reported, and get it done for the next major release.

PR #539, opened 2019-05-24
Multi-tenancy changes to shared Auth API requests

bojeil-google:
Defines BaseFirebaseAuthRequestHandler class for sending Auth requests related to user management APIs and SAML/OIDC config mgmt APIs, link generation, etc.

Defines FirebaseAuthRequestHandler which extends the base class for project level only calls and which will also be extended to include tenant mgmt APIs.
Defines FirebaseTenantRequestHandler which extends the base class for tenant level only calls.

Unit tests have been modified to run tests on both subclasses. 99% of the unit test logic is the same. It has been tweaked to accept a common interface handler. 2 tests are run for each. One for FirebaseAuthRequestHandler and another for FirebaseTenantRequestHandler.

### DIFF
--- a/src/auth/auth-api-request.ts
+++ b/src/auth/auth-api-request.ts
@@ -120,6 +120,37 @@ class AuthResourceUrlBuilder {
 }
 
 
+/** Tenant aware resource builder utility. */
+class TenantAwareAuthResourceUrlBuilder extends AuthResourceUrlBuilder {
+  /**
+   * The tenant aware resource URL builder constructor.
+   *
+   * @param {string} projectId The resource project ID.
+   * @param {string} version The endpoint API version.
+   * @param {string} tenantId The tenant ID.
+   * @constructor
+   */
+  constructor(protected projectId: string, protected version: string, protected tenantId: string) {
+    super(projectId, version);
+    // Inject tenant path in URL.
+    this.urlFormat = this.urlFormat.replace(
+        'projects/{projectId}', 'projects/{projectId}/tenants/{tenantId}');
+  }
+
+  /**
+   * Returns the resource URL corresponding to the provided parameters.
+   *
+   * @param {string=} api The backend API name.
+   * @param {object=} params The optional additional parameters to substitute in the
+   *     URL path.
+   * @return {string} The corresponding resource URL.
+   */
+  public getUrl(api?: string, params?: object) {
+    return utils.formatString(super.getUrl(api, params), {tenantId: this.tenantId});
+  }
+}
+
+
 /**
  * Validates a providerUserInfo object. All unsupported parameters
  * are removed from the original request. If an invalid field is passed
@@ -205,6 +236,7 @@ function validateCreateEditRequest(request: any, uploadAccountRequest: boolean =
     phoneNumber: true,
     customAttributes: true,
     validSince: true,
+    tenantId: true,
     passwordHash: uploadAccountRequest,
     salt: uploadAccountRequest,
     createdAt: uploadAccountRequest,
@@ -216,6 +248,10 @@ function validateCreateEditRequest(request: any, uploadAccountRequest: boolean =
     if (!(key in validKeys)) {
       delete request[key];
     }
+  }
+  if (typeof request.tenantId !== 'undefined' &&
+      !validator.isNonEmptyString(request.tenantId)) {
+    throw new FirebaseAuthError(AuthClientErrorCode.INVALID_TENANT_ID);
   }
   // For any invalid parameter, use the external key name in the error description.
   // displayName should be a string.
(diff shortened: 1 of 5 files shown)


src/auth/auth-api-request.ts
```diff
@@ -205,6 +236,7 @@ function validateCreateEditRequest(request: any, uploadAccountRequest: boolean =
     phoneNumber: true,
     customAttributes: true,
     validSince: true,
+    tenantId: true,
```
wuyanna: What does the boolean value mean here?

src/auth/auth-api-request.ts
```diff
@@ -217,6 +249,10 @@ function validateCreateEditRequest(request: any, uploadAccountRequest: boolean =
       delete request[key];
     }
   }
+  if (typeof request.tenantId !== 'undefined' &&
+      !validator.isNonEmptyString(request.tenantId)) {
```
wuyanna: Shouldn't the check be the mismatch between tenantId in the request and the tenantId in the user?

src/auth/auth-api-request.ts
```diff
@@ -120,6 +120,37 @@ class AuthResourceUrlBuilder {
...
+   *
+   * @param {string} projectId The resource project ID.
+   * @param {string} version The endpoint API version.
+   * @param {string} tenantId The tenant ID.
+   * @constructor
+   */
+  constructor(protected projectId: string, protected version: string, protected tenantId: string) {
+    super(projectId, version);
+    // Inject tenant path in URL.
+    this.urlFormat = this.urlFormat.replace(
```
hiranya911: It looks like `urlFormat` in the parent class is really just a constant. Can we just define a constant for this as well?

src/auth/auth-api-request.ts
```diff
@@ -633,10 +669,10 @@ const LIST_INBOUND_SAML_CONFIGS = new ApiSettings('/inboundSamlConfigs', 'GET')
 /**
  * Class that provides the mechanism to send requests to the Firebase Auth backend endpoints.
  */
-export class FirebaseAuthRequestHandler {
-  private readonly httpClient: AuthorizedHttpClient;
-  private readonly authUrlBuilder: AuthResourceUrlBuilder;
-  private readonly projectConfigUrlBuilder: AuthResourceUrlBuilder;
+export class BaseFirebaseAuthRequestHandler {
```
hiranya911: Can we make this class abstract and rename it to `AbstractAuthRequestHandler`? 

`getAuthUrlBuilder()` and `getProjectConfigUrlBuilder()` could be abstract methods of the class.

src/auth/auth-api-request.ts
```diff
@@ -1323,3 +1368,49 @@ export class FirebaseAuthRequestHandler {
...
   }
 }
+
+
+/**
+ * Utility for sending requests to Auth server that are Auth instance related. This includes user and
+ * tenant management related APIs. This extends the BaseFirebaseAuthRequestHandler class and defines
+ * additional tenant management related APIs.
+ */
+export class FirebaseAuthRequestHandler extends BaseFirebaseAuthRequestHandler {
```
hiranya911: My suggestions for names would be `AuthRequestHandler` and `TenantAwareAuthRequestHandler`.

bojeil-google: Done

src/auth/auth-api-request.ts
bojeil-google: Done.

src/auth/auth-api-request.ts
wuyanna: There probably won't be a mismatch tenant ID backend error. For UploadAccount you can always rely on backend returning error messages in the response for specific failed users. For the other endpoints like SetAccountInfo and SignupNewUsers, the backend won't be able to return the error since it only receives one tenant_id from the request as I mentioned in a separate email thread. Adding the client side check in a follow-up change sounds good to me.

bojeil-google: I see. Yeah I will add the mismatch check in a follow up.

src/auth/auth-api-request.ts
```diff
@@ -633,10 +671,11 @@ const LIST_INBOUND_SAML_CONFIGS = new ApiSettings('/inboundSamlConfigs', 'GET')
...
  */
-export class FirebaseAuthRequestHandler {
-  private readonly httpClient: AuthorizedHttpClient;
-  private readonly authUrlBuilder: AuthResourceUrlBuilder;
-  private readonly projectConfigUrlBuilder: AuthResourceUrlBuilder;
+export abstract class AbstractAuthRequestHandler {
+  protected readonly projectId: string;
+  protected readonly httpClient: AuthorizedHttpClient;
+  protected authUrlBuilder: AuthResourceUrlBuilder;
+  protected projectConfigUrlBuilder: AuthResourceUrlBuilder;
```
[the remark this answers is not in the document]
hiranya911: You can call the abstract methods `newAuthUrlBuilder()` and `newProjectConfigUrlBuilder()` if that sounds better.

PR #541, opened 2019-05-28
Add AppMetadata interface, declared unimplemented methods to ProjectManagement

TrCaM:
### Description
Extend project scoped operations in the NodeJS Admin SDK with the addition of `AppMetadata` interface and 2 new methods into `ProjectManagement` class (unimplemented):
- `listAppMetadata()` : Lists summary of all apps belongs to the project
- `setDisplayName()`: Update display name of the project

### In this pull request
- Declare `AppMetadata` interface in AppMetadata.ts
- Move `AndroidAppMetadata` and `IosAppMetadata` declaration into AppMetadata.ts
- Make `AndroidAppMetadata` and `IosAppMetadata` inherited from AppMetadata
- Update existing code and tests to match the change
- Update types in index.d.ts (readonly will be added in a separate pr)
- Declare `listAppMetadata()` and `setDisplayName()` which currently just throw errors.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -761,10 +761,10 @@ declare namespace admin.projectManagement {
     resourceName?: string;
   }
 
-  class AppMetadata {
-    public appId: string;
-    public displayName: string;
-    public platform: AppPlatform;
+  interface AppMetadata {
+    appId: string;
+    displayName?: string;
+    platform: AppPlatform;
   }
 
   enum AppPlatform {
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -765,6 +765,8 @@ declare namespace admin.projectManagement {
     appId: string;
     displayName?: string;
     platform: AppPlatform;
+    projectId: string;
+    resourceName: string;
   }
 
   enum AppPlatform {
@@ -775,9 +777,12 @@ declare namespace admin.projectManagement {
 
   interface AndroidAppMetadata extends AppMetadata {
     platform: AppPlatform.ANDROID;
-    resourceName: string;
-    projectId: string;
     packageName: string;
+  }
+
+  interface IosAppMetadata extends AppMetadata {
+    platform: AppPlatform.IOS;
+    bundleId: string;
   }
 
   interface AndroidApp {
@@ -789,13 +794,6 @@ declare namespace admin.projectManagement {
     addShaCertificate(certificateToAdd: ShaCertificate): Promise<void>;
     deleteShaCertificate(certificateToRemove: ShaCertificate): Promise<void>;
     getConfig(): Promise<string>;
-  }
-
-  interface IosAppMetadata extends AppMetadata {
-    platform: AppPlatform.IOS;
-    resourceName: string;
-    projectId: string;
-    bundleId: string;
   }
 
   interface IosApp {
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -243,7 +243,7 @@ declare namespace admin.auth {
     callbackURL?: string;
     enableRequestSigning?: boolean;
   }
-  
+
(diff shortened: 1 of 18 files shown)


TrCaM:
@hiranya911, I have a question for `AppMetadata` - Should we simply add the `platform` field to AndroidAppMetadata with the field's type requiring it to be set to `ANDROID`? Or should we convert AndroidAppMetadata to a class, and give it a constructor like below?

```typescript
interface AppMetadata {
  public readonly appId: string;
  public readonly displayName: string | null;
  public readonly platform: AppPlatform;
}

class AndroidAppMetadata implements AppMetadata {
  constructor(
      appId: string,
      displayName: string,
      readonly resourceName: string,
      readonly projectId: string,
      readonly packageName: string) {
    super(appId, displayName, AppPlatform.ANDROID);
  }
}
```

And would either of those options be considered a breaking change in the Admin SDK? There _should_ be no reason for users of the SDK to ever construct an instance of AndroidAppMetadata themselves, but our public interface makes it possible to do so.

hiranya911:
Is there an API proposal for these changes? 

I don't think what you've proposed above will actually work. Because `AppMetadata` is an interface, you cannot initialize its properties in `AndroidAppMetadata` by calling `super()`. You probably need to make `AppMetadata` a class for that to work.

nbegley:
@hiranya911, I'd like to re-pose @TrCaM's question after replacing `interface AppMetadata` with `class AppMetadata`. I'll send you a link to the private API proposal.

Basically, the question is "Can we change `AndroidAppMetadata` and `IosAppMetadata` from interfaces to "classes that inherit from `AppMetadata`" without that being a breaking change?

src/project-management/AppMetadata.ts
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+export interface AppMetadata {
+  readonly appId: string;
+  readonly displayName: string;
+  readonly platform: AppPlatform;
+}
+
+export enum AppPlatform {
```
hiranya911: Nit: Lets move this to the top, so the parent and child classes are arranged next to each other.

src/project-management/AppMetadata.ts
```diff
@@ -0,0 +1,41 @@
+/*!
```
hiranya911: File should be renamed to match the current convention in use. app-metadata.ts or similar.

src/index.d.ts
```diff
@@ -761,10 +761,21 @@ declare namespace admin.projectManagement {
...
+
+  enum AppPlatform {
+    PLATFORM_UNKNOWN = 'PLATFORM_UNKNOWN',
+    IOS = 'IOS',
+    ANDROID = 'ANDROID',
+  }
+
+  interface AndroidAppMetadata extends AppMetadata {
+    platform: AppPlatform.ANDROID;
+    resourceName: string;
```
nbegley: `resourceName` is a common field, so we should be able to include that in `AppMetadata`

src/project-management/app-metadata.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export interface AppMetadata {
+  readonly appId: string;
+  readonly displayName: string;
+  readonly platform: AppPlatform;
+}
+
+export interface AndroidAppMetadata extends AppMetadata {
+  readonly platform: AppPlatform.ANDROID;
+  readonly resourceName: string;
```
nbegley: (same for this file)

src/index.d.ts
TrCaM: Also add `projectId` to `AppMetadata`

src/project-management/app-metadata.ts
TrCaM: Also add `projectId` to `AppMetadata`

PR #546, opened 2019-05-31
Implement listAppMetadata() and setDisplayName() of ProjectManagement class

TrCaM:
### Description
Extend project scoped operations in the NodeJS Admin SDK with the addition of `AppMetadata` interface and 2 new methods into `ProjectManagement` class (unimplemented):
- `listAppMetadata()` : Lists summary of all apps belongs to the project
- `setDisplayName()`: Update display name of the project

### In this pull request
- Implement `listAppMetadata()` and `setDisplayName()` in ProjectManagement class and their Api requests
- Unit tests completed
- Integration tests completed

### DIFF
--- a/src/project-management/project-management.ts
+++ b/src/project-management/project-management.ts
@@ -24,11 +24,6 @@ import { IosApp } from './ios-app';
 import { ProjectManagementRequestHandler, assertServerResponse } from './project-management-api-request';
 import { AppMetadata, AppPlatform } from './app-metadata';
 
-const APP_PLATFORM_MAP: {[key: string]: AppPlatform} = {
-  ANDROID: AppPlatform.ANDROID,
-  IOS: AppPlatform.IOS,
-};
-
 /**
  * Internals of a Project Management instance.
  */
@@ -184,13 +179,16 @@ export class ProjectManagement implements FirebaseServiceInterface {
           validator.isNonEmptyString(appJson.platform),
           responseData,
           `"apps[].platform" field must be present in the listAppMetadata() response data.`);
-      return {
+      const metadata: AppMetadata = {
         appId: appJson.appId,
-        displayName: appJson.displayName,
-        platform: APP_PLATFORM_MAP[appJson.platform] || AppPlatform.PLATFORM_UNKNOWN,
+        platform: (AppPlatform as any)[appJson.platform] || AppPlatform.PLATFORM_UNKNOWN,
         projectId: this.projectId,
         resourceName: appJson.name,
       };
+      if (appJson.displayName) {
+        metadata.displayName = appJson.displayName;
+      }
+      return metadata;
     });
   }
 
--- a/src/project-management/project-management.ts
+++ b/src/project-management/project-management.ts
@@ -159,9 +159,8 @@ export class ProjectManagement implements FirebaseServiceInterface {
   /**
    * Update display name of the project
    */
-  public setDisplayName(displayName: string): Promise<void> {
-    throw new FirebaseProjectManagementError(
-      'service-unavailable', 'This service is not available');
+  public setDisplayName(newDisplayName: string): Promise<void> {
+    return this.requestHandler.setDisplayName(this.resourceName, newDisplayName);
   }
 
   private transformResponseToAppMetadata(responseData: any): AppMetadata[] {
--- a/src/project-management/project-management.ts
+++ b/src/project-management/project-management.ts
@@ -22,6 +22,7 @@ import * as validator from '../utils/validator';
 import { AndroidApp, ShaCertificate } from './android-app';
 import { IosApp } from './ios-app';
 import { ProjectManagementRequestHandler, assertServerResponse } from './project-management-api-request';
+import { AppMetadata } from './AppMetadata';
 
 /**
  * Internals of a Project Management instance.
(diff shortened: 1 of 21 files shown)


src/project-management/project-management.ts
```diff
@@ -152,16 +157,41 @@ export class ProjectManagement implements FirebaseServiceInterface {
...
+          validator.isNonEmptyString(appJson.appId),
+          responseData,
+          `"apps[].appId" field must be present in the listAppMetadata() response data.`);
+      assertServerResponse(
+          validator.isNonEmptyString(appJson.platform),
+          responseData,
+          `"apps[].platform" field must be present in the listAppMetadata() response data.`);
+      return {
+        appId: appJson.appId,
+        displayName: appJson.displayName,
```
TrCaM: @nbegley , @hiranya911, This makes `displayName` appearing as `undefined` if `displayName` is not specified for the app. Thus for [unit tests](https://github.com/firebase/firebase-admin-node/pull/546/files#diff-d9f4c407f99decc820eed61f77904d19R529) I hadto add a `displayName: undefined` to the expected result for it to pass.
If there's any good way to not include `displayName` if it's `undefined`, I'm happy to go for it

src/project-management/project-management.ts
```diff
@@ -152,16 +157,41 @@ export class ProjectManagement implements FirebaseServiceInterface {
...
+        platform: APP_PLATFORM_MAP[appJson.platform] || AppPlatform.PLATFORM_UNKNOWN,
```
TrCaM: @nbegley , @hiranya911 Is there a better way to transform the `platform` string returned by the API response to the `AppPlatform` enum?

hiranya911: Doesn't `AppPlatform[appJson.platform]` work (assuming enum values match exactly to the values returned by the API)? Might need an any cast somewhere to make the compiler happy.

src/project-management/project-management.ts
hiranya911: Something like:

```
const value = {
  appId: appJson.appId,
  // other fields except displayName
};
if (appJson.displayName) {
  value.displayName = appJson.displayName;
}
return value;
```

TrCaM: I will go for this. Thanks @hiranya911.

src/project-management/project-management.ts
TrCaM: @hiranya911, that's exactly the reason why I have to use this map to avoid complains from ts compiler. Thanks for the suggestion

hiranya911: Try `(AppPlatform as any)[appJson.platform] || AppPlatform.PLATFORM_UNDEFINED`

src/project-management/project-management.ts
hiranya911: In that case how about this?

```
const value: {[key: string]: any} = {
  appId: appJson.appId,
  // other fields except displayName
};
if (appJson.displayName) {
  value.displayName = appJson.displayName;
}
return value as AppMetadata;
```

nbegley: Using `any` here removes type safety, so in the event something changes, like a new field being added to AppMetadata, the transpiler wont detect that the field is missing from the AppMetadata being returned here.

eg. We'd want this code to show an error that `field1` is not set, but there is no error:
```
interface MyInterface {
    field1: string;
    field2: number;
}

const value: {[key: string]: any} = {};
value.field2 = 1;

let typed : MyInterface = value as MyInterface;
```

So I'd actually prefer to remove `readonly` from all fields in `AppMetadata` and use Hiranya's original suggestion.

TrCaM: @hiranya911 Can we have your opinion on removing `readonly`  in `AppMetadata`?

hiranya911: I don't feel strongly about it either way. You'd only be removing `readonly` from the internal implementation of `AppMetadata`. You can continue to keep `readonly` in the typings published in the d.ts file.

nbegley: Okay, let's move forward with that then.

TrCaM: Thanks @hiranya911. I will move on with removing `readonly` solution

TrCaM: Updated

src/project-management/project-management.ts
TrCaM: Updated

test/unit/project-management/project-management.spec.ts
```diff
@@ -380,19 +388,174 @@ describe('ProjectManagement', () => {
...
+        {
+          appId: VALID_LIST_APP_METADATA_API_RESPONSE.apps[0].appId,
+          displayName: VALID_LIST_APP_METADATA_API_RESPONSE.apps[0].displayName,
+          platform: AppPlatform.ANDROID,
+          projectId: mocks.projectId,
+          resourceName: RESOURCE_NAME_ANDROID,
+        },
+        {
+          appId: VALID_LIST_APP_METADATA_API_RESPONSE.apps[1].appId,
+          displayName: VALID_LIST_APP_METADATA_API_RESPONSE.apps[1].displayName,
```
nbegley: Can you update this test to make exactly 1 of the apps have no display name, to test your latest change?

PR #551, opened 2019-06-04
Defines TenantAwareAuth

bojeil-google:
Defines TenantAwareAuth and its user management APIs, email action link APIs, OIDC/SAML provider config mgmt APIs.

### DIFF
--- a/src/auth/auth-api-request.ts
+++ b/src/auth/auth-api-request.ts
@@ -238,7 +238,8 @@ function validateCreateEditRequest(request: any, uploadAccountRequest: boolean =
     phoneNumber: true,
     customAttributes: true,
     validSince: true,
-    tenantId: true,
+    // Pass tenantId only for uploadAccount requests.
+    tenantId: uploadAccountRequest,
     passwordHash: uploadAccountRequest,
     salt: uploadAccountRequest,
     createdAt: uploadAccountRequest,
@@ -474,6 +475,12 @@ export const FIREBASE_AUTH_SET_ACCOUNT_INFO = new ApiSettings('/accounts:update'
         AuthClientErrorCode.INTERNAL_ERROR,
         'INTERNAL ASSERT FAILED: Server request is missing user identifier');
     }
+    // Throw error when tenantId is passed in POST body.
+    if (typeof request.tenantId !== 'undefined') {
+      throw new FirebaseAuthError(
+          AuthClientErrorCode.INVALID_ARGUMENT,
+          '"tenantId" is an invalid "UpdateRequest" property.');
+    }
     validateCreateEditRequest(request);
   })
   // Set response validator.
@@ -504,6 +511,12 @@ export const FIREBASE_AUTH_SIGN_UP_NEW_USER = new ApiSettings('/accounts', 'POST
         AuthClientErrorCode.INVALID_ARGUMENT,
         `"validSince" cannot be set when creating a new user.`,
       );
+    }
+    // Throw error when tenantId is passed in POST body.
+    if (typeof request.tenantId !== 'undefined') {
+      throw new FirebaseAuthError(
+          AuthClientErrorCode.INVALID_ARGUMENT,
+          '"tenantId" is an invalid "CreateRequest" property.');
     }
     validateCreateEditRequest(request);
   })
@@ -913,15 +926,6 @@ export abstract class AbstractAuthRequestHandler {
         new FirebaseAuthError(
           AuthClientErrorCode.INVALID_ARGUMENT,
           'Properties argument must be a non-null object.',
-        ),
-      );
-    }
-
-    if (properties.hasOwnProperty('tenantId')) {
-      return Promise.reject(
-        new FirebaseAuthError(
-          AuthClientErrorCode.INVALID_ARGUMENT,
-          'Tenant ID cannot be modified on an existing user.',
         ),
       );
     }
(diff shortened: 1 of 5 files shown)


src/auth/auth.ts
```diff
@@ -599,11 +603,153 @@ class BaseAuth {
...
+   */
+  public importUsers(
+      users: UserImportRecord[], options?: UserImportOptions): Promise<UserImportResult> {
+    // Verify users have matching tenantIds if provided.
+    // When undefined tenant ID is provided (implicitly set the tenant ID of the TenantAwareAuth),
+    // the Auth server will automatically set the tenant ID.
+    users.forEach((user: UserImportRecord, index: number) => {
+      if (validator.isNonEmptyString(user.tenantId) &&
+          user.tenantId !== this.tenantId) {
+        throw new FirebaseAuthError(AuthClientErrorCode.MISMATCHING_TENANT_ID);
```
wuyanna: Do you need to tell which user has mismatched tenant id in the error message?

bojeil-google: I moved the check to the Auth API request level and added a more detailed custom message informing the developer of the index of the user and the mismatching tenant ID. I also added a check to throw an invalid argument error when tenant ID is provided in POST body for `signupNewUser` and `setAccountInfo` so you can safely rely on the tenant ID in the URL.

src/auth/auth.ts
```diff
@@ -599,11 +603,126 @@ class BaseAuth {
...
+  public readonly tenantId: string;
+
+  /**
+   * The TenantAwareAuth class constructor.
+   *
+   * @param {Auth} auth The Auth instance that created this tenant.
+   * @param tenantId The corresponding tenant ID.
+   * @constructor
+   */
+  constructor(private readonly auth: Auth, tenantId: string) {
```
hiranya911: Looks like it makes sense for this constructor to just accept `FirebaseApp` instead of `Auth`.

src/auth/auth.ts
```diff
@@ -599,11 +603,126 @@ class BaseAuth {
...
+    super(
+        utils.getProjectId(auth.app),
+        new TenantAwareAuthRequestHandler(auth.app, tenantId),
+        cryptoSignerFromApp(auth.app));
+    utils.addReadonlyGetter(this, 'tenantId', tenantId);
```
hiranya911: Just mark the constructor argument as `public readonly`

src/auth/auth.ts
```diff
@@ -599,11 +603,126 @@ class BaseAuth {
...
+}
+
+
 /**
  * Auth service bound to the provided app.
+ * An Auth instance can have multiple tenants.
  */
 export class Auth extends BaseAuth implements FirebaseServiceInterface {
   public INTERNAL: AuthInternals = new AuthInternals();
+  private tenantsMap: {[key: string]: TenantAwareAuth};
```
hiranya911: readonly

test/unit/auth/auth.spec.ts
```diff
@@ -200,2602 +219,2852 @@ function getSAMLConfigServerResponse(providerId: string): SAMLConfigServerRespon
...
 
-  it('verifyIdToken() should throw when project ID is not specified', () => {
-    const mockCredentialAuth = new Auth(mocks.mockCredentialApp());
-    const expected = 'Must initialize app with a cert credential or set your Firebase project ID ' +
-      'as the GOOGLE_CLOUD_PROJECT environment variable to call verifyIdToken().';
-    expect(() => {
-      mockCredentialAuth.verifyIdToken(mocks.generateIdToken());
-    }).to.throw(expected);
-  });
+        it('should not throw on valid tenant ID', () => {
```
hiranya911: This is pretty much the same as the following test case. Consider removing it.

test/unit/auth/auth.spec.ts
```diff
@@ -47,6 +49,15 @@ chai.use(chaiAsPromised);
 const expect = chai.expect;
 
 
+interface AuthTest {
```
hiranya911: This file is bit complicated to review. Can I assume it's just running the existing tests in 2 test configs, with the tests specific to an auth implementation enclosed in if-blocks?

src/auth/auth-api-request.ts
```diff
@@ -1470,4 +1474,34 @@ export class TenantAwareAuthRequestHandler extends AbstractAuthRequestHandler {
...
+   * @return {Promise<UserImportResult>} A promise that resolves when the operation completes
+   *     with the result of the import. This includes the number of successful imports, the number
+   *     of failed uploads and their corresponding errors.
+   */
+  public uploadAccount(
+      users: UserImportRecord[], options?: UserImportOptions): Promise<UserImportResult> {
+    // Add additional check to match tenant ID of imported user records.
+    users.forEach((user: UserImportRecord, index: number) => {
+      if (validator.isNonEmptyString(user.tenantId) &&
+          user.tenantId !== this.tenantId) {
```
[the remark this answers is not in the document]
bojeil-google: Actually it is needed. Basically, if they don't pass a `tenantId`, the backend can determine it from the URL. It is implicitly assumed to be the same tenant ID. If they pass an explicit `tenantId, we want to validate that it matches. If they pass an explicit mismatching `tenantId`, we want to fail as the developer's intention is to save to a different tenant.

src/auth/auth.ts
bojeil-google: One problem with this is that if they are not using typescript, they can overwrite this without any error thrown. I tried it and the property can be overwritten. Since this is publicly available, i think we should enforce the readonly.

src/auth/auth.ts
bojeil-google: Done

src/auth/auth.ts
bojeil-google: Done

test/unit/auth/auth.spec.ts
bojeil-google: done

test/unit/auth/auth.spec.ts
bojeil-google: Yeah, I completely understand. This is the same GitHub diff issue. This should no longer be an issue in the next PRs.
You are mostly right. Same tests are run for both. There are cases where I add specific tests for tenant specific behavior. For example:

- createCustomToken should throw unsupported tenant operation error
- verifyIdToken with token missing or mismatching tenant ID
- verifySessionCookie with missing or mismatching tenant ID
- importUsers with explicit mismatching tenant IDs
- confirm createSessionCookie calls verifyIdToken before making call to create a session cookie

I do this by checking `testConfig.Auth === TenantAwareAuth`

src/auth/auth.ts
hiranya911: I expected the TS compiler to generate a readonly getter for this, but looks like that only happens if we expose this as a getter in the code. So this is ok.

hiranya911: I expected the TS compiler to add a readonly getter, but looks like that only happens if we define a getter explicitly here. So this is ok.

bojeil-google:
Thanks for the review!

PR #556, opened 2019-06-06
Automatically retrying on 503 errors

hiranya911:
* Increased retry count to 4
* Backoff factor set to 0.5 (this results in the delay pattern of 0s, 1s, 2s and 4s)

### DIFF
--- a/test/unit/utils/api-request.spec.ts
+++ b/test/unit/utils/api-request.spec.ts
@@ -731,7 +731,7 @@ describe('HttpClient', () => {
         'content-type': 'application/json',
       });
     mockedRequests.push(scope);
-    const client = new HttpClient();
+    const client = new HttpClient(defaultRetryConfig());
     return client.send({
       method: 'GET',
       url: mockUrl,
@@ -741,7 +741,7 @@ describe('HttpClient', () => {
     });
   });
 
-  it('should not retry when RetryConfig is not set', () => {
+  it('should not retry when RetryConfig is explicitly null', () => {
     mockedRequests.push(mockRequestWithError({message: 'connection reset 1', code: 'ECONNRESET'}));
     const client = new HttpClient(null);
     const err = 'Error while making request: connection reset 1';
@@ -815,7 +815,7 @@ describe('HttpClient', () => {
         'content-type': 'application/json',
       });
     mockedRequests.push(scope2);
-    const client = new HttpClient();
+    const client = new HttpClient(testRetryConfig());
     return client.send({
       method: 'GET',
       url: mockUrl,
@@ -888,7 +888,7 @@ describe('HttpClient', () => {
         'content-type': 'application/json',
       });
     mockedRequests.push(scope);
-    const client = new HttpClient();
+    const client = new HttpClient(defaultRetryConfig());
     delayStub = sinon.stub(client as any, 'waitForRetry').resolves();
 
     return client.send({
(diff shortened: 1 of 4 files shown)


test/unit/utils/api-request.spec.ts
```diff
@@ -623,56 +623,91 @@ describe('HttpClient', () => {
     }).should.eventually.be.rejectedWith(err).and.have.property('code', 'app/network-error');
   });
 
-  it('should timeout when the response is delayed', () => {
+  it('should timeout when the response is repeatedly delayed', () => {
     const respData = {foo: 'bar'};
     const scope = nock('https://' + mockHost)
       .get(mockPath)
-      .twice()
+      .times(5)
```
bojeil-google: I think it is better to test passing explicit RetryConfig instead so you don't have to do things like:
`(client as any).retry.backOffFactor = 0;`

This also makes it easier to add changes to default config in the future.

And then add a test for default config.

hiranya911: Done

test/unit/utils/api-request.spec.ts
```diff
@@ -766,11 +815,7 @@ describe('HttpClient', () => {
         'content-type': 'application/json',
       });
     mockedRequests.push(scope2);
-    const client = new HttpClient({
-      maxRetries: 1,
-      maxDelayInMillis: 1000,
-      statusCodes: [503],
-    });
+    const client = new HttpClient();
```
bojeil-google: Shouldn't this use `testRetryConfig()` ?

test/unit/utils/api-request.spec.ts
```diff
@@ -847,13 +888,9 @@ describe('HttpClient', () => {
         'content-type': 'application/json',
       });
     mockedRequests.push(scope);
-    const client = new HttpClient({
-      maxRetries: 4,
-      backOffFactor: 0.5,
-      maxDelayInMillis: 60 * 1000,
-      statusCodes: [503],
-    });
+    const client = new HttpClient();
```
bojeil-google: Should you pass explicit test retry config with backoff factor on, here and elsewhere?

test/unit/utils/api-request.spec.ts
```diff
@@ -92,6 +92,18 @@ function mockRequestWithError(err: any) {
     .replyWithError(err);
 }
 
+/**
+ * Returns a new RetryConfig instance for testing. This is same as the default
+ * RetryConfig, with the backOffFactor set to 0 to avoid delays.
+ *
+ * @return {RetryConfig} A new RetryConfig instance.
+ */
+function testRetryConfig(): RetryConfig {
```
bojeil-google: should this take a parameter whether to reset backOffFactor and then you can use it in all tests not testing default behavior but require backOffFactor?

test/unit/utils/api-request.spec.ts
hiranya911: Done

test/unit/utils/api-request.spec.ts
hiranya911: That's same as default config. So using `defaultRetryConfig()` directly.

test/unit/utils/api-request.spec.ts
hiranya911: Used `defaultRetryConfig()` directly.

hiranya911:
@bojeil-google I used `defaultRetryConfig()` explicitly in the tests you pointed out to make the intention clear. Please take a look.

PR #570, opened 2019-06-17
Adds integration tests for tenant management APIs.

bojeil-google:
Adds integration tests for tenant management APIs.
These tests are skipped by default as multi-tenancy is a paid feature on Google Cloud Identity Platform.
To run these tests, --testMultiTenancy flag has to be added.
Adds d.ts references for multi-tenancy APIs.
Adds default email provider config when backend Auth server returns undefined for emailSignInConfig.


bojeil-google:
These tests are calling backend production endpoints. They are currently only run manually before a new binary release of the library is pushed. It is basically a sanity check for our new releases (they act more like e2e tests). They are not specific to one project.

test/integration/auth.spec.ts
```diff
@@ -428,6 +434,148 @@ describe('admin.auth', () => {
...
+    it('listTenants() should resolve with expected number of tenants', () => {
+      const tenantOptions2 = deepCopy(tenantOptions);
+      tenantOptions2.displayName = 'testTenant2';
+      return admin.auth().createTenant(tenantOptions2)
+        .then((actualTenant) => {
+          createdTenants.push(actualTenant.tenantId);
+          // Test listTenant pagination.
+          return admin.auth().listTenants(1);
+        })
+        .then((result) => {
```
hiranya911: Something we do in other SDKs to test list operations is to list everything, and see if the known results appear in the output. Something like:

```
const allTenantIds = listAndIterateThroughAllTenants();
// scan allTenantIds to see if createdTenants is a subset of it.
```

src/index.d.ts
```diff
@@ -602,6 +602,10 @@ declare namespace admin.auth {
      * resets, password or email updates, etc).
      */
     tokensValidAfterTime?: string;
+
```
hiranya911: These require a review from @egilmorez or somebody from his team. How about we make these changes in a separate PR, so the rest of the code changes don't get blocked on documentation updates?

test/integration/auth.spec.ts
bojeil-google: Done

src/index.d.ts
bojeil-google: I reverted the reference changes. Will have a follow up PR just for that.

src/index.d.ts
```diff
@@ -5191,4 +5190,4 @@ declare namespace admin.projectManagement {
 declare module 'firebase-admin' {
 }
 
-export = admin;
+export = admin;
```
hiranya911: Please fix.

bojeil-google: Done

PR #572, opened 2019-06-17
Defines Auth multi-tenancy references in index.d.ts.

bojeil-google:
Defines Auth multi-tenancy references in index.d.ts.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -604,7 +604,7 @@ declare namespace admin.auth {
     tokensValidAfterTime?: string;
   
     /**
-     * The user's tenant identifier if available.
+     * The ID of the tenant the user belongs to, if available.
      */
     tenantId?: string | null;
 
@@ -733,7 +733,7 @@ declare namespace admin.auth {
       sign_in_provider: string;
 
       /**
-       * The user's tenant ID if available.
+       * The ID of the tenant the user belongs to, if available.
        */
       tenant?: string;
       [key: string]: any;
@@ -1080,7 +1080,8 @@ declare namespace admin.auth {
    * A tenant configuration provides information such as the type of tenant (lightweight or
    * full service), display name, tenant identifier and email authentication configuration.
    * For OIDC/SAML provider configuration management, `TenantAwareAuth` instances should
-   * be used instead. When configuring these providers, note that tenants will inherit
+   * be used instead of a `Tenant` to retrieve the list of configured IdPs on a tenant.
+   * When configuring these providers, note that tenants will inherit
    * whitelisted domains and authenticated redirect URIs of their parent project.
    *
    * All other settings of a tenant will also be inherited. These will need to be managed
@@ -1089,13 +1090,13 @@ declare namespace admin.auth {
   interface Tenant {
 
     /**
-     * The current tenant identifier.
+     * The tenant identifier.
      */
     tenantId: string;
 
     /**
-     * The current tenant type: `lightweight` or `full_service`.
-     * Tenants that use separare billing and quota will require their own project and
+     * The tenant type: `lightweight` or `full_service`.
+     * Tenants that use separate billing and quota will require their own project and
      * must be defined as `full_service`.
      * `full_service` tenants may be subject to quota creation limits.
      * For additional project quota increases, refer to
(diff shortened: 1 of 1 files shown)


src/index.d.ts
```diff
@@ -952,6 +960,11 @@ declare namespace admin.auth {
      * The buffer of bytes representing the user’s password salt.
      */
     passwordSalt?: Buffer;
+  
+    /**
+     * The identifier of the tenant where user is to be imported to.
+     * When not provided, the user is uploaded to the default parent project.
```
wuyanna: If not provided, the user is uploaded to the tenant specified in the request parameter `tenantId`

src/index.d.ts
```diff
@@ -1051,31 +1064,128 @@ declare namespace admin.auth {
 
   type TenantType = 'lightweight' | 'full_service';
```
wuyanna: Do you need to add comment for this as well?

src/index.d.ts
```diff
@@ -1051,31 +1064,128 @@ declare namespace admin.auth {
...
+   * be used instead. When configuring these providers, note that tenants will inherit
+   * whitelisted domains and authenticated redirect URIs of their parent project.
+   *
+   * All other settings of a tenant will also be inherited. These will need to be managed
+   * from the Cloud Console UI.
+   */
   interface Tenant {
+
+    /**
+     * The current tenant identifier.
```
wuyanna: Do you think we should be more specific? like `The current tenant identifier to be passed in the requests.`

src/index.d.ts
wuyanna: nvm, saw you added below

src/index.d.ts
bojeil-google: For the `Tenant` interface, this is only returned in operations like `getTenant`, `createTenant`, `updateTenant` and `listTenants`. It doesn't seem to apply here. I think it applies more to `TenantAwareAuth#tenantId`. Check below to see the changes I added.

src/index.d.ts
bojeil-google: I expanded the description. Please take another look.

src/index.d.ts
```diff
@@ -1820,18 +1933,101 @@ declare namespace admin.auth {
...
+     *
+     * @param tenantId The `tenantId` corresponding to the tenant to delete.
+     *
+     * @return An empty promise fulfilled once the tenant has been deleted.
+     */
     deleteTenant(tenantId: string): Promise<void>;
+
+    /** 
+     * Creates a new tenant.
+     * When creating new tenants, tenants that use separare billing and quota will require their
```
kevinthecheung: separate

src/index.d.ts
```diff
@@ -602,6 +602,10 @@ declare namespace admin.auth {
      * resets, password or email updates, etc).
      */
     tokensValidAfterTime?: string;
+  
+    /**
+     * The user's tenant identifier if available.
```
kevinthecheung: The ID of the tenant the user belongs to

src/index.d.ts
```diff
@@ -727,6 +731,10 @@ declare namespace admin.auth {
        * `"google.com"`, `"twitter.com"`, or `"custom"`.
        */
       sign_in_provider: string;
+
+      /**
+       * The user's tenant ID if available.
```
kevinthecheung: same

src/index.d.ts
```diff
@@ -1051,31 +1067,128 @@ declare namespace admin.auth {
...
+   * be used instead. When configuring these providers, note that tenants will inherit
+   * whitelisted domains and authenticated redirect URIs of their parent project.
+   *
+   * All other settings of a tenant will also be inherited. These will need to be managed
+   * from the Cloud Console UI.
+   */
   interface Tenant {
+
+    /**
+     * The current tenant identifier.
```
kevinthecheung: Can you omit "current" here and throughout?

src/index.d.ts
```diff
@@ -1051,31 +1067,128 @@ declare namespace admin.auth {
...
+      /**
+       * Whether password is required for email sign-in. When not required,
+       * email sign-in can be performed with password or via email link sign-in.
+       */
       passwordRequired?: boolean;
     };
   }
 
+  /**
+   * Interface representing the properties to set on a new tenant to be created.
```
kevinthecheung: ... a new tenant.

src/index.d.ts
```diff
@@ -1820,18 +1933,101 @@ declare namespace admin.auth {
     ): Promise<admin.auth.AuthProviderConfig>;
   }
 
+  /**
+   * Tenant aware `Auth` interface used for managing user, configuring SAML/OIDC providers,
```
kevinthecheung: Tenant-aware

kevinthecheung: users

PR #573, opened 2019-06-18
Fix type defs for Credential interface

samtstern:
@hiranya911 before I go any farther:

- Was there a reason `getCertificate()` is not in the `.d.ts`?
- Are these typedef files normally hand-written?
- Many implementations return `null` from `getCertificate()` get the typedef says `Certificate` not `Certificate | null` or `Certificate | undefined` ... why does that even compile?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -285,7 +285,6 @@
           "version": "1.8.4",
           "resolved": "https://registry.npmjs.org/gaxios/-/gaxios-1.8.4.tgz",
           "integrity": "sha512-BoENMnu1Gav18HcpV9IleMPZ9exM+AvUjrAOV4Mzs/vfz2Lu/ABv451iEXByKiMPn2M140uul1txXCg83sAENw==",
-          "optional": true,
           "requires": {
             "abort-controller": "^3.0.0",
             "extend": "^3.0.2",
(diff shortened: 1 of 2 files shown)


hiranya911:
Hi @samtstern. 

1. `Certificate` is not part of our public API. So anything related to it doesn't appear in the typings.
2. Yes. Both Admin SDK's and JS SDK's typings are hand-written.
3. That's not something TypeScript enforces by default. You have to explicitly request it with `--strictNullChecks`.

samtstern:
So the reason I think we should expose this (I may misunderstand):
  * You can pass the `credential` field when you do `admin.initializeApp()` and we expect an `admin.credential.Credential`
  * If you do this and then use Firestore, it will explode because it calls `getCertificate()` on the credential object which is not guaranteed by the interface.  [See code for details](https://github.com/firebase/firebase-admin-node/blob/55dd22155d9ab465c103c41dbf3b709a11bad402/src/firestore/firestore.ts#L75)

So if `credential` is part of our public API then I think the API should express all of the properties and methods you need to make sure nothing crashes.

hiranya911:
Then we ought to fix that in the admin Firestore code. Let's not add "Certificate" to the public API. For one thing it's a terrible (and incorrect) name. I've let it remain in the code because it's considered an internal API. Secondly there's no equivalent concept in any of the other GCP libraries.

PR #576, opened 2019-06-19
Removing getCertificate() from the Credential interface

hiranya911:
`Credential` is a public interface and is declared as follows in the typings:

```
interface Credential {
    getAccessToken(): Promise<admin.GoogleOAuthAccessToken>;
}
```

But in the actual implementation (credential.ts), there's an additional method in the interface:

https://github.com/firebase/firebase-admin-node/blob/843df1229405d441fd5e8bd4170e38cd9fd4cbd4/src/auth/credential.ts#L291-L294

This can cause issues if the developer implements the public `Credential` interface, and passes it through `AppOptions`. Internal code that expects to find a `getCertificate()` method will fail.

This PR addresses this issue by separating the `getCertificate()` method from the `Credential` interface.

Resolves #573

### DIFF
--- a/src/auth/credential.ts
+++ b/src/auth/credential.ts
@@ -230,7 +230,7 @@ function getDetailFromResponse(response: HttpResponse): string {
 /**
  * Implementation of Credential that uses a service account certificate.
  */
-export class CertCredential implements Credential {
+export class CertCredential implements FirebaseCredential {
 
   private readonly certificate: Certificate;
   private readonly httpClient: HttpClient;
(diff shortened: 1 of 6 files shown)


samtstern:
Some other places:
https://github.com/firebase/firebase-admin-node/blob/2b4031334071476b5a6a481aa52578a713a5b719/src/storage/storage.ts#L73

https://github.com/firebase/firebase-admin-node/blob/9f770b4a96619e19297bb3e1d540fa710cd213af/src/utils/index.ts#L72

Issue #577, opened 2019-06-21
Enforce UserImportOptions when importing users with passwords


hiranya911:
When importing users with passwords, `UserImportOptions` must be specified. But the SDK does not check for this condition correctly. Instead it generates the following error when the user records contain passwords and `UserImportOptions` is not specified:

```
.../node_modules/firebase-admin/lib/auth/user-import-builder.js:176 if (!validator.isNonNullObject(options.hash)) { TypeError: Cannot read property 'hash' of undefined....
```

See https://stackoverflow.com/questions/56680087/firebase-admin-import-users-doesnt-work

bojeil-google:
Just to confirm, is the issue that we are not validating `options` is `nonNullObject` first?

hiranya911:
@bojeil-google. Yes, I think we should check `options` is specified before checking `options.hash`.

PR #578, opened 2019-06-21
Adds validation for UserImportOptions.

bojeil-google:
Adds validation for UserImportOptions.

### DIFF
--- a/src/auth/user-import-builder.ts
+++ b/src/auth/user-import-builder.ts
@@ -301,7 +301,7 @@ export class UserImportBuilder {
     if (!validator.isNonNullObject(options)) {
       throw new FirebaseAuthError(
         AuthClientErrorCode.INVALID_ARGUMENT,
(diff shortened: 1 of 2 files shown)


hiranya911:
Resolves #577

src/auth/user-import-builder.ts
```diff
@@ -298,6 +298,12 @@ export class UserImportBuilder {
     if (!requiresHashOptions) {
       return {};
     }
+    if (!validator.isNonNullObject(options)) {
+      throw new FirebaseAuthError(
+        AuthClientErrorCode.INVALID_ARGUMENT,
+        'Invalid or no "UserImportOptions" argument provided.',
```
hiranya911: Nit: Perhaps also add "UserImportOptions are required when importing users with passwords."

bojeil-google: I replaced with: "UserImportOptions are required when importing users with passwords."

PR #584, opened 2019-06-25
Adds basic integration tests for the following in multi-tenancy context:

bojeil-google:
- User management
- Custom claims
- Import users
- List users
- Token revocation
- Email link generation (skipped due to backend bug)
- OIDC/SAML management APIs.

### DIFF
--- a/test/integration/auth.spec.ts
+++ b/test/integration/auth.spec.ts
@@ -50,7 +50,7 @@ const sessionCookieUids = [
 const testPhoneNumber = '+11234567890';
 const testPhoneNumber2 = '+16505550101';
 const nonexistentPhoneNumber = '+18888888888';
-const updatedEmail = generateRandomString(20) + '@example.com';
+const updatedEmail = generateRandomString(20).toLowerCase() + '@example.com';
 const updatedPhone = '+16505550102';
 const customClaims: {[key: string]: any} = {
   admin: true,
@@ -58,7 +58,7 @@ const customClaims: {[key: string]: any} = {
 };
 const uids = [newUserUid + '-1', newUserUid + '-2', newUserUid + '-3'];
 const mockUserData = {
-  email: newUserUid + '@example.com',
+  email: newUserUid.toLowerCase() + '@example.com',
   emailVerified: false,
   phoneNumber: testPhoneNumber,
   password: 'password',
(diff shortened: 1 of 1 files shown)


bojeil-google:
I want to add testing for `verifyIdToken()` but i want a token scoped for a specific tenant. We pull in the [client SDK Node.js](https://github.com/firebase/firebase-admin-node/blob/auth-multi-tenancy/test/integration/auth.spec.ts#L24) library which can help us sign in to a specific tenant but the latest version does not support multi-tenancy (not publicly available). So for now, I have to wait for client SDK to be updated before we can add that test.

test/integration/auth.spec.ts
```diff
@@ -506,6 +505,268 @@ describe('admin.auth', () => {
...
+        hash: {
+          algorithm: 'HMAC_SHA256',
+          key: Buffer.from('secret'),
+        },
+      };
+      const rawPassword = 'password';
+      const rawSalt = 'NaCl';
+
+      before(() => {
+        tenantAwareAuth = admin.auth().forTenant(createdTenantId);
```
hiranya911: Can we do this outside `before`? Something like:

```
const tenantAwareAuth = admin.auth().forTenant(createdTenantId);
```

Here and in other similar describe blocks.

test/integration/auth.spec.ts
```diff
@@ -506,6 +505,268 @@ describe('admin.auth', () => {
...
+            });
+        }
+      });
+
+      it('createUser() should create a user in the expected tenant', () => {
+        return tenantAwareAuth.createUser(newUserData)
+          .then((userRecord) => {
+            createdUserUid = userRecord.uid;
+            expect(userRecord.tenantId).to.equal(createdTenantId);
+            expect(userRecord.email.toLowerCase()).to.equal(newUserData.email.toLowerCase());
```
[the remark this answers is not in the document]
bojeil-google: Done

test/integration/auth.spec.ts
bojeil-google: I can't because `initializeApp` has to be called before I do this. We call it in `before` at setup: https://github.com/firebase/firebase-admin-node/blob/auth-multi-tenancy/test/integration/setup.ts#L40

It is not worth a refactor.

PR #589, opened 2019-07-03
add environment variable to point admin sdk at rtdb emulator

IanWyszynski:

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/src/auth/credential.ts
+++ b/src/auth/credential.ts
@@ -419,6 +419,23 @@ export class ApplicationDefaultCredential implements FirebaseCredential {
   }
 }
 
+export class DatabaseEmulatorCredential implements FirebaseCredential {
+  public getAccessToken(): Promise<GoogleOAuthAccessToken> {
+    return Promise.resolve({
+      expires_in: 1000000,
+      access_token: 'owner',
+    });
+  }
+
+  public getCertificate(): Certificate {
+    return {
+      clientEmail: '',
+      projectId: '',
+      privateKey: '',
+    };
+  }
+}
+
 function credentialFromFile(filePath: string, httpAgent?: Agent): Credential {
   const credentialsFile = readCredentialFile(filePath);
   if (typeof credentialsFile !== 'object') {
--- a/src/auth/credential.ts
+++ b/src/auth/credential.ts
@@ -428,11 +428,7 @@ export class DatabaseEmulatorCredential implements FirebaseCredential {
   }
 
   public getCertificate(): Certificate {
-    return {
-      clientEmail: '',
-      projectId: '',
-      privateKey: '',
-    };
+    return null;
   }
 }
 
--- a/src/auth/credential.ts
+++ b/src/auth/credential.ts
@@ -422,7 +422,7 @@ export class ApplicationDefaultCredential implements FirebaseCredential {
 export class DatabaseEmulatorCredential implements Credential {
   public getAccessToken(): Promise<GoogleOAuthAccessToken> {
     return Promise.resolve({
-      expires_in: 1000000,
+      expires_in: 1000,
       access_token: 'owner',
     });
   }
--- a/src/auth/credential.ts
+++ b/src/auth/credential.ts
@@ -419,7 +419,7 @@ export class ApplicationDefaultCredential implements FirebaseCredential {
   }
 }
 
-export class DatabaseEmulatorCredential implements FirebaseCredential {
+export class DatabaseEmulatorCredential implements Credential {
   public getAccessToken(): Promise<GoogleOAuthAccessToken> {
     return Promise.resolve({
       expires_in: 1000000,
--- a/src/auth/credential.ts
+++ b/src/auth/credential.ts
@@ -426,10 +426,6 @@ export class DatabaseEmulatorCredential implements FirebaseCredential {
       access_token: 'owner',
     });
   }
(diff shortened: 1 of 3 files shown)


src/auth/credential.ts
```diff
@@ -419,6 +419,23 @@ export class ApplicationDefaultCredential implements FirebaseCredential {
...
 
+export class DatabaseEmulatorCredential implements FirebaseCredential {
+  public getAccessToken(): Promise<GoogleOAuthAccessToken> {
+    return Promise.resolve({
+      expires_in: 1000000,
+      access_token: 'owner',
+    });
+  }
+
+  public getCertificate(): Certificate {
```
samtstern: You should return `{}` (which is what the emulator does) or drop this entirely.  After https://github.com/firebase/firebase-admin-node/pull/576 I don't think it's needed.

Having stubbed values (`''`) will definitely cause more problems than it solves.

src/firebase-namespace.ts
```diff
@@ -46,6 +47,13 @@ const DEFAULT_APP_NAME = '[DEFAULT]';
  */
 export const FIREBASE_CONFIG_VAR: string = 'FIREBASE_CONFIG';
 
+/**
+ * Constant holding the fully-qualified domain URI for a database emulator
+ * instance. If specified, the contents of this variable will be used to
+ * set `databaseURL` in FirebaseAppOptions.
+ */
+const FIREBASE_DATABASE_EMULATOR_HOST_VAR: string = 'FIREBASE_DATABASE_EMULATOR_HOST';
```
samtstern: Can you add an example to the comment?  Do you include the `?ns=` param?

src/auth/credential.ts
IanWyszynski: The `FirebaseCredential` interface still requires this method, however 'null' is a valid return type according to its interface.

src/firebase-namespace.ts
```diff
@@ -46,6 +47,24 @@ const DEFAULT_APP_NAME = '[DEFAULT]';
  */
 export const FIREBASE_CONFIG_VAR: string = 'FIREBASE_CONFIG';
 
+/**
```
hiranya911: I'm a bit uneasy about this logic residing in the global firebase-namespace file. Can we somehow move this to the database module?

src/auth/credential.ts
hiranya911: #576 will remove this method from the interface.

IanWyszynski: I think that PR removes it from the `Credential` interface, not the `FirebaseCredentialInterface`. Is there a reason to use the latter as opposed to the former? I just used it because it looks like that's the interface that the ApplicationDefaultCredential implements.

hiranya911: Credential is the public interface. FirebaseCredential is for our internal use only. You should definitely be using the Credential interface.

IanWyszynski:
Abandoned and replaced by https://github.com/firebase/firebase-admin-node/pull/596

ryanpbrewster:
I think we need to re-open this PR and discuss using fake credentials for the emulators. Right now we're requiring developers to either do this themselves, or use a real production service account credential.

src/database/database.ts
```diff
@@ -1,10 +1,32 @@
...
+const APP_OPTIONS_PROJECT_ID = 'projectId';
+const APP_OPTIONS_CREDENTIAL = 'credential';
+
+/**
+ * Constant holding the fully-qualified domain URI for a database emulator
+ * instance. If specified, the contents of this variable will be used to
+ * set `databaseURL` in FirebaseAppOptions. The varaible should be a complete
+ * URI specifying a transfer protocol, hostname, and port number:
+ *
+ * FIREBASE_DATABASE_EMULATOR_HOST=http://localhost:9000
```
schmidt-sebastian: If I am not mistaken, the protocol is omitted in FIRESTORE_EMULATOR_HOST. I see that it might make sense to include it here (since we need to include the namespace). Should we come up with a consistent pattern that apploes to both RTDB and Firestore? @ryanpbrewster

ryanpbrewster: the protocol should not be included in the `FIRESOTRE_EMULATOR_HOST` variable.

schmidt-sebastian: Do we care about the consistency between FIRESTORE_EMULATOR_HOST and FIREBASE_DATABASE_EMULATOR_HOST?

ryanpbrewster: Yes, ideally neither should include the protocol.

IanWyszynski: This comment is dated, neither emulator host variable has a protocol.

dungahk:
I totally agree with reopening this, or at least reverting #602 

I am wondering how a new feature in the Firebase JS SDK is sufficient for the firebase-admin Node SDK?

For example, I am using firebase-admin with firebase-functions and I do not touch the firebase JS client SDK and having this in the admin SDK would be really appreciated.

The only weird thing I found is that if I run the emulators using emulators:start and manually invoke the functions it seems to work, but if I try to automate the steps using emulators:exec it doesn't work.

samtstern:
@dungahk can you open a new issue on the `firebase-tools` repo describing the issue you're having with `emulators:exec` versus `emulators:start`?

PR #595, opened 2019-07-18
API for managing RTDB Security Rules

hiranya911:
New APIs for getting and setting RTDB rules:

```
getRules(): Promise<string>
getRulesJSON(): Promise<object>
setRules(source: string|Buffer|object): Promise<void>
```

New methods are dynamically injected to the `Database` object returned by the `@firebase/database` package. Type-safety is maintained by defining a new interface with the same name, letting TS declaration merging to take effect.

go/firebase-rules-admin

Related to #504

### DIFF
--- a/src/database/database.ts
+++ b/src/database/database.ts
@@ -46,7 +46,6 @@ export class DatabaseService implements FirebaseServiceInterface {
   public readonly INTERNAL: DatabaseInternals = new DatabaseInternals();
 
   private readonly appInternal: FirebaseApp;
-  private readonly httpClient: AuthorizedHttpClient;
 
   constructor(app: FirebaseApp) {
     if (!validator.isNonNullObject(app) || !('options' in app)) {
@@ -56,7 +55,6 @@ export class DatabaseService implements FirebaseServiceInterface {
       });
     }
     this.appInternal = app;
-    this.httpClient = new AuthorizedHttpClient(app);
   }
 
   /**
--- a/src/database/database.ts
+++ b/src/database/database.ts
@@ -113,6 +113,9 @@ export class DatabaseService implements FirebaseServiceInterface {
 
 const RULES_URL_PATH = '.settings/rules.json';
 
+/**
+ * A helper client for managing RTDB security rules.
+ */
 class DatabaseRulesClient {
 
   private readonly httpClient: AuthorizedHttpClient;
@@ -123,6 +126,12 @@ class DatabaseRulesClient {
     this.dbUrl = dbUrl;
   }
 
+  /**
+   * Gets the currently applied security rules as a string. The return value consists of
+   * the rules source including comments.
+   *
+   * @return {Promise<string>} A promise fulfilled with the rules as a raw string.
+   */
   public getRules(): Promise<string> {
     const req: HttpRequestConfig = {
       method: 'GET',
@@ -137,6 +146,12 @@ class DatabaseRulesClient {
       });
   }
 
+  /**
+   * Gets the currently applied security rules as a parsed JSON object. Any comments in
+   * the original source are stripped away.
+   *
+   * @return {Promise<object>} A promise fulfilled with the parsed rules source.
+   */
   public getRulesJSON(): Promise<object> {
     const req: HttpRequestConfig = {
       method: 'GET',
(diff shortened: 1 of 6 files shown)


src/database/database.ts
```diff
@@ -85,3 +110,122 @@ export class DatabaseService implements FirebaseServiceInterface {
...
+  /**
+   * Gets the currently applied security rules as a parsed JSON object. Any comments in
+   * the original source are stripped away.
+   *
+   * @return {Promise<object>} A promise fulfilled with the parsed rules source.
+   */
+  public getRulesJSON(): Promise<object> {
+    const req: HttpRequestConfig = {
+      method: 'GET',
+      url: `${this.getRulesUrl()}?format=strict`,
```
ryanpbrewster: should we be doing some kind of `resolve(...)` thing here? What happens if the dbUrl looks like `s-usc1c-nss-200.firebaseio.com?ns=foo`

src/database/database.ts
```diff
@@ -85,3 +110,122 @@ export class DatabaseService implements FirebaseServiceInterface {
...
+   *
+   * @param {string|Buffer|object} source Source of the rules to apply. Must not be `null`
+   *  or empty.
+   * @return {Promise<void>} Resolves when the rules are set on the Database.
+   */
+  public setRules(source: string | Buffer | object): Promise<void> {
+    if (!validator.isNonEmptyString(source) &&
+      !validator.isBuffer(source)  &&
+      !validator.isNonNullObject(source)) {
+        throw new FirebaseDatabaseError({
```
ryanpbrewster: I am not sure about this pattern. It's tricky because although this looks like an async method, doing something like
```
db.setRules("foo").catch(err => {
```
will not actually catch this error (because it's thrown, not returned). Thoughts on this? I think it might be better to make this an `async` method or explicit return a `Promise.rejected` here

src/database/database.ts
hiranya911: The URL returned by `getRulesUrl()` is guaranteed to end with `/.settings/rules.json`. If the original dbUrl had any query params, they do get stripped away by `getRulesUrl()`. Is that something we need to be concerned of? Does the user-specified DB URL ever have any query params?

ryanpbrewster: It can (notably true for the emulator, where it looks like `http://localhost:9000?ns=dbname`).

src/database/database.ts
```diff
@@ -85,3 +110,122 @@ export class DatabaseService implements FirebaseServiceInterface {
...
+
+  private getErrorMessage(err: HttpError): string {
+    const intro = 'Error while accessing security rules';
+    try {
+      const body: {error?: string} = err.response.data;
+      if (body && body.error) {
+        return `${intro}: ${body.error.trim()}`;
+      }
+    } catch {
+      // Ignore parsing errors
```
Feiyang1: Where does the parsing happen?

src/database/database.ts
hiranya911: We don't use async in this codebase yet.  Returning with a rejection instead.

src/database/database.ts
hiranya911: Fixed, and added test cases.

src/database/database.ts
hiranya911: It happens in `response.data`. It will throw if the response payload cannot be parsed as json.

src/database/database.ts
```diff
@@ -43,6 +55,7 @@ export class DatabaseService implements FirebaseServiceInterface {
       });
     }
     this.appInternal = app;
+    this.httpClient = new AuthorizedHttpClient(app);
```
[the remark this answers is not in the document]
hiranya911: Removed

src/index.d.ts
```diff
@@ -2064,6 +2064,31 @@ declare namespace admin.database {
...
+    /**
+     * Gets the currently applied security rules as a parsed JSON object. Any comments in
+     * the original source are stripped away.
+     *
+     * @return A promise fulfilled with the parsed rules object.
+     */
+    getRulesJSON(): Promise<object>;
+
+    /**
+     * Sets the specified rules on the Firebase Database instance. If the rules source is
```
egilmorez: Suggest either "Firebase Realtime Database" or just "Realtime Database."

src/index.d.ts
```diff
@@ -2064,6 +2064,31 @@ declare namespace admin.database {
...
+     * specified as a string or a Buffer, it may include comments.
```
egilmorez: Should Buffer be capitalized?

src/index.d.ts
```diff
@@ -2064,6 +2064,31 @@ declare namespace admin.database {
...
+     *
+     * @param source Source of the rules to apply. Must not be `null` or empty.
+     * @return Resolves when the rules are set on the Database.
```
egilmorez: Suggest "Realtime Database."

I know it's a bit much, but Cloud Firestore and even Storage could also be considered databases.

src/index.d.ts

src/index.d.ts
hiranya911: Yes, in this instance it is referring to the built-in Buffer type of Node.js

src/index.d.ts

hiranya911:
Thanks @egilmorez. I've made the suggested changes.

PR #596, opened 2019-07-18
add FIREBASE_DATABASE_EMULATOR_HOST_VAR

IanWyszynski:
Support specifying database endpoint with FIREBASE_DATABASE_EMULATOR_HOST.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
   ],
   "types": "./lib/index.d.ts",
   "dependencies": {
-    "@firebase/database": "^0.4.7",
+    "@firebase/database": "^0.4.8",
     "@types/node": "^8.0.53",
     "dicer": "^0.3.0",
     "jsonwebtoken": "8.1.0",
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
   ],
   "types": "./lib/index.d.ts",
   "dependencies": {
-    "@firebase/database": "^0.4.8",
+    "@firebase/database": "^0.4.7",
     "@types/node": "^8.0.53",
     "dicer": "^0.3.0",
     "jsonwebtoken": "8.1.0",
--- a/src/database/database.ts
+++ b/src/database/database.ts
@@ -1,11 +1,9 @@
 import {FirebaseApp} from '../firebase-app';
 import {FirebaseDatabaseError} from '../utils/error';
 import {FirebaseServiceInterface, FirebaseServiceInternalsInterface} from '../firebase-service';
-import {Database} from '@firebase/database';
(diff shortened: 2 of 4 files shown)


test/unit/firebase-app.spec.ts
```diff
@@ -433,7 +433,8 @@ describe('FirebaseApp', () => {
       return app2.delete();
     });
 
-    it('should throw when databaseURL is not set', () => {
+    it('should throw when databaseURL is not set, will fail if ' +
+      'FIREBASE_DATABASE_EMULATOR_HOST is set', () => {
```
hiranya911: Let's unset the env variable in the test.

```
delete process.env['FIREBASE_DATABASE_EMULATOR_HOST'];
```

test/unit/firebase.spec.ts
```diff
@@ -167,7 +167,8 @@ describe('Firebase', () => {
       }).to.throw('The default Firebase app does not exist.');
     });
 
-    it('should throw given no databaseURL key when initializing the app', () => {
+    it('should throw given no databaseURL key when initializing the app,' +
+      ' will fail if FIREBASE_DATABASE_EMULATOR_HOST is set', () => {
```
hiranya911: Here too.

test/unit/firebase-app.spec.ts
```diff
@@ -433,7 +433,8 @@ describe('FirebaseApp', () => {
       return app2.delete();
     });
 
-    it('should throw when databaseURL is not set', () => {
+    it('should throw when databaseURL is not set, will fail if ' +
```
hiranya911: Can we add a test case for the new functionality? Set the env variable and see app.database() works as intended.

IanWyszynski:
This is currently blocked on the minor version bump in https://github.com/firebase/firebase-js-sdk/pull/2005, which must get merged first.

package.json
```diff
@@ -54,7 +54,7 @@
   ],
   "types": "./lib/index.d.ts",
   "dependencies": {
-    "@firebase/database": "^0.4.7",
+    "@firebase/database": "^0.4.8",
```
hiranya911: You will have to run `npm install` on your working copy and push the updated package lock file for the tests to pass.

IanWyszynski: Actually, the firebase-js-sdk won't be exposing the variable so we won't need to change this at all.

src/database/database.ts
```diff
@@ -5,6 +5,14 @@ import {Database} from '@firebase/database';
 
 import * as validator from '../utils/validator';
 
+/**
+ * This variable is re-defined in the firebase-js-sdk. Before modifying this
```
egilmorez: "redefined" doesn't really need a hyphen.

PR #604, opened 2019-07-24
Implementing the Firebase Security Rules API

hiranya911:
Added the base framework for the new security rules API (nothing new exposed in the public API --i.e. `admin` namespace -- yet). This PR specifically adds the `SecurityRules` class and the following 2 methods:

* `getRuleset()`
* `getFirestoreRuleset()`

Also defined the types `Ruleset`, `RulesetMetadata` and `RulesFile`  which will be added to the public API in the future.

go/firebase-rules-admin

### DIFF
--- a/src/security-rules/security-rules-api-client.ts
+++ b/src/security-rules/security-rules-api-client.ts
@@ -17,8 +17,9 @@
 import { HttpRequestConfig, HttpClient, HttpError } from '../utils/api-request';
 import { PrefixedFirebaseError } from '../utils/error';
 import { FirebaseSecurityRulesError, SecurityRulesErrorCode } from './security-rules-utils';
+import * as validator from '../utils/validator';
 
-const RULES_API_URL = 'https://firebaserules.googleapis.com/v1/';
+const RULES_API_URL = 'https://firebaserules.googleapis.com/v1';
 
 /**
  * Class that facilitates sending requests to the Firebase security rules backend API.
@@ -27,25 +28,31 @@ const RULES_API_URL = 'https://firebaserules.googleapis.com/v1/';
  */
 export class SecurityRulesApiClient {
 
-  constructor(private readonly httpClient: HttpClient) { }
+  private readonly url: string;
+
+  constructor(private readonly httpClient: HttpClient, projectId: string) {
+    if (!validator.isNonEmptyString(projectId)) {
+      throw new FirebaseSecurityRulesError(
+        'invalid-argument',
+        'Failed to determine project ID. Initialize the SDK with service account credentials, or '
+          + 'set project ID as an app option. Alternatively, set the GOOGLE_CLOUD_PROJECT '
+          + 'environment variable.');
+    }
+
+    this.url = `${RULES_API_URL}/projects/${projectId}`;
+  }
 
   /**
-   * Gets the specified resource from the rules API. Resource name must be full qualified names with project
-   * ID prefix (e.g. `projects/project-id/rulesets/ruleset-name`).
+   * Gets the specified resource from the rules API. Resource names must be the short names without project
+   * ID prefix (e.g. `rulesets/ruleset-name`).
    *
    * @param {string} name Full qualified name of the resource to get.
    * @returns {Promise<T>} A promise that fulfills with the resource.
    */
   public getResource<T>(name: string): Promise<T> {
-    if (!name.startsWith('projects/')) {
-      const err = new FirebaseSecurityRulesError(
-        'invalid-argument', 'Resource name must have a project ID prefix.');
-      return Promise.reject(err);
-    }
-
     const request: HttpRequestConfig = {
       method: 'GET',
-      url: `${RULES_API_URL}${name}`,
+      url: `${this.url}/${name}`,
     };
     return this.httpClient.send(request)
       .then((resp) => {
--- a/src/security-rules/security-rules-api-client.ts
+++ b/src/security-rules/security-rules-api-client.ts
@@ -1,0 +1,86 @@
+/*!
+ * Copyright 2019 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { HttpRequestConfig, HttpClient, HttpError } from '../utils/api-request';
+import { PrefixedFirebaseError } from '../utils/error';
+import { FirebaseSecurityRulesError, SecurityRulesErrorCode } from './security-rules-utils';
+
+const RULES_API_URL = 'https://firebaserules.googleapis.com/v1/';
+
+/**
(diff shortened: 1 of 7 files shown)


src/security-rules/security-rules-api-client.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * Gets the specified resource from the rules API. Resource name must be full qualified names with project
+   * ID prefix (e.g. `projects/project-id/rulesets/ruleset-name`).
+   *
+   * @param {string} name Full qualified name of the resource to get.
+   * @returns {Promise<T>} A promise that fulfills with the resource.
+   */
+  public getResource<T>(name: string): Promise<T> {
+    if (!name.startsWith('projects/')) {
```
ryanpbrewster: may be worth regexing this to catch malformed project ids (which, for instance, I think need to be lowercase).

src/security-rules/security-rules-api-client.ts
```diff
@@ -0,0 +1,92 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { HttpRequestConfig, HttpClient, HttpError } from '../utils/api-request';
+import { PrefixedFirebaseError } from '../utils/error';
+import { FirebaseSecurityRulesError, SecurityRulesErrorCode } from './security-rules-utils';
+
+const RULES_API_URL = 'https://firebaserules.googleapis.com/v1/';
```
ryanpbrewster: can we drop the slash here and explicitly put it in down below: `${RULES_API_URL}/${name}`?

src/security-rules/security-rules-api-client.ts
```diff
@@ -0,0 +1,92 @@
...
+      return Promise.reject(err);
+    }
+
+    const request: HttpRequestConfig = {
+      method: 'GET',
+      url: `${RULES_API_URL}${name}`,
+    };
+    return this.httpClient.send(request)
+      .then((resp) => {
+        return resp.data as T;
```
ryanpbrewster: is this going to work for things like `ListRulesets`, which returns a JSON array?

src/security-rules/security-rules.ts
```diff
@@ -0,0 +1,178 @@
...
+          + 'environment variable.');
+    }
+
+    this.projectId = projectId;
+    this.client = new SecurityRulesApiClient(new AuthorizedHttpClient(app));
+  }
+
+  /**
+   * Gets the Ruleset identified by the given name. The input name should be the short name string without
+   * the project ID prefix. Rejects with a `not-found` error if the specified Ruleset cannot be found.
```
ryanpbrewster: can we provide an example here? it seems like we're expecting `"foo"` for `projects/rpb/rulesets/foo`, yeah?

src/security-rules/security-rules.ts
```diff
@@ -0,0 +1,178 @@
...
+   * Gets the Ruleset currently applied to Cloud Firestore. Rejects with a `not-found` error if no Ruleset is
+   * applied on Firestore.
+   *
+   * @returns {Promise<Ruleset>} A promise that fulfills with the Firestore Ruleset.
+   */
+  public getFirestoreRuleset(): Promise<Ruleset> {
+    return this.getRulesetForService(SecurityRules.CLOUD_FIRESTORE);
+  }
+
+  private getRulesetForService(name: string): Promise<Ruleset> {
```
ryanpbrewster: Can we change `name` to `serviceName`?

src/security-rules/security-rules.ts
```diff
@@ -0,0 +1,178 @@
...
+    const resource = `projects/${this.projectId}/releases/${name}`;
```
ryanpbrewster: this is pretty tricky, storage has multiple releases (one per bucket), so the non-default buckets look like
```
projects/rpb/releases/firebase.storage/bucket1
```

how do you want to handle that here?

src/security-rules/security-rules-api-client.ts
hiranya911: Yes, this works for any type. It's really just a hint to the compiler, and simplifies the calling code.

src/security-rules/security-rules.ts
hiranya911: The caller will have to append the bucket name and pass it in here. Something like:

```
getStorageRuleset(bucket?: string): Promise<Ruleset> {
  let serviceName = 'firebase.storage';
  if (bucket) {
    serviceName += `/${bucket}`;
  }
  return this.getRulesetForService(serviceName);
}
```

src/security-rules/security-rules.ts
hiranya911: Done

src/security-rules/security-rules.ts
hiranya911: Done

src/security-rules/security-rules-api-client.ts
hiranya911: Done

src/security-rules/security-rules-api-client.ts
hiranya911: I simplified things a bit here by pushing the projectId to the constructor of the `RulesApiClient`. Now it stores a partial URL as part of its state:

```
this.url = `${RULES_API_URL}/projects/${projectId}`;
```

Callers should pass in the resource names without project ID:

```
client.getResource('rulesets/my-ruleset');
```

As for project ID validation, we are doing the same check here as our other modules (Auth, FCM), and so it should be sufficient.

src/security-rules/security-rules.ts
ryanpbrewster: I would find that confusing (as an internal developer, not as a consumer of the API).

Can we call this method `getRulesetForRelease`?

hiranya911:
Thanks @ryanpbrewster. I've made a pass and addressed your comments.

src/security-rules/security-rules-api-client.ts
ryanpbrewster: I'll defer to you here, but:

- there aren't that many kinds of resources
- it may be easier to explicitly spell them all out as unique methods:
```
client.getRuleset(<ruleset_name>)
client.listRulesets()
client.getRelease()
client.listReleases()
```
which directly encodes the exposed backend API. Then add another layer on top of that.

hiranya911: This is a good idea. Will implement in the next PRs.

src/security-rules/security-rules.ts
hiranya911: Done

src/security-rules/security-rules.ts
```diff
@@ -0,0 +1,169 @@
...
+  readonly updateTime: string;
+}
+
+interface RulesetResponse {
+  readonly name: string;
+  readonly createTime: string;
+  readonly source: {
+    readonly files: RulesFile[];
+  };
+}
```
rachelmyers: Are `Release` and `RulesetResponse` additions to what we proposed?

src/security-rules/security-rules.ts
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * Representa a set of Firebase security rules.
```
rachelmyers: s/Representa/Represents

src/security-rules/security-rules.ts
hiranya911: These are part of the internal implementation. They are not exported and also won't be added to the typings.

src/security-rules/security-rules.ts
hiranya911: Fixed.

PR #607, opened 2019-07-29
Implemented createRulesFileFromSource() and createRuleset() APIs

hiranya911:
In addition to the above 2 new methods, I'm also refactoring the implementation a bit based on the feedback received in #604 

1. `SecurityRulesApiClient` now closely resembles the REST API.
  - `getRuleset()`
  - `createRuleset()`
  - `getRelease()`
2.  `SecurityRules` class is now mostly a thin wrapper around `SecurityRulesApiClient`.

### DIFF
--- a/src/security-rules/security-rules-api-client.ts
+++ b/src/security-rules/security-rules-api-client.ts
@@ -21,6 +21,24 @@ import * as validator from '../utils/validator';
 
 const RULES_V1_API = 'https://firebaserules.googleapis.com/v1';
 
+export interface Release {
+  readonly name: string;
+  readonly rulesetName: string;
+  readonly createTime: string;
+  readonly updateTime: string;
+}
+
+export interface RulesetContent {
+  readonly source: {
+    readonly files: Array<{name: string, content: string}>;
+  };
+}
+
+export interface RulesetResponse extends RulesetContent {
+  readonly name: string;
+  readonly createTime: string;
+}
+
 /**
  * Class that facilitates sending requests to the Firebase security rules backend API.
  *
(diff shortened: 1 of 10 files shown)


src/security-rules/security-rules-api-client.ts
```diff
@@ -49,11 +111,31 @@ export class SecurityRulesApiClient {
...
-  public getResource<T>(name: string): Promise<T> {
+  private getResource<T>(name: string): Promise<T> {
     const request: HttpRequestConfig = {
       method: 'GET',
       url: `${this.url}/${name}`,
     };
+    return this.sendRequest<T>(request);
+  }
+
+  private getRulesetName(name: string): Promise<string> {
```
ryanpbrewster: does this need to return a `Promise`? it looks purely synchronous

src/security-rules/security-rules-api-client.ts
```diff
@@ -19,7 +19,25 @@ import { PrefixedFirebaseError } from '../utils/error';
...
+export interface Release {
+  readonly name: string;
+  readonly rulesetName: string;
+  readonly createTime: string;
+  readonly updateTime: string;
+}
+
+export interface RulesetContent {
+  readonly source: {
+    readonly files: Array<{name: string, content: string}>;
```
ryanpbrewster: it looks like we have a `RulesFile` type, can we make this a `RulesFile[]`?

src/security-rules/security-rules.ts
```diff
@@ -143,9 +115,59 @@ export class SecurityRules implements FirebaseServiceInterface {
...
+  }
+
+  /**
+   * Creates a new `Ruleset` with one or more rules files.
+   *
+   * @param {RulesFile} file Rules file to include in the new Ruleset.
+   * @param {RulesFile[]} additionalFiles Additional rules files to be included in the new Ruleset.
+   * @returns {Promise<Ruleset>} A promise that fulfills with the newly created Ruleset.
+   */
+  public createRuleset(file: RulesFile, ...additionalFiles: RulesFile[]): Promise<Ruleset> {
```
ryanpbrewster: I'm not sure we should allow developers to upload multiple rules files. It is not supported at the moment, and we'll probably want to re-think this API when it comes to adding support for multiple files.

ryanpbrewster:
I do not think the link in your PR description is correct, that issue does not seem relevant to this PR.

src/security-rules/security-rules-api-client.ts
hiranya911: This was done so that errors result in rejections as expected. But we could achieve the same by starting the promise chain before calling `getRulesetName`. 

Changed into a sync method.

src/security-rules/security-rules-api-client.ts
hiranya911: I'd prefer not to import things in `security-rules.ts` into this file. Right now the import chain flows in the other direction. We can probably address this by moving all the shared types into a new file named `security-rules-api.ts`, but that feels like overkill.

src/security-rules/security-rules.ts
hiranya911: That's a good point. Removed `additionalFiles` from the signature and changed the implementation as needed. We can add it back in the future as a non-breaking change if necessary.

hiranya911:
Thanks @ryanpbrewster. Made the suggested changes and updated the issue description.

src/security-rules/security-rules-api-client.ts
```diff
@@ -49,11 +114,29 @@ export class SecurityRulesApiClient {
...
-  public getResource<T>(name: string): Promise<T> {
+  private getResource<T>(name: string): Promise<T> {
     const request: HttpRequestConfig = {
       method: 'GET',
       url: `${this.url}/${name}`,
     };
+    return this.sendRequest<T>(request);
+  }
+
+  private getRulesetName(name: string): string {
```
ryanpbrewster: any reason we do this for rulesets but not for releases?

hiranya911: With Rulesets we will need this name processing functionality for deleteRuleset (implemented in a future PR) as well. I don't think there's any other API that requires release name processing. If it comes up, I'll refactor accordingly.

rachelmyers: This looks right to me; we don't need to look up a release by name.

src/security-rules/security-rules.ts
```diff
@@ -38,21 +38,6 @@ export interface RulesetMetadata {
...
-}
-
-interface RulesetResponse {
-  readonly name: string;
-  readonly createTime: string;
-  readonly source: {
-    readonly files: RulesFile[];
-  };
-}
-
```
rachelmyers: Are you removing these in favor of defining them in `security-rules/security-rules-api-client.ts` because they are set by the API?

hiranya911: Correct. They are being moved to the api-client module, which in this case is treated as an internal implementation module.

PR #610, opened 2019-08-01
Implemented the API for releasing rulesets

hiranya911:
Implemented the `releaseFirestoreRuleset()` API. This attempts to PATCH the existing release.

### DIFF
--- a/src/security-rules/security-rules-api-client.ts
+++ b/src/security-rules/security-rules-api-client.ts
@@ -123,15 +123,6 @@ export class SecurityRulesApiClient {
     return this.getResource<Release>(`releases/${name}`);
   }
 
-  public createRelease(name: string, rulesetName: string): Promise<Release> {
-    const request: HttpRequestConfig = {
-      method: 'POST',
-      url: `${this.url}/releases`,
-      data: this.getReleaseDescription(name, rulesetName),
-    };
-    return this.sendRequest<Release>(request);
-  }
-
   public updateRelease(name: string, rulesetName: string): Promise<Release> {
     const data = {
       release: this.getReleaseDescription(name, rulesetName),
--- a/src/security-rules/security-rules-api-client.ts
+++ b/src/security-rules/security-rules-api-client.ts
@@ -24,8 +24,8 @@ const RULES_V1_API = 'https://firebaserules.googleapis.com/v1';
 export interface Release {
   readonly name: string;
   readonly rulesetName: string;
-  readonly createTime: string;
-  readonly updateTime: string;
+  readonly createTime?: string;
+  readonly updateTime?: string;
 }
 
 export interface RulesetContent {
@@ -46,6 +46,7 @@ export interface RulesetResponse extends RulesetContent {
  */
 export class SecurityRulesApiClient {
 
+  private readonly projectIdPrefix: string;
   private readonly url: string;
 
   constructor(private readonly httpClient: HttpClient, projectId: string) {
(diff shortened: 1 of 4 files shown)


ryanpbrewster:
Under what circumstances would we expect to get a `not-found` error? In general the partner services themselves are responsible for creating the release.

hiranya911:
I'm not really sure -- just going by intuition here. If these releases are guaranteed to always exist, we can certainly get rid of the createRelease logic and further simplify the implementation. There is also a deleteRelease operation supported on the REST API, which I haven't played with. What happens if somebody deletes a release (perhaps accidentally)?

ryanpbrewster:
My mental model is that creating a release is basically "enabling" the service (e.g., linking a GCS bucket up to work with Firebase Storage), and deleting the release is the inverse (e.g., removing the bucket, disabling client access).

hiranya911:
Fair enough. I've removed the createRelease functionality from the PR.

src/security-rules/security-rules-api-client.ts
```diff
@@ -24,8 +24,8 @@ const RULES_V1_API = 'https://firebaserules.googleapis.com/v1';
 export interface Release {
   readonly name: string;
   readonly rulesetName: string;
-  readonly createTime: string;
-  readonly updateTime: string;
+  readonly createTime?: string;
+  readonly updateTime?: string;
```
rachelmyers: Is there a reason to expect that we won't get the `create` and `updateTime`s back, or you're just signaling that we don't rely on them for anything?

src/security-rules/security-rules.ts
```diff
@@ -115,6 +115,18 @@ export class SecurityRules implements FirebaseServiceInterface {
     return this.getRulesetForRelease(SecurityRules.CLOUD_FIRESTORE);
   }
 
+  /**
+   * Releases the specified ruleset as the current Cloud Firestore ruleset. This makes the specified ruleset
+   * the currently applied ruleset for Cloud Firestore.
```
rachelmyers: Total nit; seems like the second sentence is more complete and could stand on its own here.

src/security-rules/security-rules.ts
```diff
@@ -188,6 +200,21 @@ export class SecurityRules implements FirebaseServiceInterface {
         return this.getRuleset(stripProjectIdPrefix(rulesetName));
       });
   }
+
+  private releaseRuleset(ruleset: string | RulesetMetadata, releaseName: string): Promise<void> {
+    if (!validator.isNonEmptyString(ruleset) &&
+      (!validator.isNonNullObject(ruleset) || !validator.isNonEmptyString(ruleset.name))) {
```
rachelmyers: It's interesting that the name could come in either as a String or as part of the RulesetMetadata Object. Why did you opt for this and not the simpler signature of only passing in the name as a String, making it the caller's job to pluck out the name if it had an Object?

src/security-rules/security-rules-api-client.ts
hiranya911: I use the same interface as an input argument when updating/patching a release. In this case I wanted the flexibility to not specify these 2 attributes. And you're right -- we also don't rely on them for anything else.

src/security-rules/security-rules.ts
hiranya911: We can use the same private method for releasing storage rulesets. By doing it here we save a few lines from being duplicated.

src/security-rules/security-rules.ts
hiranya911: Done

PR #616, opened 2019-08-07
Implementing the remaining releaseRuleset APIs

hiranya911:
Implementing the rest of the release ruleset APIs.

```
releaseStorageRuleset(ruleset: string | RulesetMetadata, bucket?: string): Promise<void>
releaseFirestoreRulesetFromSource(source: string | Buffer): Promise<Ruleset>
releaseStorageRulesetFromSource(source: string | RulesetMetadata, bucket?: string): Promise<Ruleset>
```

### DIFF
--- a/test/unit/security-rules/security-rules.spec.ts
+++ b/test/unit/security-rules/security-rules.spec.ts
@@ -45,6 +45,24 @@ describe('SecurityRules', () => {
   };
   const CREATE_TIME_UTC = 'Fri, 08 Mar 2019 23:45:23 GMT';
 
+  const INVALID_RULESET_ERROR = new FirebaseSecurityRulesError(
+    'invalid-argument',
+    'ruleset must be a non-empty name or a RulesetMetadata object.',
+  );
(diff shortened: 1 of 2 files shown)


test/unit/security-rules/security-rules.spec.ts
```diff
@@ -377,6 +396,213 @@ describe('SecurityRules', () => {
...
+
+        return securityRules.releaseFirestoreRulesetFromSource(sources[key])
+          .then((ruleset) => {
+            expect(ruleset.name).to.equal('foo');
+            expect(ruleset.createTime).to.equal(CREATE_TIME_UTC);
+            expect(ruleset.source.length).to.equal(1);
+
+            const file = ruleset.source[0];
+            expect(file.name).equals('firestore.rules');
+            expect(file.content).equals('service cloud.firestore{\n}\n');
```
ryanpbrewster: I do not think this is a valid ruleset, where is it coming from? I see `RULES_FILE` above has content `test source {}`, does this test pass?

hiranya911: These are unit tests. They pass because they don't make any actual RPC calls.

PR #619, opened 2019-08-09
Removes tenant management methods and `forTenant` from Auth class.

bojeil-google:
Defines Auth#tenantManager().
Updates unit and integration tests.
Updates index.d.ts type definitions.


test/unit/auth/auth.spec.ts
```diff
@@ -312,39 +312,17 @@ AUTH_CONFIGS.forEach((testConfig) => {
...
-              return testAuth.forTenant(invalidTenantId as any);
-            }).to.throw('The tenant ID must be a valid non-empty string.');
-          });
-        });
-
-        it('should return a TenantAwareAuth with the expected tenant ID', () => {
-          const testAuth = new Auth(mockApp);
-          expect(testAuth.forTenant(TENANT_ID).tenantId).to.equal(TENANT_ID);
+      describe('tenantManager()', () => {
+        it('should return the expected tenant manager instance', () => {
```
wuyanna: The description is not very informative. Maybe 'should return TenantManager with same attributes'?

bojeil-google: I changed to:
`should return a TenantManager with the expected attributes`

PR #622, opened 2019-08-16
Implemented the listRulesetMetadata() API

hiranya911:
1. Added the `listRulesetMetadata()` method with support for paging through rulesets.
2. Added the `RulesetMetadataList` interface to represent the return value.

This completes the implementation of the `security-rules` module. Once this is done, we will need a couple more PRs to get this hooked up to the `admin` namespace and implement the integration tests.

### DIFF
--- a/src/security-rules/security-rules-api-client.ts
+++ b/src/security-rules/security-rules-api-client.ts
@@ -39,6 +39,11 @@ export interface RulesetResponse extends RulesetContent {
   readonly createTime: string;
 }
 
+export interface ListRulesetsResponse {
+  readonly rulesets: Array<{name: string, createTime: string}>;
+  readonly nextPageToken?: string;
+}
+
 /**
  * Class that facilitates sending requests to the Firebase security rules backend API.
  *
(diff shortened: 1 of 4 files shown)


src/security-rules/security-rules-api-client.ts
```diff
@@ -39,6 +39,11 @@ export interface RulesetResponse extends RulesetContent {
   readonly createTime: string;
 }
 
+export interface ListRulesetsResponse {
+  readonly rulesets: Array<{name: string, createTime: string}>;
```
ryanpbrewster: I assume the bracket syntax doesn't work here?

src/security-rules/security-rules-api-client.ts
```diff
@@ -119,6 +124,38 @@ export class SecurityRulesApiClient {
       });
   }
 
+  public listRulesets(pageSize: number = 100, pageToken?: string): Promise<ListRulesetsResponse> {
```
ryanpbrewster: do you even want to allow users to customize this number?

src/security-rules/security-rules-api-client.ts
```diff
@@ -119,6 +124,38 @@ export class SecurityRulesApiClient {
...
+    if (!pageToken) {
+      delete data.pageToken;
+    }
+
+    const request: HttpRequestConfig = {
+      method: 'GET',
+      url: `${this.url}/rulesets`,
+      data,
+    };
+    return this.sendRequest<ListRulesetsResponse>(request);
```
ryanpbrewster: consider providing a convenience method that repeatedly paginates until the end and then returns the entire list

src/security-rules/security-rules-api-client.ts
hiranya911: Right. Linter prefers it this way.

src/security-rules/security-rules-api-client.ts
hiranya911: I see your point. This was approved during the API review process on the grounds that other similar APIs in the admin SDK (e.g. `auth.listUsers()`) allow setting custom page sizes. So we are being consistent here. But I agree. Most users won't need it.

src/security-rules/security-rules-api-client.ts
hiranya911: We haven't really done that sort of thing in Node.js before. But that's something we can certainly consider in the future. 

I think we haven't done this for APIs like [`auth.listUsers()`](https://firebase.google.com/docs/reference/admin/node/admin.auth.Auth.html#listusers), since a project can potentially have millions of users, and we don't want an implementation that buffers all that data in memory. But since there's a hard limit on how many rulesets a project can have, it is probably not a bad idea here.

In some of the other languages like Java and Python we already have established iterable/iterator patterns to support this use case cleanly. e.g. https://firebase.google.com/docs/reference/admin/java/reference/com/google/firebase/auth/ListUsersPage.html

ryanpbrewster: clever

PR #623, opened 2019-08-16
Updates multi-tenancy integration tests.

bojeil-google:
- Enables email action link generation tests after backend bug was fixed.
- Adds email link sign-in test case.
- Fixes issues with nested multi-tenancy tests not being skipped when they are supposed to.

Confirmed multi-tenancy integration tests are passing in production.

### DIFF
--- a/test/integration/auth.spec.ts
+++ b/test/integration/auth.spec.ts
@@ -520,8 +520,12 @@ describe('admin.auth', () => {
       const rawPassword = 'password';
       const rawSalt = 'NaCl';
 
-      before(() => {
-        tenantAwareAuth = admin.auth().tenantManager().authForTenant(createdTenantId);
+      before(function() {
+        if (!createdTenantId) {
+          this.skip();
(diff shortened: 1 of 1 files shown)


test/integration/auth.spec.ts
```diff
@@ -520,8 +520,12 @@ describe('admin.auth', () => {
       const rawPassword = 'password';
       const rawSalt = 'NaCl';
 
-      before(() => {
-        tenantAwareAuth = admin.auth().tenantManager().authForTenant(createdTenantId);
+      before(function() {
+        if (!createdTenantId) {
+          this.skip();
```
hiranya911: When does this happen? if the tenant creation fails in an earlier test?

bojeil-google: It seems the skip test is not getting applied for nested tests. So i added a check when no tenant is created (typically when the test is skipped or if tenant creation fails) to skip the nested tests.

PR #625, opened 2019-08-19
Adding the rules API to the public API surface

hiranya911:
1. Exposed the `admin.securityRules` namespace and the getter.
2. Exposed `securityRules` getter from `FirebaseApp`.
3. Updated the `index.d.ts` file.

The comments in the `d.ts` file gets compiled into our API reference docs.

### DIFF
--- a/docgen/content-sources/node/toc.yaml
+++ b/docgen/content-sources/node/toc.yaml
@@ -178,6 +178,20 @@ toc:
   - title: "ShaCertificate"
     path: /docs/reference/admin/node/admin.projectManagement.ShaCertificate
 
+- title: "admin.securityRules"
+  path: /docs/reference/admin/node/admin.securityRules
+  section:
+  - title: "RulesFile"
+    path: /docs/reference/admin/node/admin.securityRules.RulesFile
+  - title: "Ruleset"
+    path: /docs/reference/admin/node/admin.securityRules.Ruleset
+  - title: "RulesetMetadata"
+    path: /docs/reference/admin/node/admin.securityRules.RulesetMetadata
+  - title: "RulesetMetadataList"
+    path: /docs/reference/admin/node/admin.securityRules.RulesetMetadataList
+  - title: "SecurityRules"
+    path: /docs/reference/admin/node/admin.securityRules.SecurityRules
+
 - title: "admin.storage"
   path: /docs/reference/admin/node/admin.storage
   section:
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -454,6 +454,7 @@ declare namespace admin.app {
     instanceId(): admin.instanceId.InstanceId;
     messaging(): admin.messaging.Messaging;
     projectManagement(): admin.projectManagement.ProjectManagement;
+    securityRules(): admin.securityRules.SecurityRules;
     storage(): admin.storage.Storage;
 
     /**
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -386,10 +386,10 @@ declare namespace admin {
   * var otherSecurityRules = admin.securityRules(otherApp);
   * ```
   *
-  * @param app Optional app whose `SecurityRules` service
-  *     to return. If not provided, the default `SecurityRules` service will
+  * @param app Optional app to return the `SecurityRules` service
+  *     for. If not provided, the default `SecurityRules` service will
   *     be returned.
-  * @return The default `SecurityRules` service if no app is provided or the
+  * @return The default `SecurityRules` service if no app is provided, or the
   *   `SecurityRules` service associated with the provided app.
   */
   function securityRules(app?: admin.app.App): admin.securityRules.SecurityRules;
@@ -5213,7 +5213,7 @@ declare namespace admin.securityRules {
   }
 
   /**
-   * Required metadata associated with a Ruleset.
+   * Required metadata associated with a ruleset.
    */
   interface RulesetMetadata {
     /**
@@ -5249,7 +5249,7 @@ declare namespace admin.securityRules {
   }
 
   /**
-   * The Firebase SecurityRules service interface.
+   * The Firebase `SecurityRules` service interface.
    *
    * Do not call this constructor directly. Instead, use
    * [`admin.securityRules()`](admin.securityRules#securityRules).
@@ -5259,8 +5259,8 @@ declare namespace admin.securityRules {
 
     /**
      * Creates a {@link admin.securityRules.RulesFile `RuleFile`} with the given name
-     * and source. Throws if any of the arguments are invalid. This is a local operation,
-     * and does not involve any network API calls.
+     * and source. Throws an error if any of the arguments are invalid. This is a local
+     * operation, and does not involve any network API calls.
      *
      *  @example
      * ```javascript
@@ -5270,7 +5270,7 @@ declare namespace admin.securityRules {
      * ```
      *
      * @param name Name to assign to the rules file. This is usually a short file name that
-     *   helps identify the file in a Ruleset.
+     *   helps identify the file in a ruleset.
      * @param source Contents of the rules file.
      * @return A new rules file instance.
      */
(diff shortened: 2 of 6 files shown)


src/index.d.ts
```diff
@@ -363,6 +363,37 @@ declare namespace admin {
...
+  * @example
+  * ```javascript
+  * // Get the SecurityRules service for a given app
+  * var otherSecurityRules = admin.securityRules(otherApp);
+  * ```
+  *
+  * @param app Optional app whose `SecurityRules` service
+  *     to return. If not provided, the default `SecurityRules` service will
+  *     be returned.
+  * @return The default `SecurityRules` service if no app is provided or the
```
egilmorez: I feel like this would read better with a comma at "provided, or"

src/index.d.ts
```diff
@@ -363,6 +363,37 @@ declare namespace admin {
...
+  * ```
+  *
+  * @example
+  * ```javascript
+  * // Get the SecurityRules service for a given app
+  * var otherSecurityRules = admin.securityRules(otherApp);
+  * ```
+  *
+  * @param app Optional app whose `SecurityRules` service
+  *     to return. If not provided, the default `SecurityRules` service will
```
egilmorez: Suggest " Optional app to return a `SecurityRules` service for."

src/index.d.ts
```diff
@@ -5169,6 +5200,196 @@ declare namespace admin.projectManagement {
...
+   * [`securityRules.createRulesFileFromSource()`](admin.securityRules.SecurityRules#createRulesFileFromSource)
+   * method to create new instances of this type.
+   */
+  interface RulesFile {
+    readonly name: string;
+    readonly content: string;
+  }
+
+  /**
+   * Required metadata associated with a Ruleset.
```
egilmorez: I think this should be literal '`Ruleset,` or not capitalized.

src/index.d.ts
```diff
@@ -5169,6 +5200,196 @@ declare namespace admin.projectManagement {
...
+    readonly rulesets: RulesetMetadata[];
+
+    /**
+     * The next page token if available. This is needed to retrieve the next batch.
+     */
+    readonly nextPageToken?: string;
+  }
+
+  /**
+   * The Firebase SecurityRules service interface.
```
egilmorez: Suggest `SecurityRules`

src/index.d.ts
```diff
@@ -5169,6 +5200,196 @@ declare namespace admin.projectManagement {
...
+   *
+   * Do not call this constructor directly. Instead, use
+   * [`admin.securityRules()`](admin.securityRules#securityRules).
+   */
+  interface SecurityRules {
+    app: admin.app.App;
+
+    /**
+     * Creates a {@link admin.securityRules.RulesFile `RuleFile`} with the given name
+     * and source. Throws if any of the arguments are invalid. This is a local operation,
```
egilmorez: Suggest "Throws an error if" or name the specific error.

src/index.d.ts
```diff
@@ -5169,6 +5200,196 @@ declare namespace admin.projectManagement {
...
+     *
+     *  @example
+     * ```javascript
+     * const source = '// Some rules source';
+     * const rulesFile = admin.securityRules().createRulesFileFromSource(
+     *   'firestore.rules', source);
+     * ```
+     *
+     * @param name Name to assign to the rules file. This is usually a short file name that
+     *   helps identify the file in a Ruleset.
```
egilmorez: Non-capped or literal?  Probably just "ruleset."

src/index.d.ts
```diff
@@ -5169,6 +5200,196 @@ declare namespace admin.projectManagement {
...
+     * @param source Contents of the rules file.
+     * @return A new rules file instance.
+     */
+    createRulesFileFromSource(name: string, source: string | Buffer): RulesFile;
+
+    /**
+     * Creates a new {@link admin.securityRules.Ruleset `Ruleset`} from the given
+     * {@link admin.securityRules.RulesFile `RuleFile`}.
+     *
+     * @param file Rules file to include in the new Ruleset.
```
egilmorez: Probably `Ruleset` best here.

src/index.d.ts
```diff
@@ -5169,6 +5200,196 @@ declare namespace admin.projectManagement {
...
+     * Creates a new {@link admin.securityRules.Ruleset `Ruleset`} from the given
+     * source, and applies it to Cloud Firestore.
+     *
+     * @param source Rules source to apply.
+     * @return A promise that fulfills when the ruleset is created and released.
+     */
+    releaseFirestoreRulesetFromSource(source: string | Buffer): Promise<Ruleset>;
+
+    /**
+     * Makes the specified {@link admin.securityRules.Ruleset `Ruleset`} the currently
```
egilmorez: Would it work to say just "Applies the specified ruleset?"

Understand if you think we we need to clarify that it's a replacement not an addition.

src/index.d.ts
```diff
@@ -5169,6 +5200,196 @@ declare namespace admin.projectManagement {
...
+     * @param bucket Optional name of the Cloud Storage bucket to apply the rules on. If
+     *   not specified, applies the ruleset on the default bucket configured via
+     *   {@link admin.AppOptions `AppOptions`}.
+     * @return A promise that fulfills when the ruleset is created and released.
+     */
+    releaseStorageRulesetFromSource(
+      source: string | Buffer, bucket?: string): Promise<Ruleset>;
+
+    /**
+     * Makes the specified {@link admin.securityRules.Ruleset `Ruleset`} the currently
```
egilmorez: Would it work to say just "Applies the specified ruleset?"

Understand if you think we we need to clarify that it's a replacement not an addition.

src/index.d.ts
hiranya911: Creating a ruleset and applying it are 2 separate ops. This does both. I think we need to be clear of that.

src/index.d.ts

src/index.d.ts

src/index.d.ts

src/index.d.ts

src/index.d.ts

src/index.d.ts

src/index.d.ts

src/index.d.ts

docgen/content-sources/node/toc.yaml
```diff
@@ -178,6 +178,20 @@ toc:
   - title: "ShaCertificate"
     path: /docs/reference/admin/node/admin.projectManagement.ShaCertificate
 
+- title: "admin.securityRules"
```
ryanpbrewster: I'll defer to you and your other reviewers here, but we use `FirebaseRules` in the same way we use `FirebaseAuth` and `FirebaseDatabase`. The extra "security" word here breaks conventions. ((Consider: what the object will be called in the Java admin SDK?))

hiranya911: Current naming convention used is: `admin.securityRules()`, `SecurityRules`, (Java: `FirebaseSecurityRules`).

It sounds like you're suggesting: `admin.rules()`, `Rules`, (Java: `FirebaseRules`)

Is that correct? I don't have preference. Let's see what @rachelmyers and others think. The existing documentation uses "Firebase Security Rules" pretty much everywhere: https://firebase.google.com/docs/rules

src/index.d.ts
```diff
@@ -5169,6 +5201,196 @@ declare namespace admin.projectManagement {
...
+    createRulesFileFromSource(name: string, source: string | Buffer): RulesFile;
+
+    /**
+     * Creates a new {@link admin.securityRules.Ruleset `Ruleset`} from the given
+     * {@link admin.securityRules.RulesFile `RuleFile`}.
+     *
+     * @param file Rules file to include in the new `Ruleset`.
+     * @returns A promise that fulfills with the newly created `Ruleset`.
+     */
+    createRuleset(file: RulesFile): Promise<Ruleset>;
```
rachelmyers: Was `createRuleset` meant to allow multiple files, so `creeateRuleset(files...: RulesFile): Promise<Ruleset>`, instead?

hiranya911: In #607 we decided to only accept one `RulesFile` in this API (since the backend impl currently doesn't allow multiple files). In the future we can change this into a varargs if necessary, as a non-breaking change.

hiranya911:
I'll merge this as is. We still have time to revise the rules vs security rules argument in a separate PR.

PR #597, opened 2019-07-20
feat(fcm): Added support for specifying the analytics label for notifications.

chemidy:
- add analytics_label
#580

PR #629, opened 2019-08-21
Allow `fcmOptions` in payload keys

parndt:
In #597 support was added for `fcmOptions` for specifying the analytics label for notifications.

Unfortunately, the list of valid payload keys wasn't extended to support it.

### DIFF
--- a/src/messaging/messaging-types.ts
+++ b/src/messaging/messaging-types.ts
@@ -182,6 +182,7 @@ export interface DataMessagePayload {
 /* Payload for fcmOptions messages */
 export interface FcmOptionsPayload {
   analyticsLabel?: string;
+  [key: string]: string;
 }
 
 /* Payload for notification messages */
--- a/src/messaging/messaging-types.ts
+++ b/src/messaging/messaging-types.ts
@@ -179,6 +179,11 @@ export interface DataMessagePayload {
   [key: string]: string;
 }
 
+/* Payload for fcmOptions messages */
+export interface FcmOptionsPayload {
+  analyticsLabel?: string;
+}
+
 /* Payload for notification messages */
 export interface NotificationMessagePayload {
   tag?: string;
@@ -196,9 +201,10 @@ export interface NotificationMessagePayload {
   [other: string]: string;
 }
 
-/* Composite messaging payload (data and notification payloads are both optional) */
+/* Composite messaging payload (data, fcmOptions, and notification payloads are all optional) */
 export interface MessagingPayload {
   data?: DataMessagePayload;
+  fcmOptions?: FcmOptionsPayload;
   notification?: NotificationMessagePayload;
 }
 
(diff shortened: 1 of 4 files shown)


parndt:
@googlebot I signed it!

parndt:
@hiranya911 I tried to add all the right specs for this to ensure that this change is correct and avoids regressions, which pass locally, but I note that `gulp build` fails because of a missing index signature. Are you able to provide guidance here for me, please?

```
src/messaging/messaging.ts(775,33): error TS7017: Element implicitly has an 'any' type because type 'DataMessagePayload | FcmOptionsPayload | NotificationMessagePayload' has no index signature.
```

parndt:
this would work but I don't think it's desired:
```typescript
/* Payload for fcmOptions messages */
export interface FcmOptionsPayload {
  analyticsLabel?: string;
  [key: string]: string;
}
```

parndt:
I opted for this implementation for the index signature:

```typescript
/* Payload for fcmOptions messages */
export interface FcmOptionsPayload {
  analyticsLabel?: string;
  [key: string]: string;
}
```

you're welcome to edit the PR to adjust if this isn't suitable :smile: or just let me know what to change.

Thanks!

hiranya911:
I have a few concerns with this change.

1. I don't think the legacy FCM API supports `fcmOptions` (see https://firebase.google.com/docs/cloud-messaging/http-server-ref). 
2. Even if it did, I don't know if we should be adding new features to the legacy API. 

@chong-shao wdyt?

parndt:
Thanks @hiranya911 for the review.

For some extra context which may help with deciding about this PR, here's how we're using this code inside a cloud function using [messaging.sendToDevice](https://firebase.google.com/docs/reference/admin/node/admin.messaging.Messaging#send-todevice):

```js
const admin = require('firebase-admin')

admin.messaging().sendToDevice(
  pushToken,
  data: {
    custom_notification: JSON.stringify(
      Object.assign({ deepLink, show_in_foreground: true }, notification)
    ),
    deepLink,
  },
  fcmOptions: {
    analyticsLabel: 'test.analytics.label',
  },
  notification,
)
```

We saw this error:

```js
{ Error: Messaging payload contains an invalid "fcmOptions" property. Valid properties are "data" and "notification".
```

I guess that is triggering usage of the legacy APIs when using [MessagingPayload](https://firebase.google.com/docs/reference/admin/node/admin.messaging.MessagingPayload); is there another way to do this that uses the non-legacy API and allows us to use `fcmOptions` with `analyticsLabel`?
Does it involve [this migration process](https://firebase.google.com/docs/cloud-messaging/migrate-v1#example_simple_notification_message) ?

Thanks for your time! :smile:

hiranya911:
You're using the legacy FCM APIs (`sendToToken()`, `sendToDevice()`, `sendToDeviceGroup()`). Those APIs don't support `fcmOptions`. You have to use the `send()` API for that. We have already added `fcmOptions` support there.

hiranya911:
@parndt please take a look at the Node.js examples on this page and update your code to use the new API: https://firebase.google.com/docs/cloud-messaging/send-message

The old API doesn't support `fcmOptions`.

parndt:
Thanks a lot! :bowing_man: that's exactly what I needed.

PR #648, opened 2019-09-13
feat(fcm): Add image in notification support

### LINKED ISSUE #598: FR: add image option in notification
There are new options for image in notification : https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages?authuser=0#notification

lahirumaramba:
### Discussion

  * Add image in notification support

### Testing

  * Add unit tests and modified integration test to reflect this change

### API Changes

  * In Firebase Cloud Messaging, added 'image' field in the general Notification class, the AndroidNotification class, and the ApnsFcmOptions class.

RELEASE NOTE: Added new APIs for specifying an image URL in notifications.

Resolves #598

### DIFF
--- a/src/messaging/messaging-types.ts
+++ b/src/messaging/messaging-types.ts
@@ -59,6 +59,7 @@ export interface MulticastMessage extends BaseMessage {
 export interface Notification {
   title?: string;
   body?: string;
+  image?: string;
 }
 
 export interface FcmOptions {
@@ -143,6 +144,7 @@ export interface ApsAlert {
 
 export interface ApnsFcmOptions {
   analyticsLabel?: string;
+  image?: string;
 }
 
 export interface AndroidConfig {
@@ -162,6 +164,7 @@ export interface AndroidNotification {
   color?: string;
   sound?: string;
   tag?: string;
+  image?: string;
   clickAction?: string;
   bodyLocKey?: string;
   bodyLocArgs?: string[];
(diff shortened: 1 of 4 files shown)


test/unit/messaging/messaging.spec.ts
```diff
@@ -2334,6 +2334,37 @@ describe('Messaging', () => {
       });
     });
 
+    const invalidImages = ['a', 'foo', 'image.jpg'];
```
hiranya911: Also test for empty string

test/unit/messaging/messaging.spec.ts
```diff
@@ -2334,6 +2334,37 @@ describe('Messaging', () => {
...
+            },
+          },
+        };
+        expect(() => {
+          messaging.send(message);
+        }).to.throw('android.notification.image must be a valid URL string');
+      });
+    });
+
+    invalidImages.forEach((image) => {
```
hiranya911: Lets move this somewhere else in this file. This section seems to be all about android config tests.

src/messaging/messaging-types.ts
```diff
@@ -59,6 +59,7 @@ export interface MulticastMessage extends BaseMessage {
 export interface Notification {
   title?: string;
   body?: string;
+  image?: string;
```
hiranya911: We called these `imageUrl` in Java. Might want to do the same here.

lahirumaramba: @hiranya911 
Do you think it is sufficient to change just the public API `index.d.ts` here or should we use the same parameter name `imageUrl` everywhere in our code?

PR #659, opened 2019-09-25
Remove unused local variables.

rsgowman:
Also add option to tsconfig.json to prevent re-occurrences.


tsconfig.json
```diff
@@ -3,6 +3,7 @@
     "module": "commonjs",
     "target": "es5",
     "noImplicitAny": true,
+    "noUnusedLocals": true,
```
hiranya911: Nice 👍

PR #660, opened 2019-09-25
Remove (base64) 'REDACTED' passwords from user records.

rsgowman:
These values *look* like passwords hashes, but aren't, leading to
potential confusion.

Additionally, added docs to CONTRIBUTING.md detailing how to add the
permission that causes password hashes to be properly returned as well
as adjusting the test failure message should the developer not add that
permission.

b/141189502

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -161,7 +161,9 @@ Then set up your Firebase/GCP project as follows:
    Dashboard" from the main menu, and click the "ENABLE APIS AND SERVICES"
    button. Search for and enable the "Identity and Access Management (IAM)
    API".
-4. Grant your service account the 'Firebase Authentication Admin' role:
+4. Grant your service account the 'Firebase Authentication Admin' role. This is
+   required to ensure that exported user records contain the password hashes of
+   the user accounts:
    1. Go to [Google Cloud Platform Console / IAM & admin](https://console.cloud.google.com/iam-admin).
    2. Find your service account in the list, and click the 'pencil' icon to edit it's permissions.
    3. Click 'ADD ANOTHER ROLE' and choose 'Firebase Authentication Admin'.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -153,13 +153,19 @@ Then set up your Firebase/GCP project as follows:
    to set up Firestore either in the locked mode or in the test mode.
 2. Enable password auth: Select "Authentication" from the "Develop" menu in
    Firebase Console. Select the "Sign-in method" tab, and enable the
-   "Email/Password" sign-in method.
+   "Email/Password" sign-in method, including the Email link (passwordless
+   sign-in) option.
 3. Enable the IAM API: Go to the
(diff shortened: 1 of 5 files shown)


src/auth/user-record.ts
```diff
@@ -173,7 +178,16 @@ export class UserRecord {
       providerData.push(new UserInfo(entry));
     }
     utils.addReadonlyGetter(this, 'providerData', providerData);
-    utils.addReadonlyGetter(this, 'passwordHash', response.passwordHash);
+
+    // If the password hash is redacted (probably due to missing permissions)
+    // then clear it out, similar to how the salt is returned. (Otherwise, it
+    // *looks* like a b64-encoded hash is present, which is confusing.)
+    if (response.passwordHash === B64_REDACTED) {
+      utils.addReadonlyGetter(this, 'passwordHash', undefined);
```
bojeil-google: Please add a unit test for this. When a user record is initialized with a server response containing a redacted password, confirm it is cleared.

CONTRIBUTING.md
```diff
@@ -153,13 +153,19 @@ Then set up your Firebase/GCP project as follows:
...
-   "Email/Password" sign-in method.
+   "Email/Password" sign-in method, including the Email link (passwordless
+   sign-in) option.
 3. Enable the IAM API: Go to the
    [Google Cloud Platform Console](https://console.cloud.google.com) and make
    sure your Firebase/GCP project is selected. Select "APIs & Services >
    Dashboard" from the main menu, and click the "ENABLE APIS AND SERVICES"
    button. Search for and enable the "Identity and Access Management (IAM)
    API".
+4. Grant your service account the 'Firebase Authentication Admin' role:
```
hiranya911: Can we also add a note here indicating why this permission is needed? 

"This is required to ensure that exported user records contain the password hashes of the user accounts."

rsgowman: Done.

src/auth/user-record.ts
rsgowman: Done.

bojeil-google: Thanks for adding to auth.spec.ts. Can you also add to user-record.spec.ts, the test associated with this class?

rsgowman: Done. I've removed the test from auth.spec.ts, as I think with this change, it's become mostly redundant. Though we can add it back in if you'd prefer to keep it.

test/unit/auth/user-record.spec.ts
```diff
@@ -653,6 +653,15 @@ describe('UserRecord', () => {
     });
   });
 
+  describe('clears REDACTED passwordHash', () => {
```
bojeil-google: This seems like a test of `passwordHash` getter.
Can you move it next to `it('should return expected undefined passwordHash', () => {`
You can define something like: `it('should clear REDACTED passwordHash', () => {`

rsgowman: Done

PR #665, opened 2019-10-04
Migrate Project Config service calls to v2.

nrsim:


test/unit/auth/auth-api-request.spec.ts
```diff
@@ -2515,7 +2515,7 @@ AUTH_REQUEST_HANDLER_TESTS.forEach((handler) => {
 
     describe('getOAuthIdpConfig()', () => {
       const providerId = 'oidc.provider';
-      const path = handler.path('v2beta1', `/oauthIdpConfigs/${providerId}`, 'project_id');
```
bojeil-google: We also want to rename the APIs to be consistent with the SAML counterparts here and elsewhere.
`oauthIdpConfigs` -> `inboundOauthConfigs`
Check the proto file for confirmation.

src/auth/auth-api-request.ts
```diff
@@ -1644,7 +1644,7 @@ export class TenantAwareAuthRequestHandler extends AbstractAuthRequestHandler {
    * @return {AuthResourceUrlBuilder} A new project config resource URL builder instance.
    */
   protected newProjectConfigUrlBuilder(): AuthResourceUrlBuilder {
-    return new TenantAwareAuthResourceUrlBuilder(this.projectId, 'v2beta1', this.tenantId);
+    return new TenantAwareAuthResourceUrlBuilder(this.projectId, 'admin/v2', this.tenantId);
```
bojeil-google: It seems they plan to remove the `admin/` prefix as it is not one platform compliant (it will break the API discovery doc).
They are working on it now.

test/unit/auth/auth-api-request.spec.ts
nrsim: Seems like oauthIdpConfigs was kept for OnePlatform compliancy.

src/auth/auth-api-request.ts
nrsim: Removed the prefix from the calls.

PR #673, opened 2019-10-18
Break test order dependencies in security-rules integration tests.

rsgowman:
Previously, this would work:
yarn test:integration --updateRules --grep 'admin.securityRules'

but this would not:
yarn test:integration --updateRules --grep 'admin.securityRules.getRuleset\(\).rejects with not-found'

because that test relied on side effects from earier tests, making this
test suite difficult to debug.

### DIFF
--- a/test/integration/security-rules.spec.ts
+++ b/test/integration/security-rules.spec.ts
@@ -45,7 +45,6 @@ const RULESET_NAME_PATTERN = /[0-9a-zA-Z-]+/;
 
 describe('admin.securityRules', () => {
 
-  let testRuleset: admin.securityRules.Ruleset = null;
   const rulesetsToDelete: string[] = [];
 
   function scheduleForDelete(ruleset: admin.securityRules.Ruleset) {
@@ -65,7 +64,7 @@ describe('admin.securityRules', () => {
     return Promise.all(promises);
   }
 
-  after(() => {
+  afterEach(() => {
     return deleteTempRulesets();
   });
 
@@ -91,7 +90,6 @@ describe('admin.securityRules', () => {
         RULES_FILE_NAME, SAMPLE_FIRESTORE_RULES);
       return admin.securityRules().createRuleset(rulesFile)
         .then((ruleset) => {
-          testRuleset = ruleset;
           scheduleForDelete(ruleset);
           verifyFirestoreRuleset(ruleset);
         });
@@ -105,23 +103,40 @@ describe('admin.securityRules', () => {
     });
   });
 
+  function getRandomInt(max: number) {
+    return Math.floor(Math.random() * Math.floor(max));
+  }
+
+  function createTemporaryRuleset(): Promise<admin.securityRules.Ruleset> {
+    const name = "random_name_" + getRandomInt(100000) + ".rules";
+    const rulesFile = admin.securityRules().createRulesFileFromSource(name, SAMPLE_FIRESTORE_RULES);
+    return admin.securityRules().createRuleset(rulesFile)
+      .then((ruleset) => {
+        scheduleForDelete(ruleset);
+        return ruleset;
+      });
+  }
+
   describe('getRuleset()', () => {
     it('rejects with not-found when the Ruleset does not exist', () => {
-      const name = 'e1212' + testRuleset.name.substring(5);
-      return admin.securityRules().getRuleset(name)
+      const uuid = '00000000-1111-2222-3333-444444444444';
+      return admin.securityRules().getRuleset(uuid)
         .should.eventually.be.rejected.and.have.property('code', 'security-rules/not-found');
     });
 
     it('rejects with invalid-argument when the Ruleset name is invalid', () => {
-      return admin.securityRules().getRuleset('invalid')
+      return admin.securityRules().getRuleset('invalid uuid')
         .should.eventually.be.rejected.and.have.property('code', 'security-rules/invalid-argument');
     });
 
     it('resolves with existing Ruleset', () => {
-      return admin.securityRules().getRuleset(testRuleset.name)
-        .then((ruleset) => {
-          verifyFirestoreRuleset(ruleset);
-        });
+      return createTemporaryRuleset()
+        .then((expectedRuleset) =>
+          admin.securityRules().getRuleset(expectedRuleset.name)
+            .then((actualRuleset) => {
+              expect(actualRuleset).to.deep.equal(expectedRuleset);
+            })
+        );
     });
   });
 
@@ -137,7 +152,7 @@ describe('admin.securityRules', () => {
       return admin.securityRules().releaseFirestoreRuleset(oldRuleset);
     }
 
-    after(() => {
+    afterEach(() => {
       return revertFirestoreRuleset();
     });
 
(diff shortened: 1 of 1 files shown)


test/integration/security-rules.spec.ts
```diff
@@ -105,23 +103,40 @@ describe('admin.securityRules', () => {
     });
   });
 
+  function getRandomInt(max: number) {
```
hiranya911: Move to the top where the other helper functions are.

test/integration/security-rules.spec.ts
```diff
@@ -105,23 +103,40 @@ describe('admin.securityRules', () => {
...
+        scheduleForDelete(ruleset);
+        return ruleset;
+      });
+  }
+
   describe('getRuleset()', () => {
     it('rejects with not-found when the Ruleset does not exist', () => {
-      const name = 'e1212' + testRuleset.name.substring(5);
-      return admin.securityRules().getRuleset(name)
+      const uuid = '00000000-1111-2222-3333-444444444444';
```
hiranya911: Nit: Lets call it `nonExistingName`.

test/integration/security-rules.spec.ts
```diff
@@ -257,26 +276,27 @@ describe('admin.securityRules', () => {
 
   describe('deleteRuleset()', () => {
     it('rejects with not-found when the Ruleset does not exist', () => {
-      const name = 'e1212' + testRuleset.name.substring(5);
-      return admin.securityRules().deleteRuleset(name)
+      const uuid = '00000000-1111-2222-3333-444444444444';
```
hiranya911: `nonExistingName`

test/integration/security-rules.spec.ts
```diff
@@ -257,26 +276,27 @@ describe('admin.securityRules', () => {
...
-          testRuleset = null;
-        });
+      return createTemporaryRuleset().then((ruleset) => {
+        return admin.securityRules().deleteRuleset(ruleset.name)
+          .then(() => {
+            return admin.securityRules().getRuleset(ruleset.name)
+              .should.eventually.be.rejected.and.have.property('code', 'security-rules/not-found');
+          })
+          .then(() => {
+            unscheduleForDelete(ruleset); // Already deleted.
```
hiranya911: Is the `testRuleset` variable still referenced anywhere? If not lets remove that.

test/integration/security-rules.spec.ts
```diff
@@ -137,7 +152,7 @@ describe('admin.securityRules', () => {
       return admin.securityRules().releaseFirestoreRuleset(oldRuleset);
     }
 
-    after(() => {
+    afterEach(() => {
```
hiranya911: This change is not really needed right. I'd say lets either keep it as `after` or lets rename the helper to something like `revertFirestoreRulesetIfModified` so the logic of the test suite becomes clearer.

test/integration/security-rules.spec.ts
```diff
@@ -105,23 +103,40 @@ describe('admin.securityRules', () => {
     });
   });
 
+  function getRandomInt(max: number) {
+    return Math.floor(Math.random() * Math.floor(max));
+  }
+
+  function createTemporaryRuleset(): Promise<admin.securityRules.Ruleset> {
+    const name = 'random_name_' + getRandomInt(100000) + '.rules';
```
hiranya911: Rules file names don't really have to be unique across rulesets. Feel free to use `firestore.rules`.

test/integration/security-rules.spec.ts
rsgowman: Done.

test/integration/security-rules.spec.ts
rsgowman: This wasn't to make them unique across rulesets, but rather for the following:
* it allows you to run tests in parallel without them tripping over each other. (Theoretical benefit only; I don't think mocha allows parallel execution of tests.)
* it allows you to call it twice to generate two different rulesets, like I'm doing on line 258, now 269... unless you get really unlucky, in which case the test will flake. Yikes. Fixed.
* it ensures`[*]` that a test that accidentally leaves a rules file behind doesn't impact the other tests. (`[*]` - or makes it unlikely to occur.)

I've added a comment to this effect, referencing the last bullet point.

test/integration/security-rules.spec.ts
rsgowman: Done.

test/integration/security-rules.spec.ts
rsgowman: I like keeping it as an afterEach. (It's not an issue here, but there are ordering implications to mixing after and afterEach which I tripped on elsewhere in this PR.) Therefore I've renamed the helper to revertFirestoreRulesetIfModified.

test/integration/security-rules.spec.ts
rsgowman: Done.

test/integration/security-rules.spec.ts
rsgowman: > Is the testRuleset variable still referenced anywhere?

No.

> If not lets remove that.

Remove the testRuleset? It's already been removed...? Or were you referring to the unscheduleForDelete call? That call is currently necessary since the deleteTempRulesets() will otherwise error out due to not being able to find the entity. We could change that to make it idempotent though. (Let me know if you'd like me to do that.)

test/integration/security-rules.spec.ts
```diff
@@ -65,7 +64,32 @@ describe('admin.securityRules', () => {
...
+   * completion of the test. (The random name "ensures" that if the ruleset
+   * doesn't get cleaned up for any reason, then it will be unlikely to impact
+   * other tests, including a second run of the same test.)
+   *
+   * @param prefix Optional prefix that will be prefixed to the name of the
+   *   rules file. Useful if your test creates multiple temporary rulesets and
+   *   you want to ensure that they don't collide.
+   */
+  function createTemporaryRuleset(prefix?: string): Promise<admin.securityRules.Ruleset> {
+    const name = (prefix ? prefix : '') + 'random_name_' + getRandomInt(100000) + '.rules';
```
hiranya911: This is really not necessary. Rulesets are given a unique name by the Rules API. Rules file names don't have to be globally unique. They just have to be unique within a given ruleset. Since this temp ruleset has only one rules file, we can just give it a fixed name like `firestore.rules`.

test/integration/security-rules.spec.ts
```diff
@@ -185,7 +211,7 @@ describe('admin.securityRules', () => {
       return admin.securityRules().releaseStorageRuleset(oldRuleset);
     }
 
-    after(() => {
+    afterEach(() => {
```
hiranya911: Rename this to `revertStorageRulesetIfModified`

test/integration/security-rules.spec.ts
hiranya911: Ah, I didn't spot that the variable has already been removed. Thanks.

test/integration/security-rules.spec.ts
rsgowman: Oh, I understand now. Done.

test/integration/security-rules.spec.ts
rsgowman: Done.

PR #676, opened 2019-10-18
Enable 'strictNullChecks' ts compiler option.

rsgowman:
In the majority of cases, I simply turned stuff like this:

```
function fname(param: T) {
  if (!param) throw Error("invalid param; must not be nullish");
  ...
}
```

into this:

```
function fname(param: T|null|undefined) {
  if (!param) throw Error("invalid param; must not be nullish");
  ...
}
```

A more sensible approach would be eliminate the possibility of
null/undef and eliminate the checks too, but that's a bit more involved
and this patch is already enormous. But I think these sorts of
improvements could be taken advantage of opportunistically as the code
is worked on in the future.

### DIFF
--- a/src/auth/auth-api-request.ts
+++ b/src/auth/auth-api-request.ts
@@ -693,7 +693,7 @@ export abstract class AbstractAuthRequestHandler {
    * @return {string|null} The error code if present; null otherwise.
    */
   private static getErrorCode(response: any): string | null {
-    return (validator.isNonNullObject(response) && (response as any).error && (response as any).error.message) || null;
+    return (validator.isNonNullObject(response) && response.error && response.error.message) || null;
   }
 
   /**
@@ -1068,9 +1068,15 @@ export abstract class AbstractAuthRequestHandler {
     let request = {requestType, email, returnOobLink: true};
     // ActionCodeSettings required for email link sign-in to determine the url where the sign-in will
     // be completed.
+    if (typeof actionCodeSettings === 'undefined' && requestType === 'EMAIL_SIGNIN') {
+      throw new FirebaseAuthError(
+        AuthClientErrorCode.INVALID_ARGUMENT,
+        "`actionCodeSettings` is required when `requestType` === 'EMAIL_SIGNIN'",
+      );
+    }
     if (typeof actionCodeSettings !== 'undefined' || requestType === 'EMAIL_SIGNIN') {
       try {
-        const builder = new ActionCodeSettingsBuilder(actionCodeSettings);
+        const builder = new ActionCodeSettingsBuilder(actionCodeSettings!);
         request = deepExtend(request, builder.buildRequest());
       } catch (e) {
         return Promise.reject(e);
--- a/src/auth/auth-api-request.ts
+++ b/src/auth/auth-api-request.ts
@@ -97,7 +97,7 @@ class AuthResourceUrlBuilder {
    * @param {string} version The endpoint API version.
    * @constructor
    */
-  constructor(protected projectId: string, protected version: string = 'v1') {
+  constructor(protected projectId: string | null, protected version: string = 'v1') {
     this.urlFormat = FIREBASE_AUTH_BASE_URL_FORMAT;
   }
 
@@ -132,7 +132,7 @@ class TenantAwareAuthResourceUrlBuilder extends AuthResourceUrlBuilder {
    * @param {string} tenantId The tenant ID.
    * @constructor
    */
-  constructor(protected projectId: string, protected version: string, protected tenantId: string) {
+  constructor(protected projectId: string | null, protected version: string, protected tenantId: string) {
     super(projectId, version);
     this.urlFormat = FIREBASE_AUTH_TENANT_URL_FORMAT;
   }
@@ -683,7 +683,7 @@ const LIST_INBOUND_SAML_CONFIGS = new ApiSettings('/inboundSamlConfigs', 'GET')
  * Class that provides the mechanism to send requests to the Firebase Auth backend endpoints.
  */
 export abstract class AbstractAuthRequestHandler {
-  protected readonly projectId: string;
+  protected readonly projectId: string | null;
   protected readonly httpClient: AuthorizedHttpClient;
   private authUrlBuilder: AuthResourceUrlBuilder;
   private projectConfigUrlBuilder: AuthResourceUrlBuilder;
@@ -693,7 +693,7 @@ export abstract class AbstractAuthRequestHandler {
    * @return {string|null} The error code if present; null otherwise.
    */
   private static getErrorCode(response: any): string | null {
-    return (validator.isNonNullObject(response) && response.error && (response.error as any).message) || null;
+    return (validator.isNonNullObject(response) && (response as any).error && (response as any).error.message) || null;
   }
 
   /**
@@ -844,7 +844,7 @@ export abstract class AbstractAuthRequestHandler {
     }
     // If no remaining user in request after client side processing, there is no need
     // to send the request to the server.
-    if (request.users.length === 0) {
+    if (!request.users || request.users.length === 0) {
       return Promise.resolve(userImportBuilder.buildResponse([]));
     }
     return this.invokeRequestHandler(this.getAuthUrlBuilder(), FIREBASE_AUTH_UPLOAD_ACCOUNT, request)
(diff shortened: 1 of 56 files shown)


src/firestore/firestore.ts
```diff
@@ -71,8 +71,11 @@ export function getFirestoreOptions(app: FirebaseApp): Settings {
     });
   }
 
-  const projectId: string = utils.getProjectId(app);
-  const cert: Certificate = tryGetCertificate(app.options.credential);
+  const projectId: string | null = utils.getProjectId(app);
+  let cert: Certificate | null = null;
+  if (app.options.credential) {
+    cert = tryGetCertificate(app.options.credential);
```
hiranya911: Does it make sense to declare this method as:

```
tryGetCertificate(credential: Credential | null | undefined): Certificate | null;
```

src/messaging/messaging-errors.ts
```diff
@@ -67,21 +67,22 @@ export function createFirebaseError(err: HttpError): FirebaseMessagingError {
  */
 export function getErrorCode(response: any): string | null {
   if (validator.isNonNullObject(response) && 'error' in response) {
-    if (validator.isString(response.error)) {
-      return response.error;
+    const error = (response as any).error;
```
hiranya911: Is this because `isNoneNullObject()` asserts the type of `response` to be something other than `any`?

src/utils/api-request.ts
```diff
@@ -459,6 +470,11 @@ class AsyncHttpCall {
     try {
       this.config = new HttpRequestConfigImpl(config);
       this.options = this.config.buildRequestOptions();
+      if (!this.options.headers) {
```
hiranya911: This doesn't look right. It is possible for this to be undefined. May be initialize to empty object?

test/integration/auth.spec.ts
```diff
@@ -246,12 +248,13 @@ describe('admin.auth', () => {
       })
       .then(() => {
         // New sign-in should succeed.
-        return firebase.auth().signInWithEmailAndPassword(
+        return firebase.auth!().signInWithEmailAndPassword(
```
hiranya911: Why is this needed?

test/integration/auth.spec.ts
```diff
@@ -1384,11 +1409,16 @@ describe('admin.auth', () => {
         } as any,
         computePasswordHash: (userImportTest: UserImportTest): Buffer => {
           const currentRawPassword = userImportTest.rawPassword;
-          const currentRawSalt = userImportTest.rawSalt;
-          const N = userImportTest.importOptions.hash.memoryCost;
-          const r = userImportTest.importOptions.hash.blockSize;
-          const p = userImportTest.importOptions.hash.parallelization;
-          const dkLen = userImportTest.importOptions.hash.derivedKeyLength;
+          expect(userImportTest.rawSalt);
+          const currentRawSalt = userImportTest.rawSalt!;
```
hiranya911: Nit: Please add some whitespace after each `const` for clarity.

test/unit/auth/auth.spec.ts
```diff
@@ -1236,7 +1242,7 @@ AUTH_CONFIGS.forEach((testConfig) => {
       });
 
       it('should be rejected given invalid properties', () => {
-        return auth.createUser(null)
+        return auth.createUser(null as unknown as CreateRequest)
```
hiranya911: Can we `null as any`?

test/unit/auth/auth.spec.ts
```diff
@@ -411,6 +411,9 @@ AUTH_CONFIGS.forEach((testConfig) => {
       const tenantId = testConfig.supportsTenantManagement ? undefined : TENANT_ID;
       const expectedUserRecord = getValidUserRecord(getValidGetAccountInfoResponse(tenantId));
       // Set auth_time of token to expected user's tokensValidAfterTime.
+      if (!expectedUserRecord.tokensValidAfterTime) {
```
hiranya911: Can we use `!` assertion here and remove this `if` condition?

test/unit/auth/tenant-manager.spec.ts
```diff
@@ -401,7 +401,7 @@ describe('TenantManager', () => {
     });
 
     it('should be rejected given invalid TenantOptions', () => {
-      return tenantManager.createTenant(null)
+      return tenantManager.createTenant(null as unknown as TenantOptions)
```
hiranya911: `null as any` if that's allowed

test/unit/messaging/messaging.spec.ts
```diff
@@ -1116,7 +1116,7 @@ describe('Messaging', () => {
 
     it('should throw given no registration token(s) argument', () => {
       expect(() => {
-        messaging.sendToDevice(undefined as string, mocks.messaging.payloadDataOnly);
+        messaging.sendToDevice(undefined as unknown as string, mocks.messaging.payloadDataOnly);
```
hiranya911: `as any`

src/auth/auth-api-request.ts
```diff
@@ -683,7 +683,7 @@ const LIST_INBOUND_SAML_CONFIGS = new ApiSettings('/inboundSamlConfigs', 'GET')
  * Class that provides the mechanism to send requests to the Firebase Auth backend endpoints.
  */
 export abstract class AbstractAuthRequestHandler {
-  protected readonly projectId: string;
+  protected readonly projectId: string | null;
```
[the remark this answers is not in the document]
hiranya911: There are some `admin.auth()` APIs that doesn't require project ID (custom token creation for example). Option (b) with deferred validation sounds like the right thing to do. We do a similar thing in `FirebaseTokenVerifier`. But that may be a larger refactor than we wish to do right now. I'm ok with keeping this as is, and changing it later.

test/unit/auth/auth.spec.ts
```diff
@@ -2169,8 +2170,9 @@ AUTH_CONFIGS.forEach((testConfig) => {
 
         if (emailActionFlow.requiresSettings) {
           it('should reject when called without actionCodeSettings', () => {
-            return (auth as any)[emailActionFlow.api](email, undefined)
-              .should.eventually.be.rejected.and.have.property('code', 'auth/argument-error');
+            expect(() => {
+              (auth as any)[emailActionFlow.api](email, undefined);
+            }).to.throw(FirebaseAuthError).with.property('code', 'auth/argument-error');
```
hiranya911: This looks like a breaking change. Instead of rejecting the API now throws. We probably don't want that.

src/utils/api-request.ts
hiranya911: I see. The method `buildEntity()` also has code that modifies headers in-place.  I think we can just do this:

```
this.entity = this.config.buildEntity(this.options.headers!);
```

If `buildRequestOptions` did not set headers, this will trigger a runtime error in `buildEntity()` when it attempts to modify headers. That sounds like an ok position for us to be in. WDYT?

test/unit/auth/auth.spec.ts
hiranya911: I have a slight preference towards `as any` since it's less things to type. But this ok too.

test/unit/auth/auth.spec.ts
rsgowman: I wouldn't call it a breaking change; I'd call it a bug fix. :)  If you violate the api contract, you expect it to fail. (See also https://g3doc.corp.google.com/firebase/jscore/g3doc/contributing/styleguide.md#public-apis-should-sanitize-their-inputs-aggressively which I'm definitely in favour of.)

But you're right that this is a behaviour change, and I'm trying to minimize those in this PR. I've reverted this for now, but I think this is worth discussing:
a) whether we should change this throughout (though obviously not during this PR)
b) if so, whether we consider it a breaking change

test/unit/auth/auth.spec.ts
```diff
@@ -1243,7 +1243,7 @@ AUTH_CONFIGS.forEach((testConfig) => {
       });
 
       it('should be rejected given invalid properties', () => {
-        return auth.createUser(null as unknown as CreateRequest)
+        return auth.createUser(null as any as CreateRequest)
```
hiranya911: `null as any` should be sufficient here right?

test/unit/auth/tenant-manager.spec.ts
```diff
@@ -401,7 +401,7 @@ describe('TenantManager', () => {
     });
 
     it('should be rejected given invalid TenantOptions', () => {
-      return tenantManager.createTenant(null as unknown as TenantOptions)
+      return tenantManager.createTenant(null as any as TenantOptions)
```
hiranya911: Again `null as any` should be sufficient?

test/unit/messaging/messaging.spec.ts
```diff
@@ -398,10 +398,10 @@ describe('Messaging', () => {
   describe('send()', () => {
     it('should throw given no message', () => {
       expect(() => {
-        messaging.send(undefined as unknown as Message);
+        messaging.send(undefined as any as Message);
```
hiranya911: I'd expect just `undefined as any` to be enough here.

PR #681, opened 2019-10-21
Populates UserRecord with multi-factor related info

bojeil-google:
Updates UserRecord to parse multi-factor related information from the GetAccountInfo server response.

First step in supporting multi-factor authentication with SMS as a second factor.

### DIFF
--- a/src/auth/user-record.ts
+++ b/src/auth/user-record.ts
@@ -58,12 +58,7 @@ export interface CreateRequest extends UpdateRequest {
   uid?: string;
 }
 
-export interface Metadata {
-  createdAt?: string;
-  lastLoginAt?: string;
-}
-
-export interface MfaInfo {
+export interface AuthFactorInfo {
   mfaEnrollmentId: string;
   displayName?: string;
   phoneInfo?: string;
@@ -81,11 +76,7 @@ export interface ProviderUserInfo {
   federatedId?: string;
 }
 
-export interface Mfa {
-  mfaInfo?: MfaInfo[];
-}
-
-export interface GetAccountInfoUserResponse extends Metadata, Mfa {
+export interface GetAccountInfoUserResponse {
   localId: string;
   email?: string;
   emailVerified?: boolean;
@@ -99,6 +90,9 @@ export interface GetAccountInfoUserResponse extends Metadata, Mfa {
   validSince?: string;
   tenantId?: string;
   providerUserInfo?: ProviderUserInfo[];
+  mfaInfo?: AuthFactorInfo[];
+  createdAt?: string;
+  lastLoginAt?: string;
   [key: string]: any;
 }
 
@@ -123,7 +117,7 @@ export abstract class MultiFactorInfo {
    * @param response The server side response.
    * @constructor
    */
-  public static initMultiFactorInfo(response: MfaInfo): MultiFactorInfo | null {
+  public static initMultiFactorInfo(response: AuthFactorInfo): MultiFactorInfo | null {
     let multiFactorInfo: MultiFactorInfo | null = null;
     // Only PhoneMultiFactorInfo currently available.
     try {
@@ -140,9 +134,8 @@ export abstract class MultiFactorInfo {
    * @param response The server side response.
    * @constructor
    */
-  constructor(response: MfaInfo) {
-    const factorId = this.getFactorId(response);
-    this.initFromServerResponse(response, factorId);
+  constructor(response: AuthFactorInfo) {
+    this.initFromServerResponse(response);
   }
 
   /** @return The plain object representation. */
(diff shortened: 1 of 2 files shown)


src/auth/user-record.ts
```diff
@@ -57,19 +58,226 @@ export interface CreateRequest extends UpdateRequest {
...
+   *     not associated with any known multi-factor ID, null is returned.
+   */
+  protected abstract getFactorId(response: MfaInfo): MultiFactorId | null;
+
+  /**
+   * Initializes the MultiFactorInfo object using the provided server response.
+   *
+   * @param response The server side response.
+   */
+  private initFromServerResponse(response: MfaInfo, factorId: MultiFactorId | null) {
```
hiranya911: You can call `getFactorId()` in this method, and remove the `factorId` argument.

src/auth/user-record.ts
```diff
@@ -57,19 +58,226 @@ export interface CreateRequest extends UpdateRequest {
...
+   * If no MultiFactorInfo is associated with the response, null is returned.
+   *
+   * @param response The server side response.
+   * @constructor
+   */
+  public static initMultiFactorInfo(response: MfaInfo): MultiFactorInfo | null {
+    let multiFactorInfo: MultiFactorInfo | null = null;
+    // Only PhoneMultiFactorInfo currently available.
+    try {
+      multiFactorInfo = new PhoneMultiFactorInfo(response);
```
hiranya911: I guess the error occurs when an unexpected response is received. I think we should just expose it to the user, instead of silently ignoring it.

src/auth/user-record.ts
```diff
@@ -57,19 +58,226 @@ export interface CreateRequest extends UpdateRequest {
...
+
+/** Enums for multi-factor identifiers. */
+export enum MultiFactorId {
+  Phone = 'phone',
+}
+
+/**
+ * Abstract class representing a multi-factor info interface.
+ */
+export abstract class MultiFactorInfo {
```
hiranya911: Just `AuthFactor`? Subclass can be called `PhoneAuthFactor` or even just `PhoneFactor`.

src/auth/user-record.ts
```diff
@@ -57,19 +58,226 @@ export interface CreateRequest extends UpdateRequest {
...
+   *     not associated with any known multi-factor ID, null is returned.
+   */
+  protected getFactorId(response: MfaInfo): MultiFactorId | null {
+    return !!response.phoneInfo ? MultiFactorId.Phone : null;
+  }
+}
+
+/** Class representing multi-factor related properties of a user. */
+export class MultiFactor {
+  public readonly enrolledFactors: ReadonlyArray<MultiFactorInfo>;
```
hiranya911: This will be a lot clear if we rename `MultiFactorInfo` to `AuthFactor`.

"`MultiFactor` has a collection of `AuthFactor`s" has a nice ring to it.

test/unit/auth/user-record.spec.ts
```diff
@@ -204,11 +239,270 @@ function getUserInfoWithPhoneNumberJSON(): object {
   };
 }
 
+describe('PhoneMultiFactorInfo', () => {
+  const serverResponse: any = {
```
hiranya911: `MultiFactorAuthInfo`

test/unit/auth/user-record.spec.ts
```diff
@@ -204,11 +239,270 @@ function getUserInfoWithPhoneNumberJSON(): object {
...
+  const phoneMultiFactorInfoMissingFields = new PhoneMultiFactorInfo({
+    mfaEnrollmentId: serverResponse.mfaEnrollmentId,
+    phoneInfo: serverResponse.phoneInfo,
+  });
+
+  describe('constructor', () => {
+    it('should throw when an empty object is provided', () => {
+      expect(() =>  {
+        return new PhoneMultiFactorInfo({} as any);
+      }).to.throw(Error);
```
hiranya911: Can we test for a specific error message?

test/unit/auth/user-record.spec.ts
```diff
@@ -204,11 +239,270 @@ function getUserInfoWithPhoneNumberJSON(): object {
...
+        enrollmentTime: null,
+        phoneNumber: '+16505551234',
+        factorId: 'phone',
+      });
+    });
+  });
+});
+
+describe('MultiFactorInfo', () => {
+  const serverResponse: any = {
```
hiranya911: Can we use the specific type instead of any?

src/auth/user-record.ts
bojeil-google: This is the name agreed upon in the API proposal and already implemented in the client SDKs which is in alpha. It can't be changed anymore.

src/auth/user-record.ts
bojeil-google: We will add new types of second factors in the future. It is possible a developer may be using an older version of the Admin SDK. We shouldn't throw an error in that case.

src/auth/user-record.ts
```diff
@@ -57,19 +58,226 @@ export interface CreateRequest extends UpdateRequest {
   uid?: string;
 }
 
+export interface Metadata {
```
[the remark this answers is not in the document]
bojeil-google: I removed this interface.

src/auth/user-record.ts
```diff
@@ -57,19 +58,226 @@ export interface CreateRequest extends UpdateRequest {
...
+  createdAt?: string;
+  lastLoginAt?: string;
+}
+
+export interface MfaInfo {
```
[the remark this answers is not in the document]
bojeil-google: done

src/auth/user-record.ts
```diff
@@ -57,19 +58,226 @@ export interface CreateRequest extends UpdateRequest {
...
+  rawId: string;
+  displayName?: string;
+  email?: string;
+  photoUrl?: string;
+  phoneNumber?: string;
+  providerId: string;
+  federatedId?: string;
+}
+
+export interface Mfa {
```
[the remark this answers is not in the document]
bojeil-google: Done.

src/auth/user-record.ts
bojeil-google: Done.

src/auth/user-record.ts
```diff
@@ -57,19 +58,226 @@ export interface CreateRequest extends UpdateRequest {
...
+      // This can be parsed directly via Date constructor.
+      // This can be computed using Data.prototype.toISOString.
+      if (response.enrolledAt) {
+        utils.addReadonlyGetter(
+            this, 'enrollmentTime', new Date(response.enrolledAt).toUTCString());
+      } else {
+        utils.addReadonlyGetter(this, 'enrollmentTime', null);
+      }
+    } else {
+      throw new FirebaseAuthError(
```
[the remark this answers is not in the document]
bojeil-google: done

src/auth/user-record.ts
bojeil-google: Same here. This can't be changed at this point. It is the same name used in all the client SDKs. The Admin Node.js implementation should align with the JS client implementation.

test/unit/auth/user-record.spec.ts
bojeil-google: Done.

test/unit/auth/user-record.spec.ts
bojeil-google: Done.

test/unit/auth/user-record.spec.ts
bojeil-google: Done.

src/auth/user-record.ts
```diff
@@ -162,33 +155,33 @@ export abstract class MultiFactorInfo {
...
-      if (response.enrolledAt) {
-        utils.addReadonlyGetter(
-            this, 'enrollmentTime', new Date(response.enrolledAt).toUTCString());
-      } else {
-        utils.addReadonlyGetter(this, 'enrollmentTime', null);
-      }
-    } else {
+  private initFromServerResponse(response: AuthFactorInfo) {
+    const factorId = this.getFactorId(response);
+    if (!factorId || !response || !response.mfaEnrollmentId) {
```
hiranya911: Should we check `response` before passing it to `getFactorId()`?

bojeil-google: I have added a check for this. I also added a test when undefined is passed to the constructor that the expected error is caught and thrown.

PR #684, opened 2019-10-22
feat(fcm): Add 12 new Android Notification Parameters Support

lahirumaramba:
### Discussion

  * Introduce support for the new Android notification parameters in FCM API to AdminSDK.

### Testing

  * Added unit tests and integration tests to reflect this change.

### API Changes

  * In `Messaging` added new fields to `AndroidNotification` interface.
  * Introduced `LightSettings` interface

RELEASE NOTE: Added a series of new parameters to the `AndroidNotification` class that allow further customization of notifications that target Android devices.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -4101,17 +4101,17 @@ declare namespace admin.messaging {
    */
   interface LightSettings {
     /**
-     * Color of the LED in `#rrggbb` or `#rrggbbaa` format.
+     * Required. Set color of the LED in `#rrggbb` or `#rrggbbaa` format.
      */
     color: string;
 
     /**
-     * Along with `light_off_duration`, define the blink rate of LED flashes.
+     * Required. Along with `light_off_duration`, define the blink rate of LED flashes.
      */
     lightOnDurationMillis: number;
 
     /**
-     * Along with `light_on_duration`, define the blink rate of LED flashes. 
+     * Required. Along with `light_on_duration`, define the blink rate of LED flashes. 
      */
     lightOffDurationMillis: number;
   }
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -4017,7 +4017,7 @@ declare namespace admin.messaging {
 
     /**
      * For notifications that inform users about events with an absolute time reference, sets
-     * the time that the event in the notification occurred in milliseconds. Notifications
+     * the time that the event in the notification occurred. Notifications
      * in the panel are sorted by this time.
      */
     eventTimestamp?: Date;
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -4023,7 +4023,7 @@ declare namespace admin.messaging {
     eventTimestamp?: Date;
 
     /**
-     * Set whether or not this notification is relevant only to the current device. 
+     * Sets whether or not this notification is relevant only to the current device. 
      * Some notifications can be bridged to other devices for remote display, such as 
      * a Wear OS watch. This hint can be set to recommend this notification not be bridged. 
      * See [Wear OS guides](https://developer.android.com/training/wearables/notifications/bridger#existing-method-of-preventing-bridging)
@@ -4031,7 +4031,7 @@ declare namespace admin.messaging {
     localOnly?: boolean;
 
     /**
-     * Set the relative priority for this notification. Low-priority notifications 
+     * Sets the relative priority for this notification. Low-priority notifications 
      * may be hidden from the user in certain situations. Note this priority differs 
      * from `AndroidMessagePriority`. This priority is processed by the client after 
      * the message has been delivered. Whereas `AndroidMessagePriority` is an FCM concept 
@@ -4040,7 +4040,7 @@ declare namespace admin.messaging {
     priority?: ('min' | 'low' | 'default' | 'high' | 'max');
 
     /**
-     * Set the vibration pattern to use. Pass in an array of milliseconds to 
+     * Sets the vibration pattern to use. Pass in an array of milliseconds to 
      * turn the vibrator on or off. The first value indicates the duration to wait before 
      * turning the vibrator on. The next value indicates the duration to keep the 
      * vibrator on. Subsequent values alternate between duration to turn the vibrator 
@@ -4078,8 +4078,8 @@ declare namespace admin.messaging {
     defaultLightSettings?: boolean;
 
     /**
-     * Set the visibility of the notification. Must be either `private`, `public`, 
-     * or `secret`. If unspecified, default to `private`.
+     * Sets the visibility of the notification. Must be either `private`, `public`, 
+     * or `secret`. If unspecified, defaults to `private`.
      */
     visibility?: ('private' | 'public' | 'secret');
 
@@ -4101,17 +4101,17 @@ declare namespace admin.messaging {
    */
   interface LightSettings {
     /**
-     * Required. Set color of the LED in `#rrggbb` or `#rrggbbaa` format.
+     * Required. Sets color of the LED in `#rrggbb` or `#rrggbbaa` format.
      */
     color: string;
 
     /**
-     * Required. Along with `light_off_duration`, define the blink rate of LED flashes.
+     * Required. Along with `light_off_duration`, defines the blink rate of LED flashes.
      */
     lightOnDurationMillis: number;
 
     /**
-     * Required. Along with `light_on_duration`, define the blink rate of LED flashes. 
+     * Required. Along with `light_on_duration`, defines the blink rate of LED flashes. 
      */
     lightOffDurationMillis: number;
   }
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -4000,6 +4000,146 @@ declare namespace admin.messaging {
      * ID specified in the app manifest.
      */
     channelId?: string;
+
+    /**
+     * Sets the "ticker" text, which is sent to accessibility services. Prior to 
+     * API level 21 (Lollipop), sets the text that is displayed in the status bar 
+     * when the notification first arrives.
+     */
+    ticker?: string;
+
+    /**
+     * When set to `false` or unset, the notification is automatically dismissed when 
+     * the user clicks it in the panel. When set to `true`, the notification persists 
+     * even when the user clicks it.
+     */
+    sticky?: boolean;
+
+    /**
+     * Set the time that the event in the notification occurred. Notifications in the 
+     * panel are sorted by this time.
+     */
+    eventTimestamp?: Date;
+
+    /**
+     * Set whether or not this notification is relevant only to the current device. 
+     * Some notifications can be bridged to other devices for remote display, such as 
+     * a Wear OS watch. This hint can be set to recommend this notification not be bridged. 
+     * See [Wear OS guides](https://developer.android.com/training/wearables/notifications/bridger#existing-method-of-preventing-bridging)
+     */
+    localOnly?: boolean;
+
+    /**
+     * Set the relative priority for this notification. Low-priority notifications 
+     * may be hidden from the user in certain situations. Note this priority differs 
+     * from `AndroidMessagePriority`. This priority is processed by the client after 
+     * the message has been delivered. Whereas `AndroidMessagePriority` is an FCM concept 
+     * that controls when the message is delivered.
+     */
+    priority?: ('min' | 'low' | 'default' | 'high' | 'max');
+
+    /**
+     * Set the vibration pattern to use. Pass in an array of milliseconds to turn 
+     * on or off the vibrator. The first value indicates the duration to wait before 
+     * turning the vibrator on. The next value indicates the duration to keep the 
+     * vibrator on. Subsequent values alternate between duration to turn the vibrator 
+     * off and to turn the vibrator on. If `vibrate_timings` is set and `default_vibrate_timings` 
+     * is set to `true`, the default value is used instead of the user-specified `vibrate_timings`.
+     */
+    vibrateTimingsMillis?: number[];
+
+    /**
+     * If set to `true`, use the Android framework's default vibrate pattern for the 
+     * notification. Default values are specified in [`config.xml`](https://android.googlesource.com/platform/frameworks/base/+/master/core/res/res/values/config.xml). 
+     * If `default_vibrate_timings` is set to `true` and `vibrate_timings` is also set, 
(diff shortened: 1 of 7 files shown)


src/utils/validator.ts
```diff
@@ -247,3 +247,19 @@ export function isTopic(topic: any): boolean {
...
+/**
+ * Validates that a value is a number and falls within the given range.
+ *
+ * @param {any} value The value to validate.
+ * @param {min} lower bound of the range to validate.
+ * @param {max} upper bound of the range to validate.
+ * @param {inclusive} indicates whether to include or exclude lower and upper bounds in the range.
+ * @return {boolean} Whether the value is a number or not and weather the value is within the range.
+ */
+export function isNumberInRange(value: any, min: number, max: number, inclusive: boolean = false) {
```
hiranya911: Lets also move under `messaging/` for now.

test/unit/messaging/messaging.spec.ts
```diff
@@ -2394,6 +2394,117 @@ describe('Messaging', () => {
...
 
+    const invalidVibrateTimings = [[null, 500], [-100]];
+    invalidVibrateTimings.forEach((vibrateTimingsMillis) => {
+      it(`should throw given an invalid vibrateTimingsMillis: ${ vibrateTimingsMillis }`, () => {
+        const message: Message = {
+          condition: 'topic-name',
+          android: {
+            notification: {
+              vibrateTimingsMillis,
+              lightSettings: {
```
hiranya911: lightSettings are probably not needed in this test.

test/unit/messaging/messaging.spec.ts
```diff
@@ -2394,6 +2394,117 @@ describe('Messaging', () => {
...
+            },
+          },
+        };
+        expect(() => {
+          messaging.send(message);
+        }).to.throw('android.notification.lightSettings.color must be in the form #RRGGBB or a `Color` object');
+      });
+    });
+
+    it(`should throw given an invalid light on duration`, () => {
```
hiranya911: s/invalid/negative

test/unit/messaging/messaging.spec.ts
```diff
@@ -2394,6 +2394,117 @@ describe('Messaging', () => {
...
+          },
+        },
+      };
+      expect(() => {
+        messaging.send(message);
+      }).to.throw(
+        'android.notification.lightSettings.lightOnDurationMillis must be a non-negative duration in milliseconds');
+    });
+
+    it(`should throw given an invalid light off duration`, () => {
```
hiranya911: s/invalid/negative

src/utils/validator.ts
lahirumaramba: Since we will be removing the `Color` class we will no longer need this validation... I will remove this, too.

src/index.d.ts
```diff
@@ -4000,6 +4000,146 @@ declare namespace admin.messaging {
...
+
+    /**
+     * When set to `false` or unset, the notification is automatically dismissed when 
+     * the user clicks it in the panel. When set to `true`, the notification persists 
+     * even when the user clicks it.
+     */
+    sticky?: boolean;
+
+    /**
+     * Set the time that the event in the notification occurred. Notifications in the
```
egilmorez: I was a little confused about what kind of "event" this was, and suggested some extra verbiage.  Check out the parallel property in the Java PR and see what you think. :)

src/index.d.ts
```diff
@@ -4000,6 +4000,146 @@ declare namespace admin.messaging {
...
+     * may be hidden from the user in certain situations. Note this priority differs 
+     * from `AndroidMessagePriority`. This priority is processed by the client after 
+     * the message has been delivered. Whereas `AndroidMessagePriority` is an FCM concept 
+     * that controls when the message is delivered.
+     */
+    priority?: ('min' | 'low' | 'default' | 'high' | 'max');
+
+    /**
+     * Set the vibration pattern to use. Pass in an array of milliseconds to turn 
+     * on or off the vibrator. The first value indicates the duration to wait before
```
egilmorez: "turn the vibrator on or off" sounds more natural.

src/index.d.ts
```diff
@@ -4000,6 +4000,146 @@ declare namespace admin.messaging {
...
+     */
+    visibility?: ('private' | 'public' | 'secret');
+
+    /**
+     * Sets the number of items this notification represents. May be displayed as a 
+     * badge count for Launchers that support badging. See [`NotificationBadge`(https://developer.android.com/training/notify-user/badges). 
+     * For example, this might be useful if you're using just one notification to
+     * represent multiple new messages but you want the count here to represent
+     * the number of total new messages. If zero or unspecified, the systems
+     * that support badging uses it's default behavior, which is to increment
```
egilmorez: I think this should be plural, for "systems use," correct?

src/index.d.ts
```diff
@@ -4000,6 +4000,146 @@ declare namespace admin.messaging {
...
+    /**
+     * Along with `light_on_duration`, define the blink rate of LED flashes. 
+     */
+    lightOffDurationMillis: number;
+  }
+
+  /**
+   * Represents a color in the RGBA color space
+   */
+  interface Color {
```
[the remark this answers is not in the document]
lahirumaramba: Thanks!
I removed the Color type and added both `RGB` and `RGBA` support in hex format.

lahirumaramba:
Thank you, @hiranya911 and @egilmorez! Updated with changes.

chong-shao:
LGTM. Thanks @lahirumaramba !

src/index.d.ts
```diff
@@ -4088,7 +4089,7 @@ declare namespace admin.messaging {
      * For example, this might be useful if you're using just one notification to
      * represent multiple new messages but you want the count here to represent
      * the number of total new messages. If zero or unspecified, the systems
-     * that support badging uses it's default behavior, which is to increment
+     * that support badging use it's default behavior, which is to increment
```
egilmorez: Sorry, missed this: it's >> its.

Counterintuitive but true, it's = "it is," not possession.

lahirumaramba: Thanks, @egilmorez! This is an embarrassing one!
I updated the docs here to match with `https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#androidnotification`. Let me know what you think. Thanks!

src/index.d.ts
```diff
@@ -4016,8 +4016,9 @@ declare namespace admin.messaging {
     sticky?: boolean;
 
     /**
-     * Set the time that the event in the notification occurred. Notifications in the 
-     * panel are sorted by this time.
+     * For notifications that inform users about events with an absolute time reference, sets
+     * the time that the event in the notification occurred in milliseconds. Notifications
```
hiranya911: This is not really a milliseconds timestamp in this API. Might want to drop that part.

src/messaging/messaging-types.ts
```diff
@@ -718,36 +699,41 @@ function validateAndroidNotification(notification: AndroidNotification) {
...
 
   if (typeof notification.eventTimestamp !== 'undefined') {
-    let zuluTimestamp: string;
+    if (!(notification.eventTimestamp instanceof Date)) {
+      throw new FirebaseMessagingError(
+        MessagingClientErrorCode.INVALID_PAYLOAD, 'android.notification.eventTimestamp must be a valid `Date` object');
+    }
     // Convert timestamp to RFC3339 UTC "Zulu" format, example "2014-10-02T15:01:23.045123456Z"
-    zuluTimestamp = notification.eventTimestamp.toISOString();
+    const zuluTimestamp: string = notification.eventTimestamp.toISOString();
```
hiranya911: Nit: You don't have to add type annotations for single-line initializers like this. But it's fine to leave them as is too.

src/messaging/messaging-types.ts
```diff
@@ -789,44 +775,49 @@ function validateLightSettings(lightSettings: LightSettings) {
...
-
-  if (typeof lightSettings.lightOffDurationMillis !== 'undefined') {
-    if (!validator.isNumber(lightSettings.lightOffDurationMillis) || lightSettings.lightOffDurationMillis < 0) {
-      throw new FirebaseMessagingError(
-        MessagingClientErrorCode.INVALID_PAYLOAD,
-        'android.notification.lightSettings.lightOffDurationMillis must be a non-negative duration in milliseconds');
-    }
-    const duration: string = transformMillisecondsToSecondsString(lightSettings.lightOffDurationMillis);
-    (lightSettings as any).lightOffDurationMillis = duration;
+  if (!validator.isNumber(lightSettings.lightOnDurationMillis) || lightSettings.lightOnDurationMillis < 0) {
```
hiranya911: This also covers the cases where `lightOnDurationMillis` is null or undefined. Therefore you can remove the previous check.

src/messaging/messaging-types.ts
```diff
@@ -789,44 +775,49 @@ function validateLightSettings(lightSettings: LightSettings) {
...
-    }
-    const rgb = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(lightSettings.color);
-    const color: Color = {
-      red: parseInt(rgb[1], 16) / 255.0,
-      green: parseInt(rgb[2], 16) / 255.0,
-      blue: parseInt(rgb[3], 16) / 255.0,
-      alpha: 1.0,
-    };
-    (lightSettings as any).color = color;
+  if (typeof lightSettings.lightOffDurationMillis === 'undefined' || lightSettings.lightOffDurationMillis === null) {
```
hiranya911: Can remove. Already covered by the following check.

src/messaging/messaging-types.ts
```diff
@@ -789,44 +775,49 @@ function validateLightSettings(lightSettings: LightSettings) {
...
+  if (!validator.isNumber(lightSettings.lightOffDurationMillis) || lightSettings.lightOffDurationMillis < 0) {
+    throw new FirebaseMessagingError(
+      MessagingClientErrorCode.INVALID_PAYLOAD,
+      'android.notification.lightSettings.lightOffDurationMillis must be a non-negative duration in milliseconds');
+  }
+  const durationOff: string = transformMillisecondsToSecondsString(lightSettings.lightOffDurationMillis);
+  (lightSettings as any).lightOffDurationMillis = durationOff;
 
-  validateColor(lightSettings.color as Color);
+  if (typeof lightSettings.color === 'undefined' || lightSettings.color === null) {
```
hiranya911: Lets remove this.

src/messaging/messaging-types.ts
```diff
@@ -789,44 +775,49 @@ function validateLightSettings(lightSettings: LightSettings) {
...
+    throw new FirebaseMessagingError(
+      MessagingClientErrorCode.INVALID_PAYLOAD, 'android.notification.lightSettings.color is required');
+  }
+  if (!/^#[0-9a-fA-F]{6}$/.test(lightSettings.color) && !/^#[0-9a-fA-F]{8}$/.test(lightSettings.color)) {
```
hiranya911: Lets add `!validator.isString(lightSettings.color) || ....` which should cover undefined and null cases.

src/messaging/messaging-types.ts
```diff
@@ -880,3 +844,28 @@ function validateAndroidFcmOptions(fcmOptions: AndroidFcmOptions) {
...
+/**
+ * Transforms milliseconds to the format expected by FCM service.
+ * Returns the duration in seconds with up to nine fractional
+ * digits, terminated by 's'. Example: "3.5s".
+ *
+ * @param {number} milliseconds The duration in milliseconds.
+ * @return {string} The resulting formatted string in seconds with up to nine fractional
+ * digits, terminated by 's'.
+ */
+export function transformMillisecondsToSecondsString(milliseconds: number): string {
```
hiranya911: Doesn't need to be exported (I think). In general be conservative about what's exported from each module.

src/index.d.ts
```diff
@@ -4000,6 +4000,120 @@ declare namespace admin.messaging {
...
+   */
+  interface LightSettings {
+    /**
+     * Required. Set color of the LED in `#rrggbb` or `#rrggbbaa` format.
+     */
+    color: string;
+
+    /**
+     * Required. Along with `light_off_duration`, define the blink rate of LED flashes.
+     */
```
lahirumaramba: @egilmorez : Updated the docs to add `Required` here. Also, should the verbs on docs be singular or plural?
Ex: `Set` vs `Sets` or `define` vs `defines`

src/messaging/messaging-types.ts
lahirumaramba: Thanks!

src/messaging/messaging-types.ts
lahirumaramba: Good catch! This was copypasta when the function was moved from utils. 
Thanks! Fixed.

src/messaging/messaging-types.ts
```diff
@@ -699,10 +747,73 @@ function validateAndroidNotification(notification: AndroidNotification) {
...
+  const durationOff = transformMillisecondsToSecondsString(lightSettings.lightOffDurationMillis);
+  (lightSettings as any).lightOffDurationMillis = durationOff;
+
+  if (!validator.isString(lightSettings.color) ||
+    (!/^#[0-9a-fA-F]{6}$/.test(lightSettings.color) && !/^#[0-9a-fA-F]{8}$/.test(lightSettings.color))) {
+    throw new FirebaseMessagingError(
+      MessagingClientErrorCode.INVALID_PAYLOAD,
+      'android.notification.lightSettings.color must be in the form #RRGGBB or #RRGGBBAA format');
+  }
+  lightSettings.color += lightSettings.color.length === 7 ? 'FF' : '';
```
lahirumaramba: @hiranya911:
Do you think it would be better to generate the color string and assigned to a temp variable without updating the `LightSettings` object directly here?
```
const colorString = lightSettings.color.length === 7 ? lightSettings.color + 'FF' : lightSettings.color;
// populate color object
(lightSettings as any).color = colorString;
```

hiranya911: Yeah, that looks cleaner.

src/index.d.ts
egilmorez: I think the verbs should all be singular unless there's a really obvious case for plural.  Since most comments describe a single property or method rather than groups of them, we can consider singular the "default" . :)

PR #706, opened 2019-11-18
Adds multi-factor support for updateUser

bojeil-google:
Adds multi-factor support for updateUser
This will allow the ability to update the list of second factors on an existing user record.

Adds related unit and integration tests for this operation.

### DIFF
--- a/src/auth/auth-api-request.ts
+++ b/src/auth/auth-api-request.ts
@@ -25,7 +25,7 @@ import {
 import {CreateRequest, UpdateRequest} from './user-record';
 import {
   UserImportBuilder, UserImportOptions, UserImportRecord,
-  UserImportResult, AuthFactorInfo,
+  UserImportResult, AuthFactorInfo, convertMultiFactorInfoToServerFormat,
 } from './user-import-builder';
 import * as utils from '../utils/index';
 import {ActionCodeSettings, ActionCodeSettingsBuilder} from './action-code-settings-builder';
@@ -304,6 +304,8 @@ function validateCreateEditRequest(request: any, uploadAccountRequest: boolean =
     lastLoginAt: uploadAccountRequest,
     providerUserInfo: uploadAccountRequest,
     mfaInfo: uploadAccountRequest,
+    // Only for non-uploadAccount requests.
+    mfa: !uploadAccountRequest,
   };
   // Remove invalid keys from original request.
   for (const key in request) {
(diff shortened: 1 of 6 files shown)


test/integration/auth.spec.ts
```diff
@@ -314,20 +314,66 @@ describe('admin.auth', () => {
   });
 
   it('updateUser() updates the user record with the given parameters', () => {
+    // TODO: replace all "any" types with explicit mfa related types after they are defined in d.ts file.
```
hiranya911: Any reason for not updating the d.ts file in this PR? We can add the documentation comments in a future PR, but add the typings here.

bojeil-google: Yeah I delayed that because of the documentation comments. I have added the types. Will defer the comments for a later PR.

src/auth/user-record.ts
```diff
@@ -51,6 +51,15 @@ export interface UpdateRequest {
   password?: string;
   phoneNumber?: string | null;
   photoURL?: string | null;
+  multiFactor?: {
```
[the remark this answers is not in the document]
bojeil-google: done

src/auth/auth-api-request.ts
```diff
@@ -1049,6 +1059,28 @@ export abstract class AbstractAuthRequestHandler {
       request.disableUser = request.disabled;
       delete request.disabled;
     }
+    // Construct mfa related user data.
+    if (validator.isNonNullObject(request.multiFactor)) {
+      if (request.multiFactor.enrolledFactors === null) {
```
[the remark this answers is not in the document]
bojeil-google: Hmm it is a bit more complicated as passing `request.mfa = {}` will clear all existing second factors on the user.
We should only pass that if the developers passing explicitly an empty array for `enrolledFactors` or null for `enrolledFactors`.

src/auth/user-record.ts
```diff
@@ -51,6 +51,15 @@ export interface UpdateRequest {
...
+    enrolledFactors: Array<{
```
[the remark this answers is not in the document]
bojeil-google: I created a local `SecondFactor` type. Didn't want to risk getting in a possible circular dependency situation.

src/index.d.ts
```diff
@@ -1000,6 +1020,16 @@ declare namespace admin.auth {
      * to the tenant corresponding to that `TenantAwareAuth` instance's tenant ID.
      */
     tenantId?: string | null;
+
+    multiFactor?: {
+      enrolledFactors: Array<{
```
hiranya911: Does it make sense to reduce this duplication by introducing an interface? Like the `SecondFactor` type used in the implementation?

bojeil-google: Yeah I think we can define an interface for this. I can do it in the next PR. This is going to be modified and more documentation added later anyway.

PR #708, opened 2019-11-28
Allow createCustomToken() to work with tenant-aware auth

rsgowman:


src/auth/token-generator.ts
```diff
@@ -254,13 +255,38 @@ export class FirebaseTokenGenerator {
...
+      uid: string, tenantId: string, developerClaims?: {[key: string]: any}): Promise<string> {
+    if (!validator.isNonEmptyString(tenantId)) {
+      throw new FirebaseAuthError(
+        AuthClientErrorCode.INVALID_ARGUMENT,
+        '`tenantId` argument must be a non-empty string uid.');
+    }
+    return this.createCustomTokenInternal(uid, tenantId, developerClaims);
+  }
+
+  private createCustomTokenInternal(
```
rsgowman: Alternatively to having two separate createCustomToken*() methods, we could have one with an options object, i.e.:
```javascript
interface CreateCustomTokenOptions {
  tenantId?: string;
  developerClaims?: {[key: string]: any};
}
public createCustomToken(uid: string, options?: CreateCustomTokenOptions) { ... }
```

test/unit/auth/token-generator.spec.ts
```diff
@@ -293,22 +294,22 @@ describe('FirebaseTokenGenerator', () => {
     it('should throw given no uid', () => {
       expect(() => {
         (tokenGenerator as any).createCustomToken();
-      }).to.throw('First argument to createCustomToken() must be a non-empty string uid');
+      }).to.throw(FirebaseAuthError).with.property('code', 'auth/argument-error');
```
rsgowman: I've switched many of the error check tests from 'verify the error message' to 'verify the error code'. Personally, I prefer just verifying the code, as the error message itself often changes which makes the tests a little fragile. But this is hardly a universal opinion. I'm happy to add the error message back in if you'd prefer.

(Obviously, these changes in particular were caused by me tweaking the error message slightly and having a bunch of tests fail as a result.)

test/unit/auth/token-generator.spec.ts
```diff
@@ -466,4 +467,215 @@ describe('FirebaseTokenGenerator', () => {
         });
     });
   });
+
+  describe('createCustomTokenWithTenantId()', () => {
```
rsgowman: These tests largely duplicate the createCustomToken() tests above, but validate that the path with tenantIds also works. Roughly half of these tests are pretty low (though not quite no) value and I'd be happy to remove them. i.e. the tests that validate the uid handling is correct, even when a tennantid is provided, e.g. 'should throw given an empty string uid'.

Opinions welcome.

test/integration/auth.spec.ts
```diff
@@ -710,6 +710,49 @@ describe('admin.auth', () => {
...
+            return tenantAwareAuth.verifyIdToken(idToken);
+          })
+          .then((token) => {
+            expect(token.uid).to.equal('uid1');
+            expect(token.firebase.tenant).to.equal(createdTenantId);
+          });
+      });
+
+      it('createCustomToken() should fail if tenantIds mismatch', async () => {
+        const aDifferentTenant = await admin.auth().tenantManager().createTenant({displayName: 'A-Different-Tenant'});
```
wuyanna: Tenants might not get cleaned up if the test is interrupted. If you just want to test a failure case, you can just set a random tenantId in the custom token or in the sign in request without creating new tenant, and it should fail with a similar message

rsgowman: I wanted to ensure that the mismatched tenant id was otherwise valid. For instance; you could imagine a backend implementation that inspected the tenant id to ensure it was valid, and *only then* checked the token to ensure it matched. By using an otherwise valid tenant id, I think this test becomes *slightly* stronger. (As a minor additional benefit, it also ensures I use a properly formatted tenant id; even in the unlikely event that the definition of a properly formatted tenant id changes in the future.)

It is true that if the test is interrupted, the tenant might not be cleaned up. But the existing code also suffers from this, and I'm not sure we can really make that bullet-proof. (The existing code creates a few tenants/users as well and then deletes them after the tests are completed.) We could just delete *all* users/tenants upon starting the tests... but this would prevent us from running two instances of the tests at once (though that probably doesn't work anyways) and would be dangerous if users accidentally supplied a "real" project to the test suite (well; more dangerous.)

Incidentally, as I was looking at this comment, I noticed that I was missing an await() in the cleanup code; fixed.

wuyanna: But it seems the thing you are checking here is backend implementation. Does Admin sdk integration test need to cover that? Also it seems a bit weird to me that only this test needs to have await()? I will leave to @bojeil-google to make the judgement here.

rsgowman: tl;dr: You're right; this test isn't necessary. I've removed it.



(The async() allows this to work without having to chain promises; which is especially useful when using try/finally. But you can get by just fine without it by either registering created tenants/etc and deleting them in the after() method, or by `.catch()`ing at the end of a promise chain. I think async() makes this more readable, but am happy to convert to either of the other approaches.)

This test really is testing the error handling on the sdk side; it behaves differently in the presence of tenant id mismatch vs invalid tenant id. (Details below.) Although, signInWithCustomToken is actually client sdk rather than admin sdk, so  this test should be moved there; you're exactly right that the admin sdk itself can't possibly care whether the tenant id is invalid or mismatched. I've removed this test.

Details: If I use an invalid tenant id (rather than a mismatching one) and then attempt to use the resulting token to login via the client sdk, I get this error: `The Auth instance's tenant ID is invalid.` with a code of `invalid-tenant-id`. But if I use a valid, but mismatching tenant id, I get `Specified tenant ID does not match the custom token.` with a code of `internal-error`. (Which is a bug in the client sdk; the code should be 'tenant-id-mismatch'; fixed here: https://github.com/firebase/firebase-js-sdk/pull/2389. In fact, that PR effectively performs this exact test.)

wuyanna: Thanks for the quick fix on the client SDK bug!

src/auth/token-generator.ts
```diff
@@ -254,13 +255,38 @@ export class FirebaseTokenGenerator {
    *                           service account key and containing the provided payload.
    */
   public createCustomToken(uid: string, developerClaims?: {[key: string]: any}): Promise<string> {
+    return this.createCustomTokenInternal(uid, null, developerClaims);
+  }
+
+  /**
+   * Creates a new Firebase Auth Custom token.
+   *
+   * @param {string} uid The user ID to use for the generated Firebase Auth Custom token.
```
bojeil-google: I know this is the current pattern but we should move away from re-stating the types, eg. `string` in the javadocs.
This is according to the internal typescript readability guideline.

test/unit/auth/auth.spec.ts
```diff
@@ -328,63 +328,49 @@ AUTH_CONFIGS.forEach((testConfig) => {
...
+        const mockCredentialAuth = testConfig.init(mocks.mockCredentialApp());
 
-        it('should be fulfilled given an app which returns null access tokens', () => {
-          // createCustomToken() does not rely on an access token and therefore works in this scenario.
-          return nullAccessTokenAuth.createCustomToken(mocks.uid, mocks.developerClaims)
-            .should.eventually.be.fulfilled;
-        });
+        expect(() => {
+          mockCredentialAuth.createCustomToken(mocks.uid, mocks.developerClaims);
+        }).not.to.throw;
```
bojeil-google: Seems to be a mismatch with the test description (should throw if...)

test/unit/auth/token-generator.spec.ts
bojeil-google: You can follow a similar approach to other tests for common API used in a tenant and non-tenant context where shared code is used. This avoids the duplication and makes it easier to make changes in the future.

rsgowman: You mean like this? https://github.com/firebase/firebase-admin-node/blob/bba61e16f2bda906833c35d25e2fbc9a4c3613cc/test/unit/auth/auth.spec.ts#L244

The problem here is that:
a) We have a TokenGenerator (rather than an Auth object) and the TokenGenerator is not tenant aware.
b) The API for creating a custom token differs based on whether you want a tenantid included or not.

This problem is the more general problem of testing method overloads; I'm not aware of a great solution. Here's the obvious:
a) Just duplicate the tests. (i.e. as currently implemented)
b) Mock the implementation method and just verify that createCustomToken() pipes it's arguments through successfully. Possibly more pragmatic, but requires you to peek at the implementation (and to assume it won't change.)

In our case, we could also:
c) Make TokenGenerator tenant aware. Then we could use the same approach as for Auth instances. It seems a little odd... but I think it's workable with some refactoring around how the Auth instance creates its TokenGenerator.
d) Iterate over a list of functions, like this (which is pretty ugly):

```typescript
const createTokenFns = [
  (uid: string, tenantId: string, developerClaims?: {[key: string]: any}): Promise<string> => {
    return tokenGenerator.createCustomToken(string, developerClaims);
  }
  (uid: string, tenantId: string, developerClaims?: {[key: string]: any}): Promise<string> => {
    return tokenGenerator.createCustomTokenWithTenantId(string, tenantId, developerClaims);
  }];

createTokensFns.forEach((fn) => {
  describe("create custom token, possibly tenant aware", () => {
    it('should work', () => {
      expect(fn('uid', 'tenantid_that_might_be_ignored')).should.eventually.be.fulfilled;
    });
  });
});
```

I don't like (d); I *mildly* don't like (b). (c) doesn't seem so bad though. wdyt?

test/unit/auth/auth.spec.ts
rsgowman: But wait! That's wrong too, since this test can't call "the IAM service in the cloud when the service account credentials are not provided." So instead of throwing immediately, it now returns a failed future. Fixed (for real this time) by ensuring it *does* fail, but not immediately. (Might be worth checking with @hiranya911 to see if this captures the original intent.)

test/unit/auth/token-generator.spec.ts
bojeil-google: `TokenGenerator` is created per `Auth` and `TenantAwareAuth` instance. The tenant ID in that case does not change after creation (could be set in the constructor). `TokenGenerator` has only one public method `createCustomToken` that mints tokens that are tenant aware. Every call of that should also set the tenant ID in the minted JWT. So maybe (c) is a good choice.  But I am not sure how much work is involved in this. Do you want to explore and make an assessment?

rsgowman: It's not too much at all. Done.

src/auth/auth.ts
```diff
@@ -102,9 +102,14 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> {
    *     for this instance.
    * @constructor
    */
-  constructor(app: FirebaseApp, protected readonly authRequestHandler: T) {
-    const cryptoSigner = cryptoSignerFromApp(app);
-    this.tokenGenerator = new FirebaseTokenGenerator(cryptoSigner);
+  constructor(app: FirebaseApp, protected readonly authRequestHandler: T, tokenGenerator?: FirebaseTokenGenerator) {
```
bojeil-google: We should update the javadocs with the right number of parameters

src/auth/token-generator.ts
```diff
@@ -247,33 +248,43 @@ export class FirebaseTokenGenerator {
...
+   *     resulting JWT.
+   */
+  constructor(signer: CryptoSigner, public readonly tenantId?: string) {
     if (!validator.isNonNullObject(signer)) {
       throw new FirebaseAuthError(
         AuthClientErrorCode.INVALID_CREDENTIAL,
         'INTERNAL ASSERT: Must provide a CryptoSigner to use FirebaseTokenGenerator.',
       );
     }
+    if (tenantId !== undefined && !validator.isNonEmptyString(tenantId)) {
```
bojeil-google: nit: `typeof tenantId !== 'undefined'`

src/auth/token-generator.ts
```diff
@@ -247,33 +248,43 @@ export class FirebaseTokenGenerator {
...
+      throw new FirebaseAuthError(
+        AuthClientErrorCode.INVALID_ARGUMENT,
+        '`tenantId` argument must be a non-empty string uid.');
```
bojeil-google: Maybe remove `uid` mention as it may confuse the developer since we use it to refer to the user unique identifier?
`tenantId` argument must be a non-empty string.

test/unit/auth/token-generator.spec.ts
```diff
@@ -285,183 +282,221 @@ describe('FirebaseTokenGenerator', () => {
...
+  });
 
-    it('should throw given a uid with a length greater than 128 characters', () => {
-      // uid of length 128 should be allowed
-      let uid = Array(129).join('a');
-      expect(uid).to.have.length(128);
-      expect(() => {
-        tokenGenerator.createCustomToken(uid);
-      }).not.to.throw();
+  const tokenGeneratorConfigs = [{
```
bojeil-google: You can also declare an interface for this.

test/integration/auth.spec.ts
```diff
@@ -722,6 +722,23 @@ describe('admin.auth', () => {
             expect(userRecord.uid).to.equal(createdUserUid);
           });
       });
+
+      it('createCustomToken() mints a JWT that can be used to sign in tenant users', async () => {
+        try {
+          firebase.auth!().tenantId = createdTenantId;
```
bojeil-google: Style guide does not seem to recommend non-nullability assertions without a reason.
https://g3doc.corp.google.com/javascript/typescript/g3doc/styleguide.md?cl=head#type-and-non-nullability-assertions

test/integration/auth.spec.ts
```diff
@@ -91,6 +91,10 @@ function randomOidcProviderId(): string {
   return 'oidc.' + generateRandomString(10, false).toLowerCase();
 }
 
+function client_auth(): FirebaseAuth {
```
bojeil-google: Please use camel case `clientAuth` for function names here and elsewhere.

PR #721, opened 2019-12-10
Changes saltSeparator type from string to Buffer

ggirotto:
Fix an inconsistency where `hash.saltSeparator` on `UserImportOptions` is expecting a `Buffer` but its type on Typescript module was specified as a `string`


ggirotto:
@googlebot I signed it!

ggirotto:
Hi @bojeil-google. There's something else that I have to do or just wait for the merge?

PR #726, opened 2019-12-12
feat(auth): Add bulk get/delete methods

rsgowman:
This PR allows callers to retrieve a list of users by unique identifier (uid, email, phone, federated provider uid) as well as to delete a list of users.

RELEASE NOTE: Added `getUsers()` and `deleteUsers()` APIs for retrieving and deleting user accounts in bulk.


src/utils/error.ts
```diff
@@ -649,6 +649,10 @@ export class AuthClientErrorCode {
     code: 'not-found',
     message: 'The requested resource was not found.',
   };
+  public static USER_NOT_DISABLED = {
+    code: 'user-not-disabled',
+    message: 'The user must be disabled in order to bulk delete it (or you must pass force=true).',
```
hiranya911: Didn't we decide against the `force=true` option in this API?

src/auth/auth.ts
```diff
@@ -263,6 +339,46 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> {
...
 
+  public deleteUsers(uids: string[]): Promise<DeleteUsersResult> {
+    return this.authRequestHandler.deleteAccounts(uids, /*force=*/true)
+      .then((batchDeleteAccountsResponse) => {
+        const result: DeleteUsersResult = {
+          failureCount: 0,
+          successCount: uids.length,
+          errors: [],
+        };
+
```
hiranya911: Nit: Remove the large if block with:

```
if (!validator.isNonEmptyArray(batchDeleteAccountsResponse.errors)) {
  return result;
}
```

src/auth/auth-api-request.ts
```diff
@@ -463,6 +490,45 @@ export const FIREBASE_AUTH_DELETE_ACCOUNT = new ApiSettings('/accounts:delete',
     }
   });
 
+interface BatchDeleteAccountsRequest {
+  localIds?: string[];
+  force?: boolean;
+}
```
hiranya911: Separate with whitespace.

src/auth/auth-api-request.ts
```diff
@@ -463,6 +490,45 @@ export const FIREBASE_AUTH_DELETE_ACCOUNT = new ApiSettings('/accounts:delete',
...
+interface BatchDeleteErrorInfo {
+  index?: number;
+  localId?: string;
+  message?: string;
+}
```
hiranya911: Here too

src/auth/auth-api-request.ts
```diff
@@ -452,6 +464,21 @@ export const FIREBASE_AUTH_GET_ACCOUNT_INFO = new ApiSettings('/accounts:lookup'
...
+ * for multiple accounts.
+ */
+export const FIREBASE_AUTH_GET_ACCOUNTS_INFO = new ApiSettings('/accounts:lookup', 'POST')
+  // Set request validator.
+  .setRequestValidator((request: GetAccountInfoRequest) => {
+    if (!request.localId && !request.email && !request.phoneNumber && !request.federatedUserId) {
+      throw new FirebaseAuthError(
+        AuthClientErrorCode.INTERNAL_ERROR,
+        'INTERNAL ASSERT FAILED: Server request is missing user identifier');
+    }
```
hiranya911: Should we also add a response validator?

src/auth/auth-api-request.ts
```diff
@@ -463,6 +490,45 @@ export const FIREBASE_AUTH_DELETE_ACCOUNT = new ApiSettings('/accounts:delete',
...
+export const FIREBASE_AUTH_BATCH_DELETE_ACCOUNTS = new ApiSettings('/accounts:batchDelete', 'POST')
+  .setRequestValidator((request: BatchDeleteAccountsRequest) => {
+    if (!request.localIds) {
+      throw new FirebaseAuthError(
+        AuthClientErrorCode.INTERNAL_ERROR,
+        'INTERNAL ASSERT FAILED: Server request is missing user identifiers');
+    }
+  })
+  .setResponseValidator((response: BatchDeleteAccountsResponse) => {
+    if (response.errors) {
```
hiranya911: Eliminate the if condition with:

`const errors = response.errors || [];`

src/auth/auth-api-request.ts
```diff
@@ -785,6 +851,62 @@ export abstract class AbstractAuthRequestHandler {
...
+  /**
+   * Looks up multiple users by their identifiers (uid, email, etc).
+   *
+   * @param {UserIdentifier[]} identifiers The identifiers indicating the users
+   *     to be looked up. Must have <= 100 entries.
+   * @param {Promise<object>} A promise that resolves with the set of successfully
+   *     looked up users. Possibly empty if no users were looked up.
+   */
+  public getAccountInfoByIdentifiers(identifiers: UserIdentifier[]): Promise<object> {
+    if (identifiers.length === 0) {
```
hiranya911: Should probably also do `!validator.isArray(identifiers)` and throw.

src/auth/auth-api-request.ts
```diff
@@ -785,6 +851,62 @@ export abstract class AbstractAuthRequestHandler {
...
+    } else if (identifiers.length > 100) {
+      throw new FirebaseAuthError(
+        AuthClientErrorCode.MAXIMUM_USER_COUNT_EXCEEDED,
+        '`identifiers` parameter must have <= 100 entries.');
+    }
+
+    const request: GetAccountInfoRequest = {};
+
+    for (const id of identifiers) {
+      if (isUidIdentifier(id)) {
```
hiranya911: I feel like a polymorphic design would have cleaned this clutter up a lot, but such APIs are not very user friendly in JS. How about we at least push some of this logic into small helper functions?

```
if (isUserIdentifier(id)) {
  handleUserIdentifier(id, request);
} else if (isEmailIdentifier(id)) {
  handleEmailIdentifier(id, request);
}
// and so on...
```

src/auth/auth-api-request.ts
```diff
@@ -882,6 +1004,30 @@ export abstract class AbstractAuthRequestHandler {
     return this.invokeRequestHandler(this.getAuthUrlBuilder(), FIREBASE_AUTH_DELETE_ACCOUNT, request);
   }
 
+  public deleteAccounts(uids: string[], force: boolean): Promise<BatchDeleteAccountsResponse> {
+    if (uids.length === 0) {
```
hiranya911: Check array.

Is the force option needed?

test/unit/auth/auth.spec.ts
```diff
@@ -1130,6 +1131,84 @@ AUTH_CONFIGS.forEach((testConfig) => {
...
+          identifiers.push({uid: 'id' + i});
+        }
+
+        expect(() => auth.getUsers(identifiers))
+          .to.throw(FirebaseAuthError)
+          .with.property('code', 'auth/maximum-user-count-exceeded');
+      });
+
+      it('should return no results when given no identifiers', () => {
+        return auth.getUsers([])
```
hiranya911: Test for null, undefined and other non-array things.

test/unit/auth/auth.spec.ts
```diff
@@ -1130,6 +1131,84 @@ AUTH_CONFIGS.forEach((testConfig) => {
...
+        stubs.push(stub);
+        const notFoundIds = [{uid: 'id that doesnt exist'}];
+        return auth.getUsers(notFoundIds)
+          .then((getUsersResult) => {
+            expect(getUsersResult.users).to.deep.equal([]);
+            expect(getUsersResult.notFound).to.deep.equal(notFoundIds);
+          });
+      });
+
+      it('should be rejected when given an invalid uid', () => {
```
hiranya911: I think we ought to clearly document this behavior. If any of the identifiers are invalid/malformed, it results in a total failure, as opposed to the partial failure that occurs when some identifiers are not available.

test/unit/auth/auth.spec.ts
```diff
@@ -1130,6 +1131,84 @@ AUTH_CONFIGS.forEach((testConfig) => {
...
+          .with.property('code', 'auth/invalid-phone-number');
+      });
+
+      it('should be rejected when given an invalid provider', () => {
+        expect(() => auth.getUsers([{providerUid: '', providerId: ''}]))
+          .to.throw(FirebaseAuthError)
+          .with.property('code', 'auth/invalid-provider-id');
+      });
+
+      it('should be rejected when given a single bad identifier', () => {
```
hiranya911: Also add a happy path test case, where all IDs are valid and all users are present.

test/unit/auth/auth.spec.ts
```diff
@@ -1207,6 +1286,63 @@ AUTH_CONFIGS.forEach((testConfig) => {
...
+          .with.property('code', 'auth/maximum-user-count-exceeded');
+      });
+
+      it('should immediately fail given an invalid id', () => {
+        expect(() => auth.deleteUsers(['too long ' + ('.' as any).repeat(128)]))
+          .to.throw(FirebaseAuthError)
+          .with.property('code', 'auth/invalid-uid');
+      });
+
+      it('should index errors correctly in result', async () => {
```
hiranya911: Add happy path test case

test/integration/auth.spec.ts
```diff
@@ -164,6 +164,136 @@ describe('admin.auth', () => {
...
+    function mapUserRecordsToUidEmailPhones(values: Array<{ uid: string, email?: string, phoneNumber?: string}>) {
+      return values.map((ur) => ({ uid: ur.uid, email: ur.email, phoneNumber: ur.phoneNumber }));
+    }
+
+    const testUser1 = { uid: 'uid1', email: 'user1@example.com', phoneNumber: '+15555550001' };
+    const testUser2 = { uid: 'uid2', email: 'user2@example.com', phoneNumber: '+15555550002' };
+    const testUser3 = { uid: 'uid3', email: 'user3@example.com', phoneNumber: '+15555550003' };
+    const usersToCreate = [ testUser1, testUser2, testUser3 ];
+
+    const importUser1: admin.auth.UserImportRecord = {
```
hiranya911: Add a comment why this is here. I assume it's because you wanted a test user with a specific provider config.

test/integration/auth.spec.ts
```diff
@@ -164,6 +164,136 @@ describe('admin.auth', () => {
...
+    it('returns found users and ignores non-existing users', async () => {
+      const users = await admin.auth().getUsers([
+          { uid: 'uid1' },
+          { uid: 'uid_that_doesnt_exist' },
+          { uid: 'uid3' },
+        ])
+        .then((getUsersResult) => getUsersResult.users)
+        .then(mapUserRecordsToUidEmailPhones);
+
+      expect(users).to.have.deep.members([testUser1, testUser3]);
```
hiranya911: Should also assert notFound users

test/integration/auth.spec.ts
```diff
@@ -1191,6 +1321,63 @@ describe('admin.auth', () => {
...
+    });
+
+    it('deletes users that exist even when non-existing users also specified', async () => {
+      const uid1 = await admin.auth().createUser({}).then((ur) => ur.uid);
+      const uid2 = 'uid-that-doesnt-exist';
+      const ids = [{uid: uid1}, {uid: uid2}];
+
+      return admin.auth().deleteUsers([uid1, uid2])
+        .then((deleteUsersResult) => {
+          expect(deleteUsersResult.successCount).to.equal(2);
```
hiranya911: This doesn't look right.

test/integration/auth.spec.ts
```diff
@@ -1191,6 +1321,63 @@ describe('admin.auth', () => {
...
+
+          return admin.auth().getUsers(ids);
+        })
+        .then((getUsersResult) => {
+          expect(getUsersResult.users).to.have.length(0);
+          expect(getUsersResult.notFound).to.have.deep.members(ids);
+        });
+    });
+
+    it('is idempotent', async () => {
```
hiranya911: Was this the decided upon behavior?

test/unit/auth/auth.spec.ts
rsgowman: Done, though I've combined that with a notFound user. (Let me know if you want that to be explicitly split out.)

test/unit/auth/auth.spec.ts

test/integration/auth.spec.ts

test/integration/auth.spec.ts

test/integration/auth.spec.ts
rsgowman: Looks ok to me? deleteUsers is idempotent, so deleteUsers(['i_exist', 'but_i_dont']) should result in a success count of 2, i.e. both users no longer exist. Right?

test/integration/auth.spec.ts
rsgowman: Yes; see the clarifications section of the api proposal: "DeleteUsers() call that specifies only users that don’t exist (or no users at all) results in an empty result set rather than an error. This allows developers to not need to special case this situation, as well as makes this call idempotent. (The returned SuccessCount value would indicate the number of users that no longer exist, regardless of their initial state, i.e. it would also include users that already did not exist.)"

If you wanted to push back on that, we still could; we haven't fully committed to this behaviour yet. (fwiw, I prefer it as is.) In order to go down that route, we'd have to make a backend change too though, since the backend also behaves this way and doesn't provide any indication as to whether an account was deleted or never existed in the first place.

test/integration/auth.spec.ts
hiranya911: Ah, idempotency was what I was missing 👍

test/integration/auth.spec.ts
hiranya911: I see. My only concern is that it's different from how the existing `deleteUser()` API works. But I also think the behavior proposed here is actually better/convenient for bulk deletes.

src/auth/auth-api-request.ts
```diff
@@ -463,12 +468,21 @@ export const FIREBASE_AUTH_DOWNLOAD_ACCOUNT = new ApiSettings('/accounts:batchGe
...
+  }>;
+}
 
 /** Instantiates the getAccountInfo endpoint settings. */
 export const FIREBASE_AUTH_GET_ACCOUNT_INFO = new ApiSettings('/accounts:lookup', 'POST')
   // Set request validator.
-  .setRequestValidator((request: any) => {
-    if (!request.localId && !request.email && !request.phoneNumber) {
+  .setRequestValidator((request: GetAccountInfoRequest) => {
+    if (!request.localId && !request.email && !request.phoneNumber && !request.federatedUserId) {
```
bojeil-google: Do you validate the `federatedUserId` content?

src/auth/auth-api-request.ts
```diff
@@ -481,6 +495,21 @@ export const FIREBASE_AUTH_GET_ACCOUNT_INFO = new ApiSettings('/accounts:lookup'
     }
   });
 
+/**
+ * Instantiates the getAccountInfo endpoint settings for use when fetching info
+ * for multiple accounts.
+ */
+export const FIREBASE_AUTH_GET_ACCOUNTS_INFO = new ApiSettings('/accounts:lookup', 'POST')
```
bojeil-google: Why do you need to duplicate `FIREBASE_AUTH_GET_ACCOUNT_INFO` above? It is the same underlying API with similar request /response format.

src/auth/auth-api-request.ts
```diff
@@ -492,6 +521,46 @@ export const FIREBASE_AUTH_DELETE_ACCOUNT = new ApiSettings('/accounts:delete',
...
+  message?: string;
+}
+
+export interface BatchDeleteAccountsResponse {
+  errors?: BatchDeleteErrorInfo[];
+}
+
+export const FIREBASE_AUTH_BATCH_DELETE_ACCOUNTS = new ApiSettings('/accounts:batchDelete', 'POST')
+  .setRequestValidator((request: BatchDeleteAccountsRequest) => {
+    if (!request.localIds) {
```
bojeil-google: shouldn't you also validate the `force` field?

src/auth/auth-api-request.ts
```diff
@@ -492,6 +521,46 @@ export const FIREBASE_AUTH_DELETE_ACCOUNT = new ApiSettings('/accounts:delete',
...
+      throw new FirebaseAuthError(
+        AuthClientErrorCode.INTERNAL_ERROR,
+        'INTERNAL ASSERT FAILED: Server request is missing user identifiers');
+    }
+  })
+  .setResponseValidator((response: BatchDeleteAccountsResponse) => {
+    const errors = response.errors || [];
+    errors.forEach((batchDeleteErrorInfo) => {
+      if (batchDeleteErrorInfo.index === undefined) {
```
bojeil-google: nit: `typeof batchDeleteErrorInfo.index === 'undefined'`

src/index.d.ts
```diff
@@ -1487,6 +1573,20 @@ declare namespace admin.auth {
...
+    /**
+     * Deletes the users specified by the given identifiers.
+     *
+     * Deletes multiple existing users. Deleting a non-existing user won't
+     * generate an error. (i.e. this method is idempotent.)
+     *
+     * @param uids The `uids` corresponding to the users to delete.
+     *
+     * @return A Promise that resolves to the total number of successful/failed
+     *     deletions, as well as the array of errors that correspond to the
```
bojeil-google: corresponds

src/index.d.ts
```diff
@@ -1487,6 +1573,20 @@ declare namespace admin.auth {
      */
     deleteUser(uid: string): Promise<void>;
 
+    /**
+     * Deletes the users specified by the given identifiers.
```
bojeil-google: identifiers in this case is always uids.

src/index.d.ts
```diff
@@ -1487,6 +1573,20 @@ declare namespace admin.auth {
...
+     * Deletes multiple existing users. Deleting a non-existing user won't
+     * generate an error. (i.e. this method is idempotent.)
+     *
+     * @param uids The `uids` corresponding to the users to delete.
+     *
+     * @return A Promise that resolves to the total number of successful/failed
+     *     deletions, as well as the array of errors that correspond to the
+     *     failed deletions.
+     */
+    deleteUsers(uids: string[]): Promise<admin.auth.DeleteUsersResult>;
```
bojeil-google: I am a bit confused. Why is `forceDelete` not included? This was approved by the API council in https://docs.google.com/document/d/1mG0ilmh8H-C6DYXA9oOQUro2MP0on_9k7Qf4oCLRyBk/

src/utils/error.ts
bojeil-google: I am confused about this. This was approved in https://docs.google.com/document/d/1mG0ilmh8H-C6DYXA9oOQUro2MP0on_9k7Qf4oCLRyBk/edit#

src/auth/auth-api-request.ts
```diff
@@ -725,6 +794,44 @@ export abstract class AbstractAuthRequestHandler {
...
+    if (!validator.isPhoneNumber(id.phoneNumber)) {
+      throw new FirebaseAuthError(AuthClientErrorCode.INVALID_PHONE_NUMBER);
+    }
+    request.phoneNumber ? request.phoneNumber.push(id.phoneNumber) : request.phoneNumber = [id.phoneNumber];
+    return request;
+  }
+
+  private static addProviderToRequest(id: ProviderIdentifier, request: GetAccountInfoRequest): GetAccountInfoRequest {
+    if (!validator.isNonEmptyString(id.providerUid) || !validator.isNonEmptyString(id.providerId)) {
+      throw new FirebaseAuthError(AuthClientErrorCode.INVALID_PROVIDER_ID);
```
bojeil-google: I think invalid provider ID is different that invalid provider uid. The error is confusing here.

src/auth/auth-api-request.ts
```diff
@@ -811,6 +918,44 @@ export abstract class AbstractAuthRequestHandler {
...
+   *
+   * @param {UserIdentifier[]} identifiers The identifiers indicating the users
+   *     to be looked up. Must have <= 100 entries.
+   * @param {Promise<object>} A promise that resolves with the set of successfully
+   *     looked up users. Possibly empty if no users were looked up.
+   */
+  public getAccountInfoByIdentifiers(identifiers: UserIdentifier[]): Promise<object> {
+    if (identifiers.length === 0) {
+      return Promise.resolve({users: []});
+    } else if (identifiers.length > 100) {
```
bojeil-google: Maybe define 100 as a constant like we do for other similar max fields?

src/auth/auth-api-request.ts
```diff
@@ -908,6 +1053,30 @@ export abstract class AbstractAuthRequestHandler {
     return this.invokeRequestHandler(this.getAuthUrlBuilder(), FIREBASE_AUTH_DELETE_ACCOUNT, request);
   }
 
+  public deleteAccounts(uids: string[], force: boolean): Promise<BatchDeleteAccountsResponse> {
+    if (uids.length === 0) {
+      return Promise.resolve({});
+    } else if (uids.length > 100) {
```
bojeil-google: Same here regarding defining a const for this upper limit.

src/auth/auth.ts
```diff
@@ -200,6 +225,61 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> {
...
+        AuthClientErrorCode.INVALID_ARGUMENT, '`identifiers` parameter must be an array');
+    }
+    return this.authRequestHandler
+      .getAccountInfoByIdentifiers(identifiers)
+      .then((response: any) => {
+        /**
+         * Checks if the specified identifier is within the list of
+         * UserRecords.
+         */
+        const isUserFound = ((id: UserIdentifier, urs: UserRecord[]): boolean => {
```
bojeil-google: what does `urs` stand for? Maybe pick a better name?

src/auth/auth.ts
```diff
@@ -271,6 +351,52 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> {
...
+          if (batchDeleteErrorInfo.index === undefined) {
+            throw new FirebaseAuthError(
+              AuthClientErrorCode.INTERNAL_ERROR,
+              'Corrupt BatchDeleteAccountsResponse detected');
+          }
+
+          const errMsgToError = (msg?: string): FirebaseAuthError => {
+            // We unconditionally set force=true, so the 'NOT_DISABLED' error
+            // should not be possible.
+            if (msg && msg.startsWith('NOT_DISABLED :')) {
```
hiranya911: Nit:

```
const code = msg && msg.startsWith('NOT_DISABLED') ? 
  AuthClientErrorCode.USER_NOT_DISABLED : AuthClientErrorCode.INTERNAL_ERROR;
return new FirebaseAuthError(code, ...);
```

src/index.d.ts
```diff
@@ -1529,6 +1629,23 @@ declare namespace admin.auth {
      */
     getUserByPhoneNumber(phoneNumber: string): Promise<admin.auth.UserRecord>;
 
+    /**
+     * Gets the user data corresponding to the specified identifiers.
+     *
+     * There are no ordering guarantees; in particular, the nth entry in the result list is not
+     * guaranteed to correspond to the nth entry in the input parameters list.
+     *
+     * Only a maximum of 100 identifiers may be supplied. If more than 100 identifiers are supplied,
```
hiranya911: I think this method rejects when array.length > 100. Let's verify and update the comment.

src/index.d.ts
```diff
@@ -1487,6 +1573,20 @@ declare namespace admin.auth {
      */
     deleteUser(uid: string): Promise<void>;
 
+    /**
+     * Deletes the users specified by the given uids.
+     *
+     * Deletes multiple existing users. Deleting a non-existing user won't
```
hiranya911: Please document array length constraints.

src/auth/auth.ts
rsgowman: Done.

src/index.d.ts
rsgowman: i.e. throw vs failed promise? It should throw. (There's a test to ensure it here: test/unit/auth/auth.spec.ts:1142.)

src/index.d.ts
rsgowman: Oops; I hadn't updated this based on the design doc. I've added the full docs now. (Which includes the 1000 thing, but also touches on the rate limitation.)

test/unit/auth/auth.spec.ts
```diff
@@ -1130,6 +1131,132 @@ AUTH_CONFIGS.forEach((testConfig) => {
...
 
+    describe('getUsers()', () => {
+      let stubs: sinon.SinonStub[] = [];
+
+      afterEach(() => {
+        stubs.forEach((stub) => stub.restore());
+        stubs = [];
+      });
+
+      it('should be rejected when given more than 100 identifiers', () => {
```
hiranya911: Nit: `should throw`

src/index.d.ts
hiranya911: Ah, I see now. Thanks. The terminology used in the test case descriptions could need some work though.

test/unit/auth/auth.spec.ts

src/index.d.ts
```diff
@@ -921,6 +940,32 @@ declare namespace admin.auth {
...
+   */
+  interface DeleteUsersResult {
+    /**
+     * The number of user records that failed to be deleted (possibly zero).
+     */
+    failureCount: number;
+
+    /**
+     * The number of users that were deleted successfully (possibly zero).
+     * Users that did not exist prior to calling deleteUsers() will be
```
egilmorez: Suggest "are considered"

src/index.d.ts
```diff
@@ -921,6 +940,32 @@ declare namespace admin.auth {
...
+     * considered to be successfully deleted.
+     */
+    successCount: number;
+
+    /**
+     * A list of FirebaseArrayIndexError instances describing the errors that
```
egilmorez: Suggest backticks.

src/index.d.ts
```diff
@@ -1487,6 +1573,31 @@ declare namespace admin.auth {
      */
     deleteUser(uid: string): Promise<void>;
 
+    /**
+     * Deletes the users specified by the given uids.
+     *
+     * Deleting a non-existing user won't generate an error. (i.e. this method
+     * is idempotent.) Non-existing users will be considered to be successfully
```
egilmorez: Suggest "are considered . . . are counted"

src/index.d.ts
```diff
@@ -1487,6 +1573,31 @@ declare namespace admin.auth {
...
+     * deleted, and will therefore be counted in the
+     * DeleteUsersResult.successCount value.
```
egilmorez: Suggest backticks.

src/index.d.ts
```diff
@@ -1487,6 +1573,31 @@ declare namespace admin.auth {
...
+     *
+     * Only a maximum of 1000 identifiers may be supplied. If more than 1000
+     * identifiers are supplied, this method will immediately throw a
```
egilmorez: Suggest "method throws"

src/index.d.ts
```diff
@@ -1529,6 +1640,23 @@ declare namespace admin.auth {
...
     getUserByPhoneNumber(phoneNumber: string): Promise<admin.auth.UserRecord>;
 
+    /**
+     * Gets the user data corresponding to the specified identifiers.
+     *
+     * There are no ordering guarantees; in particular, the nth entry in the result list is not
+     * guaranteed to correspond to the nth entry in the input parameters list.
+     *
+     * Only a maximum of 100 identifiers may be supplied. If more than 100 identifiers are supplied,
+     * this method will immediately throw a FirebaseAuthError.
```
egilmorez: Suggest "method throws"

src/index.d.ts
rsgowman: Done.

src/index.d.ts
rsgowman: Done.

src/index.d.ts
rsgowman: Done.

src/index.d.ts
rsgowman: Done.

src/index.d.ts
rsgowman: Done.

test/integration/auth.spec.ts
```diff
@@ -168,6 +168,138 @@ describe('admin.auth', () => {
...
+        uid: 'lastRefreshTimeUser',
+        email: 'lastRefreshTimeUser@example.com',
+        password: 'p4ssword',
+      });
+
+      try {
+        // New users should not have a lastRefreshTime set.
+        expect(newUserRecord.metadata.lastRefreshTime).to.be.null;
+
+        // Login to cause the lastRefreshTime to be set.
```
RelaxedPear: Suggest:
"// Log in to set the lastRefreshTime."

src/index.d.ts
```diff
@@ -492,10 +492,16 @@ declare namespace admin.auth {
 
     /**
      * The date the user was created, formatted as a UTC string.
-     *
      */
     creationTime: string;
 
+    /**
+     * The time at which the user was last active (ID token refreshed), or null
```
RelaxedPear: "The time at which the user was last active (ID token refreshed), formatted as a UTC Date string (e.g., 'Sat, 03 Feb 2001 04:05:06 GMT'). Returns null if the user was never active."

src/index.d.ts
```diff
@@ -1487,6 +1573,30 @@ declare namespace admin.auth {
      */
     deleteUser(uid: string): Promise<void>;
 
+    /**
+     * Deletes the users specified by the given uids.
+     *
+     * Deleting a non-existing user won't generate an error. (i.e. this method
```
RelaxedPear: Deleting a non-existing user won't generate an error (i.e. this method is idempotent).

rsgowman: Done.

src/index.d.ts
rsgowman: Done.

test/integration/auth.spec.ts
rsgowman: Done.

src/index.d.ts
```diff
@@ -604,7 +604,7 @@ declare namespace admin.credential {
    */
   interface Credential {
 
-    /**
+   /**
```
bojeil-google, line 607: Comment alignment seem to be off here.

bojeil-google:
Granted coverage is still adequate but the testing format does not align with the existing pattern used here. Personally I prefer sticking with the current pattern.

The pattern has been to test the RPC logic in auth-api-request.spec.ts
Then test the adequate auth-api-request calls are made in auth.spec.ts.

rsgowman:
> Granted coverage is still adequate but the testing format does not align with the existing pattern used here. Personally I prefer sticking with the current pattern.
> 
> The pattern has been to test the RPC logic in auth-api-request.spec.ts
> Then test the adequate auth-api-request calls are made in auth.spec.ts.

Mostly done. I've moved some tests over entirely (where possible), kept some tests in auth.spec.ts (when those tests test behaviour entirely confined to auth.ts) and split some between the two (as auth.ts modifies the results from auth-api-request.ts, most notably adding a notFound parameter to the results as well as parsing everything in to a UserRecord.)

rsgowman:
> LG, thanks!
> 
> There's an instance of "deleteUsers()" in line is 680 of auth.d.ts that should be backticked, but we can get that later if it's a PITA.

Not a problem; fixed.

test/unit/auth/auth-api-request.spec.ts
```diff
@@ -1727,6 +1920,35 @@ AUTH_REQUEST_HANDLER_TESTS.forEach((handler) => {
       });
     });
 
+    describe('deleteAccounts', () => {
```
bojeil-google, line 1923: Can you add tests for rpc logic for this API?

test/unit/auth/auth-api-request.spec.ts
```diff
@@ -1947,6 +1950,20 @@ AUTH_REQUEST_HANDLER_TESTS.forEach((handler) => {
           .to.throw(FirebaseAuthError)
           .with.property('code', 'auth/invalid-uid');
       });
+
+      it('should be fulfilled given a valids uids', async () => {
```
bojeil-google, line 1954: given valid

PR #728, opened 2019-12-13
Using async project ID discovery in ProjectManagement and InstanceId

hiranya911:
Continuation of #715

Using `utils.findProjectId()` in ProjectManagement and InstanceId.

### DIFF
--- a/src/project-management/project-management.ts
+++ b/src/project-management/project-management.ts
@@ -43,11 +43,11 @@ class ProjectManagementInternals implements FirebaseServiceInternalsInterface {
  * ProjectManagement service bound to the provided app.
  */
 export class ProjectManagement implements FirebaseServiceInterface {
+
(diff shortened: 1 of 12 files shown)


src/project-management/project-management.ts
```diff
@@ -43,11 +43,11 @@ class ProjectManagementInternals implements FirebaseServiceInternalsInterface {
  * ProjectManagement service bound to the provided app.
  */
 export class ProjectManagement implements FirebaseServiceInterface {
+
   public readonly INTERNAL: ProjectManagementInternals = new ProjectManagementInternals();
 
-  private readonly resourceName: string;
-  private readonly projectId: string;
   private readonly requestHandler: ProjectManagementRequestHandler;
+  private projectId: string;
```
nbegley: It looks like this should be `private projectId?: string;` now

hiranya911: Doesn't seem to be required for string and other primitive types.

PR #730, opened 2019-12-16
Refactored the credentials implementation

hiranya911:
Refactoring and simplifying the existing credential implementation for better design and async project ID discovery.

* `CertCredential` class was renamed to `ServiceAccountCredential`.
* `Certificate` type is renamed to `ServiceAccount`, and is no longer exported. Instead the `projectId`, `clientEmail` and `privateKey` properties are directly exposed from the `ServiceAccountCredential` class.
* `MetadataServiceCredential` class was renamed to `ComputeEngineCredential`. This class now also has `getProjectId(): Promise<string>` method to discover the project ID from the Compute Engine environment.
* `ApplicationDefaultCredential` class has been removed. Instead now there's a `getApplicationDefault()` function that returns a `Credential` implementation suitable for the environment.

### DIFF
--- a/src/auth/credential.ts
+++ b/src/auth/credential.ts
@@ -29,7 +29,8 @@ const GOOGLE_AUTH_TOKEN_PATH = '/o/oauth2/token';
 
 // NOTE: the Google Metadata Service uses HTTP over a vlan
 const GOOGLE_METADATA_SERVICE_HOST = 'metadata.google.internal';
-const GOOGLE_METADATA_SERVICE_PATH = '/computeMetadata/v1/instance/service-accounts/default/token';
+const GOOGLE_METADATA_SERVICE_TOKEN_PATH = '/computeMetadata/v1/instance/service-accounts/default/token';
+const GOOGLE_METADATA_SERVICE_PROJECT_ID_PATH = '/computeMetadata/v1/project/project-id';
 
 const configDir = (() => {
   // Windows has a dedicated low-rights location for apps at ~/Application Data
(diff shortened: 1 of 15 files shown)


src/auth/token-generator.ts
```diff
@@ -82,28 +82,19 @@ interface JWTBody {
...
-    if (!certificate) {
+  constructor(private readonly credential: ServiceAccountCredential) {
+    if (!credential) {
       throw new FirebaseAuthError(
         AuthClientErrorCode.INVALID_CREDENTIAL,
-        'INTERNAL ASSERT: Must provide a certificate to initialize ServiceAccountSigner.',
+        'INTERNAL ASSERT: Must provide a service account credential to initialize ServiceAccountSigner.',
       );
     }
-    if (!validator.isNonEmptyString(certificate.clientEmail) || !validator.isNonEmptyString(certificate.privateKey)) {
```
lahirumaramba: Do we not need this validation with the new changes?

hiranya911: These validations happen during the construction of `ServiceAccountCredential` object. So we can assume, they are all present.

src/auth/credential.ts
```diff
@@ -364,116 +263,162 @@ export class RefreshTokenCredential implements Credential {
...
- * ApplicationDefaultCredential implements the process for loading credentials as
- * described in https://developers.google.com/identity/protocols/application-default-credentials
- */
-export class ApplicationDefaultCredential implements FirebaseCredential {
-  private credential_: Credential;
+  constructor(json: object) {
+    copyAttr(this, json, 'clientId', 'client_id');
+    copyAttr(this, json, 'clientSecret', 'client_secret');
+    copyAttr(this, json, 'refreshToken', 'refresh_token');
+    copyAttr(this, json, 'type', 'type');
```
bojeil-google: I am wondering why the `type` property is needed.

src/auth/credential.ts
```diff
@@ -364,116 +263,162 @@ export class RefreshTokenCredential implements Credential {
...
+      return new RefreshTokenCredential(refreshToken, httpAgent);
+    }
   }
 
-  public getCertificate(): Certificate | null {
-    return tryGetCertificate(this.credential_);
+  return new ComputeEngineCredential(httpAgent);
+}
+
+function copyAttr(to: {[key: string]: any}, from: {[key: string]: any}, key: string, alt: string) {
```
bojeil-google: This seems like it should be a utility function. Also add some javadoc for some of these function that may not be trivial to figure out. For example, it was not clear to me what `alt` stands for.

test/unit/auth/credential.spec.ts
```diff
@@ -406,24 +377,30 @@ describe('Credential', () => {
...
     });
 
     it('should throw if a the gcloud login cache is invalid', () => {
       delete process.env.GOOGLE_APPLICATION_CREDENTIALS;
       fsStub = sinon.stub(fs, 'readFileSync').returns('invalidjson');
-      expect(() => new ApplicationDefaultCredential()).to.throw(Error);
+      expect(() => getApplicationDefault()).to.throw(Error);
+    });
+
+    it('should throw if a the credentials file content is not an object', () => {
```
bojeil-google: if the

hiranya911: Done

src/auth/credential.ts
hiranya911: Added some jsdoc comments. This was also an pre-existing function.

src/auth/credential.ts
hiranya911: Yes, it's kind of redundant. It's always going to have the constant value "authorized_user" in this case. I think this type is meant to represent the refresh token JSON file, which usually contains a "type" field. I'd keep it as is, since removal of the property is not really related to this refactor. We can remove it in a future PR if needed.

src/auth/credential.ts
```diff
@@ -276,46 +122,109 @@ export class CertCredential implements FirebaseCredential {
...
+      // Throw a nicely formed error message if the file contents cannot be parsed
+      throw new FirebaseAppError(
+        AppErrorCodes.INVALID_CREDENTIAL,
+        'Failed to parse service account json file: ' + error,
+      );
+    }
+  }
+
+  constructor(json: object) {
+    if (typeof json !== 'object' || json === null) {
```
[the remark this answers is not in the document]

PR #755, opened 2020-01-07
Discovering project ID from the local metadata service

hiranya911:
Support for auto discovering the project ID when running on Compute Engine, Cloud Run etc.

Verified by running a build on a Compute Engine instance.

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -56,7 +56,7 @@ export function addReadonlyGetter(obj: object, prop: string, value: any): void {
 }
 
 /**
- * Returns the Google Cloud project ID associated with a Firebase app, if its explicitly
+ * Returns the Google Cloud project ID associated with a Firebase app, if it's explicitly
  * specified in either the Firebase app options, credentials or the local environment.
  * Otherwise returns null.
  *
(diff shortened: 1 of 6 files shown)


src/utils/index.ts
```diff
@@ -56,14 +56,15 @@ export function addReadonlyGetter(obj: object, prop: string, value: any): void {
 }
 
 /**
- * Determines the Google Cloud project ID associated with a Firebase app by examining
- * the Firebase app options, credentials and the local environment in that order.
+ * Returns the Google Cloud project ID associated with a Firebase app, if its explicitly
```
bojeil-google: it's

test/unit/auth/credential.spec.ts
```diff
@@ -328,6 +329,47 @@ describe('Credential', () => {
         });
       });
     });
+
+    it('should cache discovered project id', () => {
+      const expected = 'test-project-id';
```
bojeil-google: nit: Rename `expected` to `expectedProjectId` here and below.

src/utils/index.ts
hiranya911: Done

test/unit/auth/credential.spec.ts
hiranya911: Done

hiranya911:
@bojeil-google I've made the suggested changes.

PR #762, opened 2020-01-15
Fixing duplicate descriptions in AppOptions.html

egilmorez:
This fixes a duplicated and incorrect description for storageBucket, and adds some detail on specifying a valid value.

Thanks!


egilmorez:
Staged internally at: https://firebase.devsite.corp.google.com/docs/reference/admin/node/admin.AppOptions.html#storagebucket

PR #768, opened 2020-01-19
fix(auth): Improve customClaims Typing

wSedlacek:
With the Type Definition of Object it leaves it on the consumer of `firebase-admin` to override the type definition to access things like `customClaims.admin`, many users may opt to use any here breaking type safety.

With that in mind would it be possible to use something like this typing in order to still allow type safety to some extent without needing the user to override the typing to consume it?

Error without this change
Screen Shot 2020-01-18 at 22 13 44

Example code overriding the typing
Screen Shot 2020-01-18 at 22 13 55

RELEASE NOTE: Updated typings of the `UserRecord.customClaims` attribute to allow indexing by string keys.


hiranya911:
Looks reasonable to me. @bojeil-google what do you think?

Another aspect to consider is, does this change constitute a breaking change for some developers.

bojeil-google:
Looks good to me.

wSedlacek:
Anything more I can do to get this pushed through?

wSedlacek:
Thank you!

PR #769, opened 2020-01-20
feat(auth): Implement getUserByProviderId

rsgowman:
RELEASE NOTE: Added a new `getUserByProviderId()` to lookup user accounts by their providers.


src/auth/auth.ts
```diff
@@ -200,6 +200,36 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> {
...
+   * @return A promise fulfilled with the user data corresponding to the
+   *   given provider id.
+   */
+  public getUserByProviderId(providerId: string, providerUid: string): Promise<UserRecord> {
+    // Although we don't really advertise it, we want to also handle
+    // non-federated idps with this call. So if we detect one of them, we'll
+    // reroute this request appropriately.
+    if (providerId === 'phone') {
+      return this.getUserByPhoneNumber(providerUid);
+    } else if (providerId === 'email') {
```
nrsim: Could authRequestHandler.getAccountByFederatedId be renamed to getAccountByProviderId (or Uid?) as well?
The logic for 'phone' and 'email' would need to go into authRequestHandler in that case.

src/auth/auth.ts
```diff
@@ -200,6 +200,36 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> {
...
+   * for code samples and detailed documentation.
+   *
+   * @param providerId The provider ID, for example, "google.com" for the
+   *   Google provider.
+   * @param providerUid The user identifier for the given provider.
+   *
+   * @return A promise fulfilled with the user data corresponding to the
+   *   given provider id.
+   */
+  public getUserByProviderId(providerId: string, providerUid: string): Promise<UserRecord> {
```
nrsim: What was the final decision on this name? Uid?

test/integration/auth.spec.ts
```diff
@@ -66,6 +66,7 @@ const mockUserData = {
   displayName: 'Random User ' + newUserUid,
   photoURL: 'http://www.example.com/' + newUserUid + '/photo.png',
   disabled: false,
+
```
nrsim: nit: remove the blank line

test/integration/auth.spec.ts
```diff
@@ -168,6 +169,44 @@ describe('admin.auth', () => {
...
+        email: 'user@example.com',
+        phoneNumber: '+15555550000',
+        photoURL: 'http://example.com/user',
+        toJSON: () => { throw new Error('Unimplemented'); },
+        providerId: 'google.com',
+        uid: 'google_uid',
+      }],
+    };
+
+    await safeDelete(importUser.uid);
```
nrsim: There is another safeDelete below.

test/integration/auth.spec.ts
```diff
@@ -356,6 +395,11 @@ describe('admin.auth', () => {
       .should.eventually.be.rejected.and.have.property('code', 'auth/user-not-found');
   });
 
+  it('getUserByProviderId() fails when called with a non-existing federated id', () => {
```
nrsim: Add unit tests for invalid provider id/uid here as well?

src/auth/auth.ts
rsgowman: Technically neither, since the final decision hasn't been made yet. :)

But you're quite right that the current state uses Uid. Fixed. (Nice catch.)

src/auth/auth.ts
rsgowman: It could. But I've been interpreting the auth-api-request.ts file as the object that implements the backend API. And since the backend makes a distinction between federated ids and non-federated ids, I thought that worth preserving here. (NB: my interpretation may be incorrect.)

Though it probably should at least use Uid rather than Id. (Fixed).

test/integration/auth.spec.ts

test/integration/auth.spec.ts
rsgowman: I think they're already covered by the unit tests. Specifically, these ones:
'should be rejected given no federated id'
'should be rejected given an invalid federated id'
'should be rejected given an invalid federated uid'

(Or have I missed something?)

src/auth/auth.ts
nrsim: My personal preference would be to avoid the federated[Id/Uid] terminology where possible, so that it doesn't get confused with the more general provider[Id/Uid] term.

In the long run, we should probably move towards renaming the field in the backend api as well. So, it would be reasonable to move the logic to auth-api-request.ts even if the file represents the backend implementation API.

Having said that, I would be OK with keeping the logic in auth.

test/integration/auth.spec.ts
nrsim: Sorry, I meant to ask for unit tests in test/unit/auth/auth-api-request.spec.ts.

src/auth/auth-api-request.ts
```diff
@@ -811,6 +811,21 @@ export abstract class AbstractAuthRequestHandler {
     return this.invokeRequestHandler(this.getAuthUrlBuilder(), FIREBASE_AUTH_GET_ACCOUNT_INFO, request);
   }
 
+  public getAccountInfoByFederatedUid(federatedId: string, federatedUid: string): Promise<object> {
```
hiranya911: Nit: The 2 args are only off by one character. Can we make them a bit more distinct?

src/auth/auth.ts
```diff
@@ -200,6 +200,36 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> {
...
+   * for code samples and detailed documentation.
+   *
+   * @param providerId The provider ID, for example, "google.com" for the
+   *   Google provider.
+   * @param providerUid The user identifier for the given provider.
+   *
+   * @return A promise fulfilled with the user data corresponding to the
+   *   given provider id.
+   */
+  public getUserByProviderUid(providerId: string, providerUid: string): Promise<UserRecord> {
```
hiranya911: Here too. Perhaps `provider` and `providerUid`?

test/integration/auth.spec.ts
```diff
@@ -168,6 +168,44 @@ describe('admin.auth', () => {
...
+        email: 'user@example.com',
+        phoneNumber: '+15555550000',
+        photoURL: 'http://example.com/user',
+        toJSON: () => { throw new Error('Unimplemented'); },
+        providerId: 'google.com',
+        uid: 'google_uid',
+      }],
+    };
+
+    await safeDelete(importUser.uid);
```
hiranya911: Use a random UID and drop this line.

test/unit/auth/auth.spec.ts
```diff
@@ -1130,6 +1130,120 @@ AUTH_CONFIGS.forEach((testConfig) => {
...
+      // Stubs used to simulate underlying api calls.
+      let stubs: sinon.SinonStub[] = [];
+      beforeEach(() => sinon.spy(validator, 'isEmail'));
+      afterEach(() => {
+        (validator.isEmail as any).restore();
+        _.forEach(stubs, (stub) => stub.restore());
+        stubs = [];
+      });
+
+      it('should be rejected given no federated id', () => {
```
hiranya911: Use consistent terminology. It appears at this level we should be using `provider` instead of `federated`.

src/auth/auth-api-request.ts
rsgowman: These are meant to match the proto parameters, so we can easily change them to more closely reflect that. Done.

Note that federatedId now becomes providerId... which sounds slightly looser, but I think is fine since the method itself has 'federated' right there in the name.

src/auth/auth.ts
rsgowman: This is what was agreed to in the api review, so I'm reluctant to change that. But if we were to do so:

`providerId` has prior art (UserINfo, AuthProviderConfig, getProviderConfig, deleteProviderConfig, updateProviderConfig) so I'd prefer to leave that alone.

`providerUid` however does not. Although a bit overloaded, we could use simply `uid` here instead. I think it's sufficiently clear.

If we do want to change this, we should also revisit the bulk get PR which uses similar names.

wdyt?

test/integration/auth.spec.ts

test/unit/auth/auth.spec.ts

test/integration/auth.spec.ts
rsgowman: Oh; I hadn't noticed that. Done.

(More thought required, but I suspect some of those tests may be redundant since we're already effectively testing them from the api level. It might be possible to eliminate some of them.)

src/auth/auth.ts
nrsim: I'm not a fan of uid, as it may be confused with Firebase uid (see getUser(uid: string), for example). Having a different name for provider uid would make the distinction explicit.

src/index.d.ts
```diff
@@ -1530,6 +1530,21 @@ declare namespace admin.auth {
      */
     getUserByPhoneNumber(phoneNumber: string): Promise<admin.auth.UserRecord>;
 
+    /**
+     * Gets the user data for the user corresponding to a given provider id.
```
egilmorez: Suggest caps like line 1539.

test/integration/auth.spec.ts
```diff
@@ -221,6 +221,42 @@ describe('admin.auth', () => {
...
+        photoURL: 'http://example.com/user',
+        providerId: 'google.com',
+        uid: 'google_uid',
+      }],
+    };
+
+    await admin.auth().importUsers([importUser]);
+
+    try {
+      await admin.auth().getUserByProviderUid('google.com', 'google_uid')
```
hiranya911, line 251: Should we also call this API with email and phone and see if we get the expected result?

hiranya911 (approved):
Thanks. This LGTM. Just one optional suggestion on integration tests.

test/integration/auth.spec.ts
rsgowman: I think we've already got that on lines 210,217. Or were you referring to something else?

hiranya911: I was referring to testing invocations like `getUserByProviderUid('email', emailAddr)`.

rsgowman: Ah, I understand now. Yeah; we should have those too. Done.

hiranya911: Thanks. LGTM 👍

src/auth/auth.ts
Thaina: @rsgowman Is it solidated? The released name also conflict with the method name I see here

https://firebase.google.com/support/release-notes/admin/node#9.5.0

rsgowman: Hi @Thaina; the final decision here was to use *Uid*, i.e. getUserByProviderUid. The release notes are incorrect (caused by my description of this PR not being updated to reflect this change). I'll see if I can get that fixed. Thanks for noticing!

PR #770, opened 2020-01-20
feat(auth): Add ability to link a federated ID with the `updateUser()` method.

rsgowman:


src/auth/auth-api-request.ts
```diff
@@ -265,6 +265,8 @@ function validateCreateEditRequest(request: any, uploadAccountRequest: boolean =
     phoneNumber: true,
     customAttributes: true,
     validSince: true,
+    // Pass linkProviderUserInfo only for updates (i.e. not for uploads.)
+    linkProviderUserInfo: !uploadAccountRequest,
```
nrsim: Add deleteProvider as well?

src/auth/auth-api-request.ts
```diff
@@ -410,6 +412,11 @@ function validateCreateEditRequest(request: any, uploadAccountRequest: boolean =
       validateProviderUserInfo(providerUserInfoEntry);
     });
   }
+
+  // linkProviderUserInfo must be a (single) UserInfo value.
```
nrsim: UserProvider?

src/auth/auth-api-request.ts
```diff
@@ -995,13 +1027,25 @@ export abstract class AbstractAuthRequestHandler {
...
+    if (typeof(request.providerToLink) !== 'undefined') {
+      request.linkProviderUserInfo = deepCopy(request.providerToLink);
+      delete request.providerToLink;
+
+      request.linkProviderUserInfo.rawId = request.linkProviderUserInfo.uid;
+      delete request.linkProviderUserInfo.uid;
+    }
+
+    if (typeof(request.providersToDelete) !== 'undefined') {
+      if (!validator.isArray(request.deleteProvider)) {
```
nrsim: Isn't this check redundant, considering the condition gets checked at validation?

src/auth/user-record.ts
```diff
@@ -132,6 +134,41 @@ export class UserInfo {
...
 }
 
+/**
+ * Represents a user identity provider that can be associated with a Firebase user.
+ */
+interface UserProvider {
+  /**
+   * The user identifier for the linked provider.
+   */
+  uid?: string;
```
nrsim: providerUid?

src/index.d.ts
```diff
@@ -686,6 +722,16 @@ declare namespace admin.auth {
...
      */
     photoURL?: string | null;
+
+    /**
+     * Links this user to the specified federated provider.
+     */
+    providerToLink?: UserProvider;
+
+    /**
+     * Unlinks this user from the specified federated providers.
```
nrsim: s/federated providers/providers?

test/integration/auth.spec.ts
```diff
@@ -323,22 +323,106 @@ describe('admin.auth', () => {
...
+      expect(userRecord.emailVerified).to.be.true;
+      expect(userRecord.displayName).to.equal(updatedDisplayName);
+      // Confirm expected email.
+      expect(userRecord.email).to.equal(updatedEmail);
+      // Confirm expected phone number.
+      expect(userRecord.phoneNumber).to.equal(updatedPhone);
+    });
+
+    it('can link/unlink with a federated provider', async () => {
+      const federatedUid = 'google_uid_' + generateRandomString(10);
```
nrsim: Could this be changed to googleUid/googleFederatedUid, to avoid an unnecessary confusion between federatedUid and providerUid?

test/unit/auth/auth.spec.ts
```diff
@@ -1401,10 +1407,204 @@ AUTH_CONFIGS.forEach((testConfig) => {
...
+              providerId: 'google.com',
+              rawId: 'google_uid',
+            },
+          });
+      });
+
+      it('should be rejected given an empty providersToDelete list', () => {
+        expect(() => {
+          auth.updateUser(uid, {
+            providersToDelete: [],
```
nrsim: Could this be a no-op instead?

test/unit/auth/auth.spec.ts
```diff
@@ -1401,10 +1407,204 @@ AUTH_CONFIGS.forEach((testConfig) => {
...
+            },
+          });
+          expect(invokeRequestHandlerStub).to.have.been.calledOnce.and.calledWith(
+            sinon.match.any, sinon.match.any, {
+              localId: uid,
+              phoneNumber: '+15555550001',
+            });
+        });
+
+        it('specifying both phoneNumber=null and providersToDelete=phone should be rejected', () => {
```
nrsim: nit: does this need to be reject?

test/unit/auth/auth.spec.ts
rsgowman: You're right; there's not really any reason to reject this. I've moved the test to an integration test (to ensure the backend is happy with the noop. If that ever changes, we can add code to special case this.)

test/unit/auth/auth.spec.ts
rsgowman: I don't know. It does indicate a slight misuse of our API as there's not really any situation where the user should be deleting both by setting phoneNumber=null and also by adding phone to providersToDelete. OTOH, the user's intent is clear in this case. See also the comment in the implementation (src/auth/auth.ts:updateUser:~313)

I'm inclined to be a bit conservative and reject this situation as an error and then relax it later if necessary. But am also willing to be convinced otherwise. We'll get a few more reviews before this lands, so we can probably just go with the consensus that results from that.

I've left this alone for now.

src/auth/auth-api-request.ts
rsgowman: Done.

src/auth/auth-api-request.ts
```diff
@@ -961,6 +968,31 @@ export abstract class AbstractAuthRequestHandler {
           'Properties argument must be a non-null object.',
         ),
       );
+    } else if (validator.isNonNullObject(properties.providerToLink)) {
```
nrsim: Aren't the validations at validateCreateEditRequest sufficient? Seems like the validation is added in two places.

hiranya911:
Hi @rsgowman. Is this PR ready for a round of review? Looks like the build failed due to an out of date API report. You might wan to run `npm run api-extractor:local` and commit the changes.

etc/firebase-admin.api.md
```diff
@@ -390,6 +400,7 @@ export namespace auth {
         tokensValidAfterTime?: string;
         uid: string;
     }
+    {};
```
hiranya911, line 403: Any idea what this is?

src/auth/auth-api-request.ts
hiranya911: This will achieve the same result without the if condition I think: 

```
request.deleteProvider = (request.deleteProvider || []).concat(request.providersToUnlink);
```

src/auth/auth.ts
```diff
@@ -351,6 +352,49 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
...
    */
   public updateUser(uid: string, properties: UpdateRequest): Promise<UserRecord> {
+    // Although we don't really advertise it, we want to also handle linking of
+    // non-federated idps with this call. So if we detect one of them, we'll
+    // adjust the properties parameter appropriately. This *does* imply that a
+    // conflict could arise, e.g. if the user provides a phoneNumber property,
+    // but also provides a providerToLink with a 'phone' provider id. In that
+    // case, we'll throw an error.
+    properties = deepCopy(properties);
+    if (validator.isNonNullObject(properties) && typeof properties.providerToLink !== 'undefined') {
```
hiranya911, line 362:
```
if (properties?.providerToLink) {

}
```

src/auth/auth.ts
```diff
@@ -351,6 +352,49 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
...
+          throw new FirebaseAuthError(
+            AuthClientErrorCode.INVALID_ARGUMENT,
+            "Both UpdateRequest.phoneNumber and UpdateRequest.providerToLink.providerId='phone' were set. To "
+            + 'link to a phone provider, only specify the UpdateRequest.phoneNumber field.');
+        }
+        properties.phoneNumber = properties.providerToLink.uid;
+        delete properties.providerToLink;
+      }
+    }
+    if (validator.isNonNullObject(properties) && typeof properties.providersToUnlink !== 'undefined') {
```
hiranya911, line 383:
```
if (properties?.providersToUnlink) {

}
```

test/unit/auth/auth.spec.ts
```diff
@@ -1557,8 +1563,194 @@ AUTH_CONFIGS.forEach((testConfig) => {
...
+                localId: 1,
+              }],
+            });
+        });
+        afterEach(() => {
+          invokeRequestHandlerStub.restore();
+          getAccountInfoByUidStub.restore();
+        });
+
+
```
hiranya911, line 1677: Extra empty line

hiranya911 (changes requested):
Looks pretty good. I just had minor nits.

etc/firebase-admin.api.md
rsgowman: None. (I just ran the tool.) I'll remove it and see if anything complains.

src/auth/auth-api-request.ts
rsgowman: The user could set this value to anything they like which could interact badly:

```javascript
request.deleteProvider = "foo"
request.providersToUnlink = [ "phone", "email" ]
request.deleteProvider = (request.deleteProvider || []).concat(request.providersToUnlink);
console.log(request.deleteProvider)
> ["foophone", "email"]
```

Such use would be pretty questionable since `deleteProvider` doesn't show up in the UpdateRequest interface... though that doesn't stop the user from setting it anyways for their own purposes. (Although we modify it here, we also take a copy first, so this won't impact the user's copy.) Given that, it seems easiest to just check to see if it's an array first.

Thinking about it a bit more though, another alternative would be to just set deleteProvider to undefined (or []) after we take the copy, thus ensuring any value that the user sets can't interfere with this value.

I've left it as is for now, though could switch to the alternative if you'd prefer.

src/auth/auth.ts
rsgowman: Done. I've also applied it to the providerToUnlink variable a few lines below.

src/auth/auth.ts
rsgowman: Oh; yeah, this is what I was referring to. :)

test/unit/auth/auth.spec.ts

etc/firebase-admin.api.md
hiranya911: Strange. Also removing it seems to cause a build break. Let's just add it for now. We can follow up with the API Extractor project separately to see what's going on (looks like a bug in the tool itself).

hiranya911:
@rsgowman looks like the manual update to the API report is building the break. Let's revert that change (see my comment above). Otherwise LGTM

src/auth/index.ts
```diff
@@ -384,6 +420,26 @@ export namespace auth {
...
      */
     multiFactor?: MultiFactorUpdateSettings;
+
+    /**
+     * Links this user to the specified provider.
+     *
+     * Linking a provider to an existing user account does not invalidate the
+     * refresh token of that account. In other words, the existing account
+     * would continue to be able to access resources, despite not having used
+     * the newly linked provider to login. If you wish to force the user to
```
egilmorez, line 430: As a verb or infinitive, this is two words: "log in"

hiranya911 (approved):
LGTM 👍

egilmorez (commented):
Only found one tiny nit.  Are all the user-facing comments in this one file Rich?  Thanks!

egilmorez (approved):
LG, thanks!

PR #780, opened 2020-02-06
chore: Experimental release flow based on Actions

hiranya911:
Just updated the workflow in the Python repo to use Node.js and NPM.

Sample execution: https://github.com/firebase/firebase-admin-node/runs/444354754


.github/workflows/ci.yml
```diff
@@ -1,6 +1,6 @@
 name: Continuous Integration
 
-on: [push, pull_request]
+on: push
```
lahirumaramba: Does the CI run on every `push` event even when there is no pull request?

hiranya911: Yes, it runs on every `git push` to remote.

PR #794, opened 2020-02-26
Enabling additional ESLint checks

hiranya911:
Enabled 8 additional checks. Most violations were auto fixed with `eslint --fix`, but a few required some manual intervention and/or ignore directives.


src/auth/user-import-builder.ts
```diff
@@ -359,7 +359,7 @@ export class UserImportBuilder {
         rounds,
       };
       break;
-
+    }
     case 'PBKDF_SHA1':
     case 'PBKDF2_SHA256':
```
lahirumaramba: Why are the braces `{ }` used only for some `case:` clauses and not others?

hiranya911: It's only used for case blocks that declare variables with const or let.

See https://eslint.org/docs/rules/no-case-declarations

PR #795, opened 2020-02-27
Added ListModels functionality for Firebase ML

ifielker:
Added ListModels functionality for Firebase ML

### DIFF
--- a/src/machine-learning/machine-learning-api-client.ts
+++ b/src/machine-learning/machine-learning-api-client.ts
@@ -161,11 +161,17 @@ export class MachineLearningApiClient {
       const err = new FirebaseMachineLearningError('invalid-argument', 'Invalid list filter.');
       return Promise.reject(err);
     }
-    if (typeof options.pageSize !== 'undefined' && !validator.isNumber(options.pageSize)) {
-      const err = new FirebaseMachineLearningError('invalid-argument', 'Invalid page size.');
-      return Promise.reject(err);
+    if (typeof options.pageSize !== 'undefined') {
+      if (!validator.isNumber(options.pageSize)) {
+        const err = new FirebaseMachineLearningError('invalid-argument', 'Invalid page size.');
+        return Promise.reject(err);
+      }
+      if (options.pageSize < 1 || options.pageSize > 100) {
+        const err = new FirebaseMachineLearningError(
+          'invalid-argument', 'Page size must be between 1 and 100.');
+        return Promise.reject(err);
+      }
     }
-
     if (typeof options.pageToken !== 'undefined' && !validator.isNonEmptyString(options.pageToken)) {
       const err = new FirebaseMachineLearningError(
         'invalid-argument', 'Next page token must be a non-empty string.');
--- a/src/machine-learning/machine-learning-api-client.ts
+++ b/src/machine-learning/machine-learning-api-client.ts
@@ -214,7 +214,6 @@ export class MachineLearningApiClient {
         return resp.data as T;
       })
       .catch((err) => {
-        //console.log(`SendRequest got error: ${JSON.stringify(err)}`);
         throw this.toFirebaseError(err);
       });
   }
--- a/src/machine-learning/machine-learning-api-client.ts
+++ b/src/machine-learning/machine-learning-api-client.ts
@@ -45,6 +45,13 @@ export interface ModelUpdateOptions extends ModelOptions {
   state?: { published?: boolean; };
 }
 
+/** Interface representing listModels options. */
+export interface ListModelsOptions {
+  filter?: string;
+  pageSize?: number;
+  pageToken?: string;
+}
+
 export interface ModelContent {
   readonly displayName?: string;
   readonly tags?: string[];
@@ -64,6 +71,11 @@ export interface ModelResponse extends ModelContent {
   readonly updateTime: string;
   readonly etag: string;
   readonly modelHash?: string;
+}
+
+export interface ListModelsResponse {
+  readonly models?: ModelResponse[];
+  readonly nextPageToken?: string;
 }
 
 export interface OperationResponse {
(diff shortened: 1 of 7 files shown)


src/machine-learning/machine-learning-api-client.ts
```diff
@@ -140,6 +152,31 @@ export class MachineLearningApiClient {
       });
   }
 
+  public listModels(options: ListModelsOptions = {}): Promise<ListModelsResponse> {
+    if (typeof options.filter !== 'undefined' && !validator.isNonEmptyString(options.filter)) {
```
hiranya911: Best to not rely on default value in the method signature. Otherwise an explicit `null` value may behave unexpectedly.

```
if (tyepeof options !== 'undefined') {
  if (!options.isNonNullObject(options)) {
    // reject
  }
}
```

test/integration/machine-learning.spec.ts
```diff
@@ -324,6 +324,124 @@ describe('admin.machineLearning', () => {
...
+      });
+    });
+
+    it('handles pageTokens properly', () => {
+      return Promise.all([
+        createTemporaryModel({displayName: 'node-integration-list6a'}),
+        createTemporaryModel({displayName: 'node-integration-list6b'}),
+        createTemporaryModel({displayName: 'node-integration-list6c'}),
+      ])
+      .then(([]) => {
```
hiranya911: Just `then(() =>`

src/machine-learning/machine-learning-api-client.ts
```diff
@@ -140,6 +152,31 @@ export class MachineLearningApiClient {
...
+      const err = new FirebaseMachineLearningError('invalid-argument', 'Invalid list filter.');
+      return Promise.reject(err);
+    }
+    if (typeof options.pageSize !== 'undefined' && !validator.isNumber(options.pageSize)) {
```
hiranya911: Put a cap on the pageSize.

test/integration/machine-learning.spec.ts
```diff
@@ -324,6 +324,124 @@ describe('admin.machineLearning', () => {
     });
   });
 
+  describe('listModels()', () => {
```
hiranya911: I'd recommend organizing this test as follows to reduce the rpc overhead of creating/deleting many temporary models.

```
describe('listModels()', () => {
  before(() => {
     // create 3-4 temp models
  });

  // individual it() cases each reusing the models created in before()
 it('resolves with a list of models', () => {
    return admin.machineLearning().listModels(...);
 });
});
```

This has the added benefit that each test case only calls `listModels()` and none of the setup/create methods.

test/unit/machine-learning/machine-learning.spec.ts
```diff
@@ -190,6 +227,26 @@ describe('MachineLearning', () => {
     });
   });
 
+  describe('model', () => {
```
hiranya911: `Model`

src/machine-learning/machine-learning-api-client.ts
```diff
@@ -45,6 +45,13 @@ export interface ModelUpdateOptions extends ModelOptions {
   state?: { published?: boolean; };
 }
 
+/** Interface representing listModels options. */
+export interface ListModelsOptions {
```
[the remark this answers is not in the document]
ifielker: API proposal had ListModelsOptions, I changed everything to match.

src/machine-learning/machine-learning.ts
```diff
@@ -161,8 +154,24 @@ export class MachineLearning implements FirebaseServiceInterface {
    *     token. For the last page, an empty list of models and no page token are
    *     returned.
    */
-  public listModels(options: ListModelsOptions): Promise<ListModelsResult> {
-    throw new Error('NotImplemented');
+  public listModels(options: ListModelsOptions = {}): Promise<ListModelsResult> {
+     return this.client.listModels(options)
+      .then((resp) => {
+        if (resp == null) {
```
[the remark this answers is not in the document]
ifielker: done

src/machine-learning/machine-learning-api-client.ts
ifielker: done (minus the 'undefined' that we discussed)

test/integration/machine-learning.spec.ts
ifielker: done

src/machine-learning/machine-learning-api-client.ts
ifielker: There's a cap on the number of models people are allowed to create, so it automatically functions as a pageSize cap too. Is that sufficient?

hiranya911: We usually put an explicit cap on all list operations. This is documented and the inputs are checked and rejected in the SDK if they exceed the cap. I think we need something similar here. You can use the same underlying cap as the backend API.

test/integration/machine-learning.spec.ts
ifielker: done

test/unit/machine-learning/machine-learning.spec.ts
ifielker: done

test/integration/machine-learning.spec.ts
```diff
@@ -324,6 +322,120 @@ describe('admin.machineLearning', () => {
...
+    let model1: admin.machineLearning.Model;
+    let model2: admin.machineLearning.Model;
+    let model3: admin.machineLearning.Model;
+
+    before(() => {
+      return Promise.all([
```
hiranya911: Loop might be slightly more readable here:

```
const promises = [];
for (let i = 0; i < 3; i++) {
  promises.push(...);
}

return Promise.all(promises);
```

But it's your call.

PR #796, opened 2020-02-28
Adding AutoML support for Firebase ML

ifielker:
Adding AutoML support for Firebase ML

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -5226,6 +5226,7 @@ declare namespace admin.machineLearning {
     readonly sizeBytes: number;
 
     readonly gcsTfliteUri?: string;
+    readonly automlModelId?: string;
   }
 
   /**
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -5205,38 +5205,33 @@ declare namespace admin.messaging {
 
 declare namespace admin.machineLearning {
   /**
-   * Interface representing options for listing Models.
+   * Firebase ML Model input objects
    */
-  interface ListModelsOptions {
-    filter?: string;
-    pageSize?: number;
-    pageToken?: string;
+  interface ModelOptionsBase {
+    displayName?: string;
+    tags?: string[];
   }
-
-  /** Response object for a listModels operation. */
-  interface ListModelsResult {
-    readonly models: Model[];
-    readonly pageToken?: string;
+  interface GcsTfliteModelOptions extends ModelOptionsBase {
+    tfliteModel: {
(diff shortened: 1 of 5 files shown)


src/index.d.ts
```diff
@@ -5226,6 +5226,7 @@ declare namespace admin.machineLearning {
     readonly sizeBytes: number;
 
     readonly gcsTfliteUri?: string;
+    readonly automlModelId?: string;
```
hiranya911: Just noticed that the type name is `TFLiteModel` whereas the field is `gcsTfliteUri`. We probably should be consistent.

src/machine-learning/machine-learning.ts
```diff
@@ -298,7 +303,9 @@ export class Model {
 export interface TFLiteModel {
   readonly sizeBytes: number;
 
-  readonly gcsTfliteUri: string;
+  // Oneof these two
```
hiranya911: Another way to represent this kind of relationships is with union types.

```
interface GcsTFLiteModel {
  sizeBytes: number;
  gcsTFliteUri: string;
}

interface AutoMLModel {
  sizeBytes: number;
  automlModelId: string;
}

type TFLiteModel = GcsTFLiteModel | AutoMLModel
```

You can also take the inheritance approach if you want:

```
interface TFLiteModel {
  sizeBytes: number;
}

interface GcsTFLiteModel extends TFLiteModel {
  gcsTfliteUri: string;
}

interface AutoMLModel extends TFLiteModel {
  autoMlModelId: string;
}
```

Not sure if either of these options pay off in this case. But consider the pros and cons before ruling them out.

src/index.d.ts
```diff
@@ -5205,37 +5205,33 @@ declare namespace admin.messaging {
...
+    tfliteModel: {
+      gcsTfliteUri: string;
+    }
   }
+  interface AutoMLTfliteModelOptions extends ModelOptionsBase {
+    tfliteModel: {
+      automlModelName: string;
+    }
+  }
+  type ModelOptions = ModelOptionsBase | GcsTfliteModelOptions | AutoMLTfliteModelOptions;
```
hiranya911: May be this is a bit cleaner:

```
interface ModelOptionsBase {
  displayName?: string;
  tags?: string[];
  tfliteModel?: GcsTfliteModelOptions | AutoMLTfliteModelOptions;
}

interface GcsTfliteModelOptions {
  gcsTfliteUrl: string;
}
```

PR #798, opened 2020-03-03
Fix compilation error in integration tests

rsgowman:
Introduced by https://github.com/firebase/firebase-admin-node/pull/790


hiranya911:
Seems to have been caused by the removal of the following line:

```
// @ts-ignore: purposely unused method.
```

Turns out it's not a TSLint directive, but a compiler directive. 🤦‍♂

PR #799, opened 2020-03-03
Add getTemplate functionality

lahirumaramba:
- Implement `getTemplate()`
- Introduce new `RemoteConfigParameterValue` type
- To closely represent the backend REST API, change `parameters` in `RemoteConfigTemplate` to a hashmap
- Add unit tests for `getTemplate`

### DIFF
--- a/src/remote-config/remote-config-api-client.ts
+++ b/src/remote-config/remote-config-api-client.ts
@@ -114,12 +114,11 @@ export class RemoteConfigApiClient {
             'invalid-argument',
             'ETag header is not present in the server response.');
         }
-        const remoteConfigResponse: RemoteConfigResponse = {
+        return {
           conditions: resp.data.conditions,
           parameters: resp.data.parameters,
           etag: resp.headers['etag'],
-        }
-        return remoteConfigResponse;
+        };
       })
       .catch((err) => {
         throw this.toFirebaseError(err);
--- a/src/remote-config/remote-config-api-client.ts
+++ b/src/remote-config/remote-config-api-client.ts
@@ -21,7 +21,8 @@ import { FirebaseApp } from '../firebase-app';
 import * as utils from '../utils/index';
 import * as validator from '../utils/validator';
 
-const REMOTE_CONFIG_V1_API = 'https://firebaseremoteconfig.googleapis.com/v1';
+// Remote Config backend constants
+const FIREBASE_REMOTE_CONFIG_V1_API = 'https://firebaseremoteconfig.googleapis.com/v1';
 const FIREBASE_REMOTE_CONFIG_HEADERS = {
   'X-Firebase-Client': 'fire-admin-node/<XXX_SDK_VERSION_XXX>',
   // There is a known issue in which the ETag is not properly returned in cases where the request
@@ -128,7 +129,7 @@ export class RemoteConfigApiClient {
   private getUrl(): Promise<string> {
     return this.getProjectIdPrefix()
       .then((projectIdPrefix) => {
-        return `${REMOTE_CONFIG_V1_API}/${projectIdPrefix}`;
+        return `${FIREBASE_REMOTE_CONFIG_V1_API}/${projectIdPrefix}`;
       });
   }
 
--- a/src/remote-config/remote-config-api-client.ts
+++ b/src/remote-config/remote-config-api-client.ts
@@ -21,7 +21,7 @@ import { FirebaseApp } from '../firebase-app';
 import * as utils from '../utils/index';
 import * as validator from '../utils/validator';
 
-const REMOTECONFIG_V1_API = 'https://firebaseremoteconfig.googleapis.com/v1';
+const REMOTE_CONFIG_V1_API = 'https://firebaseremoteconfig.googleapis.com/v1';
 const FIREBASE_REMOTE_CONFIG_HEADERS = {
   'X-Firebase-Client': 'fire-admin-node/<XXX_SDK_VERSION_XXX>',
   // There is a known issue in which the ETag is not properly returned in cases where the request
@@ -32,7 +32,7 @@ const FIREBASE_REMOTE_CONFIG_HEADERS = {
 };
 
 enum ConditionDisplayColor {
-  CONDITION_DISPLAY_COLOR_UNSPECIFIED = "Unspecified",
+  UNSPECIFIED = "Unspecified",
   BLUE = "Blue",
   BROWN = "Brown",
   CYAN = "Cyan",
@@ -46,22 +46,17 @@ enum ConditionDisplayColor {
   TEAL = "Teal",
 }
 
-interface Version {
-  readonly versionNumber: string;
-  readonly updateTime: string;
-  readonly updateUser: { readonly name?: string; readonly email?: string; readonly imageUrl?: string };
-  readonly updateOrigin: string;
-  readonly updateType: string;
-  readonly description?: string;
-  readonly rollbackSource?: string;
-  readonly isLegacy?: boolean;
+/** Interface representing a Remote Config parameter `value` in value options. */
+export interface ExplicitParameterValue {
+  readonly value: string;
 }
 
-/** Interface representing a Remote Config parameter value. */
-export interface RemoteConfigParameterValue {
-  readonly value?: string;
-  readonly useInAppDefault?: boolean;
+/** Interface representing a Remote Config parameter `useInAppDefault` in value options. */
+export interface InAppDefaultValue {
+  readonly useInAppDefault: boolean;
 }
+
+export type RemoteConfigParameterValue = ExplicitParameterValue | InAppDefaultValue;
 
 /** Interface representing a Remote Config parameter. */
 export interface RemoteConfigParameter {
@@ -79,7 +74,6 @@ interface RemoteConfigCondition {
 export interface RemoteConfigResponse {
   readonly conditions?: RemoteConfigCondition[];
   readonly parameters?: { [key: string]: RemoteConfigParameter };
-  readonly version?: Version; //only undefined when there is no active template in the project
   readonly eTag: string;
 }
 
@@ -117,11 +111,13 @@ export class RemoteConfigApiClient {
         if (!Object.prototype.hasOwnProperty.call(resp.headers, 'etag')) {
           throw new FirebaseRemoteConfigError(
             'invalid-argument',
-            'ETag is unavailable');
+            'ETag header is not present in the server response.');
         }
-        const remoteConfigResponse = resp.data as RemoteConfigResponse;
-        // Include the eTag in RemoteConfigResponse
-        (remoteConfigResponse as any).eTag = resp.headers['etag'];
+        const remoteConfigResponse: RemoteConfigResponse = {
+          conditions: resp.data.conditions,
+          parameters: resp.data.parameters,
+          eTag: resp.headers['etag'],
+        }
         return remoteConfigResponse;
       })
       .catch((err) => {
@@ -132,7 +128,7 @@ export class RemoteConfigApiClient {
   private getUrl(): Promise<string> {
     return this.getProjectIdPrefix()
       .then((projectIdPrefix) => {
-        return `${REMOTECONFIG_V1_API}/${projectIdPrefix}`;
+        return `${REMOTE_CONFIG_V1_API}/${projectIdPrefix}`;
       });
   }
 
@@ -145,7 +141,7 @@ export class RemoteConfigApiClient {
       .then((projectId) => {
         if (!validator.isNonEmptyString(projectId)) {
           throw new FirebaseRemoteConfigError(
-            'invalid-argument',
+            'unknown-error',
             'Failed to determine project ID. Initialize the SDK with service account credentials, or '
             + 'set project ID as an app option. Alternatively, set the GOOGLE_CLOUD_PROJECT '
             + 'environment variable.');
--- a/src/remote-config/remote-config-api-client.ts
+++ b/src/remote-config/remote-config-api-client.ts
@@ -75,7 +75,7 @@ interface RemoteConfigCondition {
 export interface RemoteConfigResponse {
   readonly conditions?: RemoteConfigCondition[];
   readonly parameters?: { [key: string]: RemoteConfigParameter };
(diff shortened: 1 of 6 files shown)


src/remote-config/remote-config.ts
```diff
@@ -56,22 +52,69 @@ class RemoteConfigInternals implements FirebaseServiceInternalsInterface {
...
+  }
 }
 
 /**
  * Remote Config template class.
  */
 export class RemoteConfigTemplate {
 
-  public parameters: RemoteConfigParameter[];
+  public parameters: { [key: string]: RemoteConfigParameter };
```
lahirumaramba: Changing the `parameters` to use a hashmap to better represent the underlying REST API

src/remote-config/remote-config-api-client.ts
```diff
@@ -14,23 +14,184 @@
...
+  private getProjectIdPrefix(): Promise<string> {
+    if (this.projectIdPrefix) {
+      return Promise.resolve(this.projectIdPrefix);
+    }
+
+    return utils.findProjectId(this.app)
+      .then((projectId) => {
+        if (!validator.isNonEmptyString(projectId)) {
+          throw new FirebaseRemoteConfigError(
+            'invalid-argument',
```
hiranya911: unknown-error is probably more appropriate here.

src/remote-config/remote-config.ts
```diff
@@ -56,22 +52,69 @@ class RemoteConfigInternals implements FirebaseServiceInternalsInterface {
...
+    if (!validator.isNonNullObject(config) ||
+      !validator.isNonEmptyString(config.eTag)) {
+      throw new FirebaseRemoteConfigError(
+        'invalid-argument',
+        `Invalid Remote Config template response: ${JSON.stringify(config)}`);
+    }
+
+    this.parameters = {};
+    this.conditions = [];
+    this.eTagInternal = config.eTag;
```
hiranya911: etagInternal

src/remote-config/remote-config.ts
```diff
@@ -56,22 +52,69 @@ class RemoteConfigInternals implements FirebaseServiceInternalsInterface {
...
+
+    if (typeof config.parameters !== 'undefined') {
+      if (!validator.isNonNullObject(config.parameters)) {
+        throw new FirebaseRemoteConfigError(
+          'invalid-argument',
+          `Remote Config parameters must be a non-null object`);
+      }
+      this.parameters = config.parameters;
+    }
```
hiranya911:
```
} else {
  this.parameters = {};
}
```

src/remote-config/remote-config.ts
```diff
@@ -56,22 +52,69 @@ class RemoteConfigInternals implements FirebaseServiceInternalsInterface {
...
+
+    if (typeof config.conditions !== 'undefined') {
+      if (!validator.isArray(config.conditions)) {
+        throw new FirebaseRemoteConfigError(
+          'invalid-argument',
+          `Remote Config conditions must be an array`);
+      }
+      this.conditions = this.parseConditions(config.conditions);
```
hiranya911: config.conditions.map(...)

test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+        .stub(HttpClient.prototype, 'send')
+        .resolves(utils.responseFrom({
+          conditions: [{ name: 'ios', expression: 'exp' }],
+          parameters: { param: { defaultValue: { value: 'true' } } },
+          version: {},
+        }, 200, { etag: 'etag-123456789012-1' }));
+      stubs.push(stub);
+      return apiClient.getTemplate()
+        .then((resp) => {
+          expect(resp.conditions).to.deep.equal([{ name: 'ios', expression: 'exp' }]);
```
hiranya911: You can reduce some of this duplication if the test response was assigned to a variable. Then you can:

```
expect(resp.conditions).to.deep.equal(testResponse.conditions);
```

test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+          expect(resp.parameters).to.deep.equal({ param: { defaultValue: { value: 'true' } } });
```
hiranya911: Assert in etag

test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+          expect(stub).to.have.been.calledOnce.and.calledWith({
+            method: 'GET',
+            url: 'https://firebaseremoteconfig.googleapis.com/v1/projects/test-project/remoteConfig',
+            headers: EXPECTED_HEADERS,
+          });
+        });
+    });
+
+    it('should throw when the etag does not present in the headers', () => {
```
hiranya911: should "reject" when the etag "is" not present

test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+      const stub = sinon
+        .stub(HttpClient.prototype, 'send')
+        .resolves(utils.responseFrom({
+          conditions: [], parameters: {}, version: {},
+        }));
+      stubs.push(stub);
+      const expected = new FirebaseRemoteConfigError('invalid-argument', 'ETag is unavailable');
```
hiranya911: Make this error message a bit more descriptive. `ETag header not present in server response.`

test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+      return apiClient.getTemplate()
+        .should.eventually.be.rejected.and.deep.equal(expected);
+    });
+
+    it('should throw when a full platform error response is received', () => {
```
hiranya911: should reject

Same in other test cases.

test/unit/remote-config/remote-config.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+    }],
+    parameters: {
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      holiday_promo_enabled: {
+        defaultValue: { value: 'true' },
+        conditionalValues: { ios: { useInAppDefault: true } },
+        description: 'this is a promo',
+      },
+    },
+    version: {
```
hiranya911: Not relevant for us iiuc

src/remote-config/remote-config.ts
```diff
@@ -89,7 +132,7 @@ export class RemoteConfigTemplate {
    * @return {RemoteConfigParameter} The Remote Config parameter with the provided key.
    */
   public getParameter(key: string): RemoteConfigParameter | undefined {
-    return this.parameters.find((p) => p.key === key);
+    return this.parameters[key];
```
hiranya911: I think we no longer need this method. Developers can always do `template.parameters[key]`

test/unit/remote-config/remote-config.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+
+          const p2 = template.getParameter(key);
+          if (p2) {
+            expect(p2.defaultValue).deep.equals({ value: 'true' });
+            expect(p2.conditionalValues).deep.equals({ ios: { useInAppDefault: true } });
+            expect(p2.description).equals('this is a promo');
+          }
+
+          const c = template.getCondition('ios');
+          if (c) {
```
hiranya911: `expect(c).to.be.not.undefined`

src/remote-config/remote-config-api-client.ts
```diff
@@ -14,23 +14,184 @@
...
-import { FirebaseRemoteConfigError } from './remote-config-utils';
+import { HttpRequestConfig, HttpClient, HttpError, AuthorizedHttpClient } from '../utils/api-request';
+import { PrefixedFirebaseError } from '../utils/error';
+import { FirebaseRemoteConfigError, RemoteConfigErrorCode } from './remote-config-utils';
 import { FirebaseApp } from '../firebase-app';
+import * as utils from '../utils/index';
 import * as validator from '../utils/validator';
 
+const REMOTECONFIG_V1_API = 'https://firebaseremoteconfig.googleapis.com/v1';
+const FIREBASE_REMOTE_CONFIG_HEADERS = {
```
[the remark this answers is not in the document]
lahirumaramba: Good catch! Thanks!

src/remote-config/remote-config-api-client.ts
```diff
@@ -14,23 +14,184 @@
...
+  readonly updateUser: { readonly name?: string; readonly email?: string; readonly imageUrl?: string };
+  readonly updateOrigin: string;
+  readonly updateType: string;
+  readonly description?: string;
+  readonly rollbackSource?: string;
+  readonly isLegacy?: boolean;
+}
+
+/** Interface representing a Remote Config parameter value. */
+export interface RemoteConfigParameterValue {
```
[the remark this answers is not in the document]
lahirumaramba: I think your suggestion makes more sense. Thanks!
1. Are both value and useInAppDefault ever set together? **No, only one should be set at a given time**
2. Does `useInAppDefualt: false` make any sense? It doesn't. There is also no way to set `useInAppDefualt` to `false` in the UI (Firebase console). If `useInAppDefualt` is set to `false`, it in turn means that the `value` should be set. I will add a validation to cover this case.

src/remote-config/remote-config-api-client.ts
```diff
@@ -14,23 +14,184 @@
...
+  name: string;
+  expression: string;
+  tagColor?: ConditionDisplayColor;
+}
+
+export interface RemoteConfigResponse {
+  readonly conditions?: RemoteConfigCondition[];
+  readonly parameters?: { [key: string]: RemoteConfigParameter };
+  readonly version?: Version; //only undefined when there is no active template in the project
+  readonly eTag: string;
```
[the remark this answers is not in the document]
lahirumaramba: Shouldn't we use camel-case here? Since it is `ETag`: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag

src/remote-config/remote-config-api-client.ts
```diff
@@ -14,23 +14,184 @@
...
+      })
+      .then((resp) => {
+        if (!Object.prototype.hasOwnProperty.call(resp.headers, 'etag')) {
+          throw new FirebaseRemoteConfigError(
+            'invalid-argument',
+            'ETag is unavailable');
+        }
+        const remoteConfigResponse = resp.data as RemoteConfigResponse;
+        // Include the eTag in RemoteConfigResponse
+        (remoteConfigResponse as any).eTag = resp.headers['etag'];
```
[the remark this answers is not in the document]
lahirumaramba: Thanks! Updated.

src/remote-config/remote-config.ts
lahirumaramba: Since we initialize `this.parameters = {}` at line 95, do you think we need this else clause here?

test/unit/remote-config/remote-config.spec.ts
lahirumaramba: It looks like even with the above assert, without the `if` I get the following error `remote-config.spec.ts (211,18): Object is possibly 'undefined'.`

Also, the docs recommend that it is best to asset the target is equal to its expected value, instead to the negation. https://www.chaijs.com/api/bdd/#method_undefined

Should we keep both?

hiranya911:
```
expect(c).to.be.not.undefined;
const cond = c as RemoteConfigCondition;
// other assertions on cond
```

src/remote-config/remote-config-api-client.ts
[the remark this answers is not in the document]
hiranya911: It seems the backend already sends this as `etag`. It's quite common to write this as `etag`. You will find many instances of that in that same page you've referenced. But it's almost never written as `eTag` in my experience.

test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -87,16 +93,13 @@ describe('RemoteConfigApiClient', () => {
...
-        }, 200, { etag: 'etag-123456789012-1' }));
+        .resolves(utils.responseFrom(testResponse, 200, { etag: 'etag-123456789012-1' }));
       stubs.push(stub);
       return apiClient.getTemplate()
         .then((resp) => {
-          expect(resp.conditions).to.deep.equal([{ name: 'ios', expression: 'exp' }]);
-          expect(resp.parameters).to.deep.equal({ param: { defaultValue: { value: 'true' } } });
+          expect(resp.conditions).to.deep.equal(testResponse.conditions);
+          expect(resp.parameters).to.deep.equal(testResponse.parameters);
+          expect(resp.eTag).to.equal('etag-123456789012-1');
```
hiranya911: Add some assertions to check `etag` is readonly.

src/remote-config/remote-config.ts
hiranya911: My suggestion was to move that initialization here. That way we init parameters within the same if-else block.

test/unit/remote-config/remote-config-api-client.spec.ts
lahirumaramba: Added this `read-only` check in `remote-config.spec.ts`

lahirumaramba: https://github.com/firebase/firebase-admin-node/pull/799/files#diff-682cd62db9958cb8633703d346f1ca87R202

src/remote-config/remote-config-api-client.ts
```diff
@@ -14,23 +14,181 @@
...
+        };
+        return this.httpClient.send(request);
+      })
+      .then((resp) => {
+        if (!Object.prototype.hasOwnProperty.call(resp.headers, 'etag')) {
+          throw new FirebaseRemoteConfigError(
+            'invalid-argument',
+            'ETag header is not present in the server response.');
+        }
+        const remoteConfigResponse: RemoteConfigResponse = {
```
hiranya911: Nit: You can collapse this into one statement.

```
return {
  ...
};
```

PR #800, opened 2020-03-03
Build integration tests during CI (and release)

rsgowman:
Note that this won't actually run them.

Additionally, the *unit* tests are also built, impying that we're
building them twice (once during this step, and possibly again when
running the unit tests.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,5 +21,5 @@ jobs:
       run: |
         npm ci
         npm run build
-        npm test
         npm run build:tests
+        npm test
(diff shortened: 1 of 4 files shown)


rsgowman:
Ex CI workflow resulting from the purposely introduced compilation error: https://github.com/firebase/firebase-admin-node/pull/800/checks?check_run_id=483043466

.github/workflows/ci.yml
```diff
@@ -22,3 +22,4 @@ jobs:
         npm ci
         npm run build
         npm test
+        npm run build:tests
```
hiranya911: Nit: Move this before `npm test`

rsgowman: sgtm

PR #802, opened 2020-03-05
Allow enabling of anonymous provider via tenant configuration

rsgowman:
Internal API proposal: http://doc/1ikY2GoQZrwHNJ5mBa---t2o9R-MdjA4np4uW9SJ9dTg#heading=h.x94y9zzc13xn


rsgowman:
> Can you add a reference to the API proposal in the description.

Done.

dgibney:
Looks like this was almost over the line before the holidays hit. Hope this comment may bring it back on someone's radar. Thanks.

rsgowman:
Thanks for the reminder. I'll get this in front of the appropriate people this week, and we'll see if we can finish it up. (It had stalled due to some internal launch processes, but those were cleared up just before the holidays.)

src/auth/index.ts
```diff
@@ -1014,6 +1014,11 @@ export namespace auth {
       passwordRequired?: boolean;
     };
 
+    /**
+     * Whether the anonymous provider is enabled.
+     */
+    anonymousSignInEnabled?: boolean;
```
hiranya911, line 1020: Is this really optional? Since `Tenant` is mainly an output type I'd think this can always be present. In fact it is always present in the `Tenant` implementation class.

test/unit/auth/tenant-manager.spec.ts
```diff
@@ -388,6 +389,7 @@ describe('TenantManager', () => {
         enabled: true,
         passwordRequired: true,
       },
+      anonymousSignInEnabled: true,
```
hiranya911, line 392: Do we have tests for where this is undefined or explicitly false?

hiranya911 (commented):
Looks mostly good. Just a couple of questions.

test/unit/auth/tenant-manager.spec.ts
rsgowman: Yes, though they're in the integration tests. grep test/integration/auth.spec.ts for anonymousSignInEnabled. (Broadly, they start at line 809, though there's an update test on line 1265 that explicitly goes through all possibilities; undefined, true, false.)

hiranya911 (approved):
LGTM

PR #803, opened 2020-03-06
Directly initialize properties in UserMetadata

rsgowman:
Prompted by attempting to enable tsconfig options
'strictPropertyInitialization'.

The properties were being indirectly set via Object.defineProperty.
Typescript isn't able to follow that. Setting the properties directly
resolves it. (And also exposed some minor type errors.)

### DIFF
--- a/src/auth/user-record.ts
+++ b/src/auth/user-record.ts
@@ -66,16 +66,19 @@ export interface CreateRequest extends UpdateRequest {
  * @constructor
  */
 export class UserMetadata {
-  public readonly creationTime: string;
-  public readonly lastSignInTime: string;
+  public readonly creationTime?: string | null;
+  public readonly lastSignInTime?: string | null;
 
   constructor(response: any) {
     // Creation date should always be available but due to some backend bugs there
     // were cases in the past where users did not have creation date properly set.
     // This included legacy Firebase migrating project users and some anonymous users.
     // These bugs have already been addressed since then.
-    utils.addReadonlyGetter(this, 'creationTime', parseDate(response.createdAt));
-    utils.addReadonlyGetter(this, 'lastSignInTime', parseDate(response.lastLoginAt));
+    this.creationTime = parseDate(response.createdAt);
+    utils.enforceReadonly(this, 'creationTime');
+
+    this.lastSignInTime = parseDate(response.lastLoginAt);
+    utils.enforceReadonly(this, 'lastSignInTime');
   }
 
   /** @return {object} The plain object representation of the user's metadata. */
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -491,13 +491,12 @@ declare namespace admin.auth {
     /**
      * The date the user last signed in, formatted as a UTC string.
      */
-    lastSignInTime: string;
+    lastSignInTime?: string | null;
 
     /**
      * The date the user was created, formatted as a UTC string.
-     *
      */
-    creationTime: string;
+    creationTime?: string | null;
 
     /**
      * @return A JSON-serializable representation of this object.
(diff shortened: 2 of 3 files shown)


src/auth/user-record.ts
```diff
@@ -66,16 +66,19 @@ export interface CreateRequest extends UpdateRequest {
  * @constructor
  */
 export class UserMetadata {
-  public readonly creationTime: string;
-  public readonly lastSignInTime: string;
+  public readonly creationTime?: string | null;
```
rsgowman: The `?` isn't strictly necessary as `parseDate` always returns `null` if the input date is undefined. But I think that's more implementation detail-ish, so I've left open the possibility of undef here.

src/index.d.ts
```diff
@@ -491,13 +491,12 @@ declare namespace admin.auth {
     /**
      * The date the user last signed in, formatted as a UTC string.
      */
-    lastSignInTime: string;
+    lastSignInTime?: string | null;
```
rsgowman: nb: technically, a public api change. Arguably breaking since typscript users may have to adjust their code. (Note that we've *always* allowed this to be null... we've just not defined it as such. IOW, I think this is a bugfix more than a change.)

src/auth/user-record.ts
hiranya911: These are not currently defined as optional (both here and in the d.ts file). So may be we can keep it as it was?

src/index.d.ts
hiranya911: I think the `?` is more of a significant change. Can we remove that?

src/auth/user-record.ts
rsgowman: The problem is that these reflect the proto, so when the values come back from the server, there isn't any guarantee that these will actually be populated. (For instance, the lastSignInTime will not be present if the user has never logged in.) As it happens, our implementation will notice that and set lastSignInTime to null. I'm not sure if that happens for all of our objects or not... though we could explicitly set things to null in that situation should there be cases where things are otherwise left out, thereby dropping the '?'.

But not adding the '| null' isn't great since it could legitimately be null (or missing). We just didn't notice before because the property was being set indirectly. We could detect this and use a sentinel value instead (empty string in this case?). To be clear, in some circumstances, we are setting these non-nullable values to null today; typescript just isn't noticing.

In other words, I suspect many instances of this throughout the code base are legitimately optional and/or nullable, but because they've been set indirectly, we haven't noticed.

hiranya911: Marking is nullable is fine (and is the right thing to do, since we explicitly initialize it to null). But marking it optional (`?`) seems wrong to me, since it's never undefined.

rsgowman: Ah, I understand you now. I've removed the '?'.

Note that other classes (that I haven't made this conversion to yet) may not explicitly set to null and instead just not set the property. (Possibly; haven't checked yet.) I'll set the types to reflect what they are (like we've now done here) and we can figure out then if there's an inconsistency issue that needs addressing.

src/index.d.ts
```diff
@@ -491,13 +491,12 @@ declare namespace admin.auth {
     /**
      * The date the user last signed in, formatted as a UTC string.
      */
-    lastSignInTime: string;
+    lastSignInTime: string | null;
```
hiranya911: Nit: Let's also update the doc comment to indicate when these values are null.

rsgowman: Done. Though note that there's not a great explanation as to why creationTime might be null. As an alternative, we could keep it as non-nullable and change the implementation to throw an internal error if this situation occurs.

hiranya911: That sounds like the right approach to me.

rsgowman: Done. A number of unit tests needed to be changed as a result. I think the changes are correct, but definitely will want Bassam to weigh in on this too.

PR #804, opened 2020-03-07
feat(auth): Multi-factor Auth support with SMS for Google Cloud Identity Platform

bojeil-google:
Defines multi-factor auth administrative APIs for Google Cloud Identity Platform.


src/auth/user-record.ts
```diff
@@ -41,6 +42,14 @@ function parseDate(time: any): string | null {
   return null;
 }
 
+interface SecondFactor {
```
hiranya911: The public API uses `UpdateMultiFactorInfoRequest` and `CreateMultiFactorInfoRequest` instead of this. Can we use the same here for consistency?

src/index.d.ts
```diff
@@ -644,12 +686,69 @@ declare namespace admin.auth {
...
+    /**
+     * The multi-factor related properties for the current user, if available.
+     */
+    multiFactor?: {
+
+      /**
+       * List of second factors enrolled with the current user.
+       * Currently only phone second factors are supported.
+       */
+      enrolledFactors: PhoneMultiFactorInfo[];
```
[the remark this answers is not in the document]
hiranya911: Actually changing from `PhoneMultiFactorInfo` to `MultiFactorInfo` won't be a breaking change. So this is fine.

src/auth/user-record.ts
bojeil-google: Done.

src/auth/user-record.ts
```diff
@@ -50,26 +59,232 @@ export interface UpdateRequest {
...
 
 /** Parameters for create user operation */
 export interface CreateRequest extends UpdateRequest {
   uid?: string;
+  multiFactor?: {
+    enrolledFactors: SecondFactor[];
+  };
+}
+
+export interface AuthFactorInfo {
```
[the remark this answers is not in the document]
bojeil-google: Done.

src/index.d.ts
hiranya911: Now that you've changed the other instances in this file, it makes sense to change this to `MultiFactorInfo` too. I think that's safer for users. Otherwise code like the following snippet will break in a future release:

```
const user = await admin.auth().getUser(uid);
if (user.multiFactor) {
  const factor = user.multiFactor.enrolledFactors[0];
  console.log(factor.phoneNumber); // this line will break when we change the type in a future release
}
```

bojeil-google: Done.

PR #806, opened 2020-03-09
Add validate template functionality

lahirumaramba:
- Add `validateTemplate(...)` operation
- Add unit tests for `validateTemplate`

### DIFF
--- a/src/remote-config/remote-config-api-client.ts
+++ b/src/remote-config/remote-config-api-client.ts
@@ -31,16 +31,15 @@ const FIREBASE_REMOTE_CONFIG_GET_HEADERS = {
   // https://firebase.google.com/docs/remote-config/use-config-rest#etag_usage_and_forced_updates
   'Accept-Encoding': 'gzip',
 };
-const FIREBASE_REMOTE_CONFIG_PUT_HEADERS = {
+/*const FIREBASE_REMOTE_CONFIG_PUT_HEADERS = {
   'X-Firebase-Client': 'fire-admin-node/<XXX_SDK_VERSION_XXX>',
   // There is a known issue in which the ETag is not properly returned in cases where the request
   // does not specify a compression type. Currently, it is required to include the header
   // `Accept-Encoding: gzip` or equivalent in all requests.
   // https://firebase.google.com/docs/remote-config/use-config-rest#etag_usage_and_forced_updates
   'Accept-Encoding': 'gzip',
-  'If-Match': '',
-  'content-type': 'application/json; charset=utf-8',
-};
+  //'If-Match': '',
+};*/
 
 export enum RemoteConfigConditionDisplayColor {
   BLUE = "Blue",
@@ -137,12 +136,12 @@ export class RemoteConfigApiClient {
   public validateTemplate(template: RemoteConfigContent): Promise<RemoteConfigContent> {
     return this.getUrl()
       .then((url) => {
-        const headers = FIREBASE_REMOTE_CONFIG_PUT_HEADERS;
-        headers["If-Match"] = template.etag;
+        //const headers = { ...FIREBASE_REMOTE_CONFIG_GET_HEADERS, 'If-Match': template.etag };
+        //headers['If-Match'] = template.etag;
         const request: HttpRequestConfig = {
           method: 'PUT',
           url: `${url}/remoteConfig?validate_only=true`,
-          headers: headers,
+          headers: { ...FIREBASE_REMOTE_CONFIG_GET_HEADERS, 'If-Match': template.etag },
           data: {
             conditions: template.conditions,
             parameters: template.parameters,
--- a/src/remote-config/remote-config-api-client.ts
+++ b/src/remote-config/remote-config-api-client.ts
@@ -80,7 +80,7 @@ interface RemoteConfigCondition {
   tagColor?: RemoteConfigConditionDisplayColor;
 }
 
-export interface RemoteConfigContent {
+export interface RemoteConfigTemplateContent {
   readonly conditions?: RemoteConfigCondition[];
   readonly parameters?: { [key: string]: RemoteConfigParameter };
   readonly etag: string;
@@ -106,7 +106,7 @@ export class RemoteConfigApiClient {
     this.httpClient = new AuthorizedHttpClient(app);
   }
 
-  public getTemplate(): Promise<RemoteConfigContent> {
+  public getTemplate(): Promise<RemoteConfigTemplateContent> {
     return this.getUrl()
       .then((url) => {
         const request: HttpRequestConfig = {
@@ -133,7 +133,7 @@ export class RemoteConfigApiClient {
       });
   }
 
-  public validateTemplate(template: RemoteConfigContent): Promise<RemoteConfigContent> {
+  public validateTemplate(template: RemoteConfigTemplateContent): Promise<RemoteConfigTemplateContent> {
     return this.getUrl()
       .then((url) => {
         //const headers = { ...FIREBASE_REMOTE_CONFIG_GET_HEADERS, 'If-Match': template.etag };
@@ -237,6 +237,6 @@ const ERROR_CODE_MAPPING: { [key: string]: RemoteConfigErrorCode } = {
   OUT_OF_RANGE: 'out-of-range',
   PERMISSION_DENIED: 'permission-denied',
   RESOURCE_EXHAUSTED: 'resource-exhausted',
-  UNAUTHENTICATED: 'authentication-error',
+  UNAUTHENTICATED: 'unauthenticated',
   UNKNOWN: 'unknown-error',
 };
--- a/src/remote-config/remote-config-api-client.ts
+++ b/src/remote-config/remote-config-api-client.ts
@@ -230,8 +230,13 @@ interface Error {
 }
 
 const ERROR_CODE_MAPPING: { [key: string]: RemoteConfigErrorCode } = {
+  ABORTED: 'aborted',
+  ALREADY_EXISTS: `already-exists`,
   INVALID_ARGUMENT: 'invalid-argument',
+  FAILED_PRECONDITION: 'failed-precondition',
   NOT_FOUND: 'not-found',
+  OUT_OF_RANGE: 'out-of-range',
+  PERMISSION_DENIED: 'permission-denied',
   RESOURCE_EXHAUSTED: 'resource-exhausted',
   UNAUTHENTICATED: 'authentication-error',
   UNKNOWN: 'unknown-error',
--- a/src/remote-config/remote-config-api-client.ts
+++ b/src/remote-config/remote-config-api-client.ts
@@ -48,32 +48,34 @@ export enum RemoteConfigConditionDisplayColor {
 
 /** Interface representing a Remote Config parameter `value` in value options. */
 export interface ExplicitParameterValue {
-  readonly value: string;
+  value: string;
 }
 
 /** Interface representing a Remote Config parameter `useInAppDefault` in value options. */
 export interface InAppDefaultValue {
-  readonly useInAppDefault: boolean;
+  useInAppDefault: boolean;
 }
 
 export type RemoteConfigParameterValue = ExplicitParameterValue | InAppDefaultValue;
 
 /** Interface representing a Remote Config parameter. */
 export interface RemoteConfigParameter {
-  readonly defaultValue?: RemoteConfigParameterValue;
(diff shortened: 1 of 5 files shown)


src/remote-config/remote-config-api-client.ts
```diff
@@ -23,17 +23,26 @@ import * as validator from '../utils/validator';
...
 };
+const FIREBASE_REMOTE_CONFIG_PUT_HEADERS = {
+  'X-Firebase-Client': 'fire-admin-node/<XXX_SDK_VERSION_XXX>',
+  // There is a known issue in which the ETag is not properly returned in cases where the request
+  // does not specify a compression type. Currently, it is required to include the header
+  // `Accept-Encoding: gzip` or equivalent in all requests.
+  // https://firebase.google.com/docs/remote-config/use-config-rest#etag_usage_and_forced_updates
+  'Accept-Encoding': 'gzip',
+  'If-Match': '',
+  'content-type': 'application/json; charset=utf-8',
```
hiranya911: You typically don't need this. It will be added by the `HttpClient`.

src/remote-config/remote-config-api-client.ts
```diff
@@ -23,17 +23,26 @@ import * as validator from '../utils/validator';
...
   'Accept-Encoding': 'gzip',
 };
+const FIREBASE_REMOTE_CONFIG_PUT_HEADERS = {
+  'X-Firebase-Client': 'fire-admin-node/<XXX_SDK_VERSION_XXX>',
+  // There is a known issue in which the ETag is not properly returned in cases where the request
+  // does not specify a compression type. Currently, it is required to include the header
+  // `Accept-Encoding: gzip` or equivalent in all requests.
+  // https://firebase.google.com/docs/remote-config/use-config-rest#etag_usage_and_forced_updates
+  'Accept-Encoding': 'gzip',
+  'If-Match': '',
```
hiranya911: Looks unnecessary.

src/remote-config/remote-config-utils.ts
```diff
@@ -17,14 +17,14 @@
 import { PrefixedFirebaseError } from '../utils/error';
 
 export type RemoteConfigErrorCode =
-  'authentication-error'
-  | 'duplicate-key'
+  'aborted'
+  | 'already-exists'
+  | 'authentication-error'
```
hiranya911: Not a OnePlatform error. What are we using this for? We should use `unauthenticated` or `permission-denied` instead.

src/remote-config/remote-config-api-client.ts
```diff
@@ -69,10 +78,10 @@ export interface RemoteConfigParameter {
 interface RemoteConfigCondition {
   name: string;
   expression: string;
-  tagColor?: ConditionDisplayColor;
+  tagColor?: RemoteConfigConditionDisplayColor;
 }
 
-export interface RemoteConfigResponse {
+export interface RemoteConfigContent {
```
hiranya911: RemoteConfigTemplateContent

src/remote-config/remote-config-api-client.ts
```diff
@@ -125,6 +134,42 @@ export class RemoteConfigApiClient {
       });
   }
 
+  public validateTemplate(template: RemoteConfigContent): Promise<RemoteConfigContent> {
+    return this.getUrl()
+      .then((url) => {
+        const headers = FIREBASE_REMOTE_CONFIG_PUT_HEADERS;
+        headers["If-Match"] = template.etag;
```
hiranya911: Mutating a constant

src/remote-config/remote-config-api-client.ts
```diff
@@ -125,6 +134,42 @@ export class RemoteConfigApiClient {
...
+          headers: headers,
+          data: {
+            conditions: template.conditions,
+            parameters: template.parameters,
+          }
+        };
+        return this.httpClient.send(request);
+      })
+      .then((resp) => {
+        if (!Object.prototype.hasOwnProperty.call(resp.headers, 'etag')) {
```
hiranya911: I think just checking `validator.isNonEmptyString(resp.headers['etag'])` is enough here?

src/remote-config/remote-config.ts
```diff
@@ -73,6 +74,30 @@ export class RemoteConfig implements FirebaseServiceInterface {
...
   }
+
+  /**
+   * Validates a Remote Config template.
+   *
+   * @param {RemoteConfigTemplate} template The Remote Config template to be validated.
+   *
+   * @return {Promise<RemoteConfigTemplate>} A Promise that fulfills when a template is validated.
+   */
+  public validateTemplate(template: RemoteConfigTemplate): Promise<RemoteConfigTemplate> {
```
hiranya911: I'm starting to lean towards keeping `RemoteConfigTemplate` a simple JS object. That is:

1. Remove any instance methods exposed on it (toJSON, getCondition).
2. Expose it as a simple interface both in typings and here.

Seems we can avoid a lot of these mapping to-and-from with that approach.

test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -159,4 +190,120 @@ describe('RemoteConfigApiClient', () => {
...
+      const stub = sinon
+        .stub(HttpClient.prototype, 'send')
+        .rejects(utils.errorFrom(ERROR_RESPONSE, 404));
+      stubs.push(stub);
+      const expected = new FirebaseRemoteConfigError('not-found', 'Requested entity not found');
+      return apiClient.validateTemplate(remoteConfigTemplate)
+        .should.eventually.be.rejected.and.deep.equal(expected);
+    });
+
+    it('should reject unknown-error when error code is not present', () => {
```
hiranya911: reject with

test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -159,4 +190,120 @@ describe('RemoteConfigApiClient', () => {
...
+      const stub = sinon
+        .stub(HttpClient.prototype, 'send')
+        .rejects(utils.errorFrom({}, 404));
+      stubs.push(stub);
+      const expected = new FirebaseRemoteConfigError('unknown-error', 'Unknown server error: {}');
+      return apiClient.validateTemplate(remoteConfigTemplate)
+        .should.eventually.be.rejected.and.deep.equal(expected);
+    });
+
+    it('should reject unknown-error for non-json response', () => {
```
hiranya911: Here too

test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -159,4 +190,120 @@ describe('RemoteConfigApiClient', () => {
...
+        .stub(HttpClient.prototype, 'send')
+        .rejects(utils.errorFrom(invalidTemplateError, 400));
+      stubs.push(stub);
+      const expected = new FirebaseRemoteConfigError('failed-precondition',
+        '[VALIDATION_ERROR]: [foo] are not valid condition names. All keys in all conditional value maps must be valid condition names.');
+      return apiClient.validateTemplate(remoteConfigTemplate)
+        .should.eventually.be.rejected.and.deep.equal(expected);
+    });
+
+    it('should reject with failed-precondition when a validation error occurres', () => {
```
hiranya911: Combine with the previous test case, via an array.

```
const errorMessages = [...];
errorMessages.forEach((message) => {
  it(....)
});
```

src/remote-config/remote-config-api-client.ts
lahirumaramba: This was added as we need to set the `etag` at runtime. I was getting the following error without adding the property here.
```
Element implicitly has an 'any' type because expression of type '"If-Match"' can't be used to index type '{ 'X-Firebase-Client': string; 'Accept-Encoding': string; }'.
```

src/remote-config/remote-config-api-client.ts
lahirumaramba: Aren't we just mutating a property here? Not the constant itself?
As an alternative, we could create the headers in-place:
```
headers: {
'X-Firebase-Client': 'fire-admin-node/<XXX_SDK_VERSION_XXX>',
If-Match: etag, 
'Accept-Encoding': 'gzip',
}
```

lahirumaramba: We could also decompose the objects (rest/spread) and use it as below. Any preferences?
```
headers: { ...FIREBASE_REMOTE_CONFIG_GET_HEADERS, 'If-Match': template.etag },
```

src/remote-config/remote-config-utils.ts
lahirumaramba: Good catch! Thanks. Fixed.

test/unit/remote-config/remote-config-api-client.spec.ts
lahirumaramba: Thanks! Done.

src/remote-config/remote-config.ts
lahirumaramba: I agree that using an interface will make things simple. However, having a class enables us to add validations in the constructor and also let us make the etag read-only. What do you think about the following:
- keep `RemoteConfigTemplate` as a class here.
- remove` toJSON` and `getCondition` methods.
- add `interface RemoteConfigTemplate` to `remote-config-api-client.ts` instead of `RemoteConfigTemplateContent`.
The above steps will reduce the mappings, but we will still need to map the conditions (color -> tagColor -> color).

src/remote-config/remote-config-api-client.ts
lahirumaramba: Wanted to differentiate the checks for `property exists` and `is a string`. I think `isNonEmptyString` makes sense, too. Is one preferred over the other?

src/remote-config/remote-config-api-client.ts
hiranya911: Define the type of this object to be `{[key: string]: any}` or similar to avoid the TS error.

src/remote-config/remote-config-api-client.ts
hiranya911: Decomposition looks a little better IMO.

src/remote-config/remote-config-api-client.ts
hiranya911: Given that this shouldn't occur in normal conditions, I prefer the approach that requires less typing (i.e. isNonEmptyString).

src/remote-config/remote-config.ts
hiranya911: You can still have a class, but make it interchangeable with a plain object. This makes it possible to call the put operations by simply:

```
admin.remoteConfig().validateTemplate({
  conditions: ...,
  parameters: ...,
  etag: ...,
});
```

src/remote-config/remote-config-api-client.ts
```diff
@@ -81,7 +71,7 @@ interface RemoteConfigCondition {
   tagColor?: RemoteConfigConditionDisplayColor;
 }
 
-export interface RemoteConfigContent {
+export interface RemoteConfigTemplateContent {
```
hiranya911: If we turn `RemoteConfigTemplate` into a pure JS object, then we can get rid of this type.

```
export interface RemoteConfigTemplate {

}

// Not exported -- For our internal use only
class RemoteConfigTemplateImpl implements RemoteConfigTemplate {
  // class with validation
}

getTemplate() {
  return new RemoteConfigTemplateImpl(resp.data);
}
```

test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -270,40 +261,30 @@ describe('RemoteConfigApiClient', () => {
...
-      };
-      const stub = sinon
-        .stub(HttpClient.prototype, 'send')
-        .rejects(utils.errorFrom(invalidTemplateError, 400));
-      stubs.push(stub);
-      const expected = new FirebaseRemoteConfigError('failed-precondition',
-        '[VERSION_MISMATCH]: Expected version 6, found 8 for project: 123456789012');
-      return apiClient.validateTemplate(remoteConfigTemplate)
-        .should.eventually.be.rejected.and.deep.equal(expected);
+    const errorMessages = [{
```
hiranya911: Just keep a list of errorMessage strings. You can construct the error json in the test case.

src/remote-config/remote-config.ts
lahirumaramba: Added public interface `RemoteConfigTemplate` and an internal implementation `RemoteConfigTemplateImpl`

PR #810, opened 2020-03-11
chore: Adding a .npmrc file to the root of the repo

hiranya911:
Creating the .npmrc file at the root of the repo as instructed in https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow#set-the-token-as-an-environment-variable-on-the-cicd-server

### DIFF
--- a/.github/scripts/publish_package.sh
+++ b/.github/scripts/publish_package.sh
@@ -17,5 +17,7 @@
 set -e
 set -u
 
+echo "//wombat-dressing-room.appspot.com/:_authToken=${NPM_AUTH_TOKEN}" >> .npmrc
+
 readonly UNPREFIXED_VERSION=`echo ${VERSION} | cut -c 2-`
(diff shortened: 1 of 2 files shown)


.github/scripts/publish_package.sh
```diff
@@ -17,7 +17,7 @@
 set -e
 set -u
 
-echo "//wombat-dressing-room.appspot.com/:_authToken=${NPM_AUTH_TOKEN}" >> ~/.npmrc
+echo "//wombat-dressing-room.appspot.com/:_authToken=${NPM_AUTH_TOKEN}" >> .npmrc
```
lahirumaramba: Can assume that this runs in root? should we explicitly say `/. npmrc`?

hiranya911: The file should be added to the root of the repo/project. All the commands are already executed from the root of the repo. So file paths get resolved relative to that directory.

PR #814, opened 2020-03-13
Add publish template operation

lahirumaramba:
- Implement publishTemplate()
- Add force publish support
- Add unit tests for publish template

### DIFF
--- a/src/remote-config/remote-config-api-client.ts
+++ b/src/remote-config/remote-config-api-client.ts
@@ -127,7 +127,7 @@ export class RemoteConfigApiClient {
   }
 
   public validateTemplate(template: RemoteConfigTemplate): Promise<RemoteConfigTemplate> {
-    return this.sendPutRequest(template, template.etag, 'remoteConfig?validate_only=true')
+    return this.sendPutRequest(template, template.etag, true)
       .then((resp) => {
         if (!validator.isNonEmptyString(resp.headers['etag'])) {
           throw new FirebaseRemoteConfigError(
@@ -155,7 +155,7 @@ export class RemoteConfigApiClient {
       // and circumvent the ETag, and the protection from that it provides.
       ifMatch = '*';
     }
-    return this.sendPutRequest(template, ifMatch, 'remoteConfig')
+    return this.sendPutRequest(template, ifMatch)
       .then((resp) => {
         if (!validator.isNonEmptyString(resp.headers['etag'])) {
           throw new FirebaseRemoteConfigError(
@@ -173,12 +173,13 @@ export class RemoteConfigApiClient {
       });
   }
 
-  private sendPutRequest(template: RemoteConfigTemplate, etag: string, path: string): Promise<HttpResponse> {
+  private sendPutRequest(template: RemoteConfigTemplate, etag: string, validateOnly?: boolean): Promise<HttpResponse> {
     if (!validator.isNonEmptyString(etag)) {
       throw new FirebaseRemoteConfigError(
         'invalid-argument',
         'ETag must be a non-empty string.');
     }
+    const path = validateOnly ? 'remoteConfig?validate_only=true' : 'remoteConfig';
     return this.getUrl()
       .then((url) => {
         const request: HttpRequestConfig = {
--- a/src/remote-config/remote-config-api-client.ts
+++ b/src/remote-config/remote-config-api-client.ts
@@ -179,7 +179,10 @@ export class RemoteConfigApiClient {
         'invalid-argument',
         'ETag must be a non-empty string.');
     }
-    const path = validateOnly ? 'remoteConfig?validate_only=true' : 'remoteConfig';
+    let path = 'remoteConfig';
+    if (validateOnly) {
+      path += '?validate_only=true';
+    }
     return this.getUrl()
       .then((url) => {
         const request: HttpRequestConfig = {
--- a/src/remote-config/remote-config-api-client.ts
+++ b/src/remote-config/remote-config-api-client.ts
@@ -160,6 +160,43 @@ export class RemoteConfigApiClient {
       });
   }
 
+  public publishTemplate(template: RemoteConfigTemplate, options?: { force: boolean }): Promise<RemoteConfigTemplate> {
+    let ifMatch: string = template.etag;
+    if (options.force == true) {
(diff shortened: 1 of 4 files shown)


src/remote-config/remote-config-api-client.ts
```diff
@@ -160,6 +160,43 @@ export class RemoteConfigApiClient {
...
+  public publishTemplate(template: RemoteConfigTemplate, options?: { force: boolean }): Promise<RemoteConfigTemplate> {
+    let ifMatch: string = template.etag;
+    if (options && options.force == true) {
+      // setting `If-Match: *` forces the Remote Config template to be updated
+      // and circumvent the ETag, and the protection from that it provides.
+      ifMatch = '*';
+    }
+    return this.getUrl()
+      .then((url) => {
+        const request: HttpRequestConfig = {
```
lahirumaramba: This block is pretty similar to what we use in `validateTemplate`. I considered moving this to a separate function and make the code reusable, but it seemed like an overkill for this. I am open to suggestions though.

test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -280,4 +280,108 @@ describe('RemoteConfigApiClient', () => {
...
+      const stub = sinon
+        .stub(HttpClient.prototype, 'send')
+        .rejects(utils.errorFrom(ERROR_RESPONSE, 404));
+      stubs.push(stub);
+      const expected = new FirebaseRemoteConfigError('not-found', 'Requested entity not found');
+      return apiClient.publishTemplate(REMOTE_CONFIG_TEMPLATE)
+        .should.eventually.be.rejected.and.deep.equal(expected);
+    });
+
+    it('should reject with unknown-error when error code is not present', () => {
```
lahirumaramba: Most of these test cases are repeated for `getTemplate`, `validateTemplate`, and `publishTemplate`. I am open to suggestions on reducing repeated code if needed.

src/remote-config/remote-config-api-client.ts
hiranya911: May be a private `publishTemplate(template, etag)` method makes sense.

src/remote-config/remote-config-api-client.ts
```diff
@@ -160,6 +160,43 @@ export class RemoteConfigApiClient {
       });
   }
 
+  public publishTemplate(template: RemoteConfigTemplate, options?: { force: boolean }): Promise<RemoteConfigTemplate> {
+    let ifMatch: string = template.etag;
```
hiranya911: What if `template.etag` is null or undefined and force is not true?

test/unit/remote-config/remote-config-api-client.spec.ts
hiranya911: Probably not worth the refactor. But try to share constants, helper methods etc.

src/remote-config/remote-config-api-client.ts
lahirumaramba: If force is not true and `template.etag` is undefined or null the developer will see the following `network-error`:
```
{ code: 'app/network-error',
     message:
      'Error while making request: Invalid value "undefined" for header "If-Match". Error code: ERR_HTTP_INVALID_HEADER_VALUE' },
  codePrefix: 'app' }
```

hiranya911: They should see a better error message. We should handle this as an invalid-argument case.

lahirumaramba: For `null` etag we get the following error:
```
{ code: 'remote-config/failed-precondition',
     message:
      '[VERSION_MISMATCH]: The provided ETag [null] is not of the format "etag-<project number>-<version number>".' },
  codePrefix: 'remote-config' }
```

lahirumaramba: I don't mind adding a check for `null` an `undefined` locally if we think that adds any value though

lahirumaramba: Just notice your comment. I will add a check to handle this as an invalid-argument case.

lahirumaramba: Added the check to `sendPutRequest()`

test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -220,13 +221,20 @@ describe('RemoteConfigApiClient', () => {
...
         .should.eventually.be.rejected.and.deep.equal(expected);
     });
 
+    [null, undefined, ''].forEach((etag) => {
+      it('should reject when the etag in template is null, undefined, or an empty string', () => {
+        expect(() => apiClient.validateTemplate({ conditions: [], parameters: {}, etag: etag as any }))
+          .to.throw('ETag must be a non-empty string.');
+      });
+    });
+
```
lahirumaramba: New test case for `null`, `undefined`, and `empty` etags.

src/remote-config/remote-config-api-client.ts
```diff
@@ -160,6 +148,52 @@ export class RemoteConfigApiClient {
...
+          parameters: resp.data.parameters,
+          etag: resp.headers['etag'],
+        };
+      })
+      .catch((err) => {
+        throw this.toFirebaseError(err);
+      });
+  }
+
+  private sendPutRequest(template: RemoteConfigTemplate, etag: string, path: string): Promise<HttpResponse> {
```
hiranya911: Instead of path, pass `validateOnly?: boolean` here.

src/remote-config/remote-config-api-client.ts
```diff
@@ -160,6 +148,53 @@ export class RemoteConfigApiClient {
...
+      });
+  }
+
+  private sendPutRequest(template: RemoteConfigTemplate, etag: string, validateOnly?: boolean): Promise<HttpResponse> {
+    if (!validator.isNonEmptyString(etag)) {
+      throw new FirebaseRemoteConfigError(
+        'invalid-argument',
+        'ETag must be a non-empty string.');
+    }
+    const path = validateOnly ? 'remoteConfig?validate_only=true' : 'remoteConfig';
```
hiranya911:
```suggestion
    let path = 'remoteConfig';
    if (validateOnly) {
      path += '?validate_only=true';
    }
```

Issue #805, opened 2020-03-09
admin.database.Query.on callback wrong params


RANGELJ:
### [REQUIRED] Step 2: Describe your environment

  * Operating System version: macOS Catalina (10.15.3)
  * Firebase SDK version: 8.9.2
  * Firebase Product: database
  * Node.js version: 12.13.0
  * NPM version: 6.14.1

### [REQUIRED] Step 3: Describe the problem
The type definition for admin.database.Query.on
containse a callback that could be undefined:
```
    on(
      eventType: admin.database.EventType,
      callback: (a: admin.database.DataSnapshot | null, b?: string) => any,
      cancelCallbackOrContext?: Object | null,
      context?: Object | null
    ): (a: admin.database.DataSnapshot | null, b?: string) => any;
```

But in the client version of firebase, the type of the a parameter of the callback is always admin.database.DataSnapshot, and I am unable to locate some documentation that states that in the admin sdk, the on callback could pass a null instead of a snapshot

#### Steps to reproduce:

Write this code in a project that has the firebase-admin dependency in a ts file:
```
import admin from 'firebase-admin'

admin.database().ref().on('value', (snapshot: admin.database.DataSnapshot) => {
    console.log(snapshot.val())
})
```

What happened? How can we make the problem occur?
It will show an error when compiling the ts file:
Types of parameters 'snapshot' and 'a' are incompatible.
    Type 'DataSnapshot | null' is not assignable to type 'DataSnapshot'.
      Type 'null' is not assignable to type 'DataSnapshot'

```
import admin from 'firebase-admin'

admin.database().ref().on('value', (snapshot: admin.database.DataSnapshot) => {
    console.log(snapshot.val())
})
```

hiranya911:
@schmidt-sebastian wdyt? Can we just update the typings of the Admin SDK?

schmidt-sebastian:
We fixed them in the JS SDK in https://github.com/firebase/firebase-js-sdk/commit/aca99669dd8ed096f189578c47a56a8644ac62e6 (we reverted the original fix to release this with I/O).

Ideally, we would port all fixes of the index.d.ts file to firebase-admin.

hiranya911:
@schmidt-sebastian can you help us port the necessary changes in our d.ts file?

[...]

PR #830, opened 2020-03-30
Update Database types

schmidt-sebastian:
This syncs the types with https://github.com/firebase/firebase-js-sdk/blob/master/packages/database-types/index.d.ts

Fixes https://github.com/firebase/firebase-admin-node/issues/805

### DIFF
--- a/src/database.d.ts
+++ b/src/database.d.ts
@@ -1272,6 +1272,8 @@ export namespace admin.database {
      * ```
      */
     root: admin.database.Reference;
+    /** @deprecated Removed in next major release to match Web SDK typings. */
+    path: string;
 
     /**
      * Gets a `Reference` for the location at the specified relative path.
--- a/src/database.d.ts
+++ b/src/database.d.ts
@@ -986,8 +986,8 @@ export namespace admin.database {
       */
     on(
       eventType: admin.database.EventType,
-      callback: (a: admin.database.DataSnapshot | null, b?: string) => any,
-      cancelCallbackOrContext?: Object | null,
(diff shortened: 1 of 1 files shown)


src/database.d.ts
```diff
@@ -1272,7 +1272,6 @@ export namespace admin.database {
      * ```
      */
     root: admin.database.Reference;
-    path: string;
```
schmidt-sebastian: Not sure if we should make this change, but this doesn't exist in the Web types.

hiranya911: If `Reference.path` is truly undefined let's remove it. Otherwise, let's call it deprecated for now and remove from a future major version.

schmidt-sebastian: We can make it deprecated. We removed in a major release on Web as well.

schmidt-sebastian, line 1275:
```suggestion
    /** @deprecated Removal in next major release to match Web SDK typings. */
    path: string;
```

puf:
Not sure where else to post this, so putting it here: please indicate how devs are supposed to get the path without this property. See https://stackoverflow.com/q/63425252

src/database.d.ts
Haidar-Hammoud: It makes no sense to remove this. Why require developers to individually create helper functions to fetch the string path from every reference instead of keeping this useful variable?

Every time I deploy my functions I'm greeted by a wall of "path is depreciated" messages... Why should I need to make a function called getPath(reference) that is prone to errors and needed to be referenced throughout my entire codebase when this does exactly what it needs to do?

Consider continuing support for this since the justification to depreciate it is so flimsy.

PR #835, opened 2020-04-02
fix(auth): Fixing UserImportRecord typings declaration

hiranya911:
Typings of the `UserImportRecord` interface had been broken by https://github.com/firebase/firebase-admin-node/commit/564cbe5ffa8766d74c06bada951071efe1b361e9

Further more the `multifactor` configuration was never added to this type. The integration tests use `any` for this type and hence the errors were never surfaced.

This PR addresses all the above.

### DIFF
--- a/src/auth.d.ts
+++ b/src/auth.d.ts
@@ -625,6 +625,59 @@ export namespace admin.auth {
   }
 
   /**
+   * User metadata to include when importing a user.
+   */
+  interface UserMetadataRequest {
+
+    /**
+     * The date the user last signed in, formatted as a UTC string.
+     */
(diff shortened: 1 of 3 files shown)


src/auth.d.ts
```diff
@@ -624,6 +624,59 @@ export namespace admin.auth {
...
+  interface UserMetadataRequest {
+
+    /**
+     * The date the user last signed in, formatted as a UTC string.
+     */
+    lastSignInTime?: string;
+
+    /**
+     * The date the user was created, formatted as a UTC string.
+     *
```
bojeil-google, line 639: Extra line

PR #838, opened 2020-04-03
Convert TagColor enum to a union type to fix integration tests

lahirumaramba:
- Convert TagColor enum to a union type to fix integration tests

### DIFF
--- a/test/integration/remote-config.spec.ts
+++ b/test/integration/remote-config.spec.ts
@@ -40,15 +40,15 @@ const VALID_PARAMETERS = {
   }
 };
 
-const VALID_CONDITIONS = [{
+const VALID_CONDITIONS: admin.remoteConfig.RemoteConfigCondition[] = [{
   name: 'ios',
   expression: 'device.os == \'ios\'',
-  tagColor: 'INDIGO' as admin.remoteConfig.TagColor
+  tagColor: 'INDIGO',
 },
 {
   name: 'andriod',
   expression: 'device.os == \'android\'',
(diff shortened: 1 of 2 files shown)


test/integration/remote-config.spec.ts
```diff
@@ -43,12 +43,12 @@ const VALID_PARAMETERS = {
 const VALID_CONDITIONS = [{
   name: 'ios',
   expression: 'device.os == \'ios\'',
-  tagColor: 'BLUE'
+  tagColor: 'INDIGO' as admin.remoteConfig.TagColor
```
hiranya911, line 46: The cast shouldn't be necessary here.

lahirumaramba: Interesting... I get a typescript compile error if I don't do the cast. Any thoughts?
```
TSError: ⨯ Unable to compile TypeScript
test/integration/remote-config.spec.ts (71,7): Type '{ name: string; expression: string; tagColor: string; }[]' is not assignable to type 'RemoteConfigCondition[]'.
  Type '{ name: string; expression: string; tagColor: string; }' is not assignable to type 'RemoteConfigCondition'.
    Types of property 'tagColor' are incompatible.
      Type 'string' is not assignable to type '"INDIGO" | "GREEN" | "BLUE" | "BROWN" | "CYAN" | "DEEP_ORANGE" | "LIME" | "ORANGE" | "PINK" | "PURPLE" | "TEAL" | undefined'. (2322)
```

lahirumaramba: Fixed by setting the type for `VALID_CONDITIONS`!

PR #840, opened 2020-04-06
Add support for Remote Config parameter groups

lahirumaramba:
- Add support for RC Parameter Groups
- Add unit tests
- Add integration tests

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -835,6 +835,14 @@ declare namespace admin.remoteConfig {
     parameters: { [key: string]: RemoteConfigParameter };
 
     /**
+     * Map of parameter group names to their parameter group objects.
+     * A group's name is mutable but must be unique among groups in the config.
+     * The name is limited to 256 characters and intended to be human-readable. Any Unicode
+     * characters are allowed.
+     */
+    parameterGroups: { [key: string]: RemoteConfigParameterGroup };
+
+    /**
      * ETag of the current Remote Config template (readonly).
      */
     readonly etag: string;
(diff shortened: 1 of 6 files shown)


src/remote-config/remote-config-api-client.ts
```diff
@@ -118,6 +125,7 @@ export class RemoteConfigApiClient {
         return {
           conditions: resp.data.conditions,
           parameters: resp.data.parameters,
+          parameterGroups: resp.data.parameterGroups,
```
hiranya911, line 128: Is this always present on the response?

lahirumaramba: If `parameterGroups` is not present then we assign it to an empty object. Same for `conditions` (empty array) and `parameters`. Good catch for tests! Added more unit tests to cover these. Thanks!

hiranya911: Also run the code coverage tools and see if anything else has slipped through the cracks.

lahirumaramba:
@hiranya911 Thanks! Added more unit tests to cover cases where `conditions`, `parameters`, and `parameterGroups` are not present in the API response.

lahirumaramba:
@egilmorez please review the docs for Remote Config Parameter Groups. Thank you!!

src/index.d.ts
```diff
@@ -834,6 +834,14 @@ declare namespace admin.remoteConfig {
      */
     parameters: { [key: string]: RemoteConfigParameter };
 
+    /**
```
lahirumaramba, line 837: @egilmorez please find the new docs/comments below. Thanks!

test/unit/remote-config/remote-config.spec.ts
```diff
@@ -394,6 +548,51 @@ describe('RemoteConfig', () => {
...
+      return remoteConfig.publishTemplate(REMOTE_CONFIG_TEMPLATE)
+        .then((template) => {
+          expect(template.conditions).deep.equals([]);
+          expect(template.parameters).deep.equals({});
+          // if parameterGroups are not present in the response, we set it to an empty object.
+          expect(template.parameterGroups).deep.equals({});
+        });
+    });
+
+    it('should resolve with conditions:[] when no conditions present in the response', () => {
```
hiranya911, line 581: There's a lot of redundancy in these tests. As a future improvement, think about ways to clean it up a bit.

src/index.d.ts
```diff
@@ -834,6 +834,14 @@ declare namespace admin.remoteConfig {
...
+     * Map of parameter group names to their parameter group objects.
+     * A group's name is mutable but must be unique among groups in the config.
```
egilmorez, line 839: Not sure, but where we say "the config" here, do we mean the Remote Config template?  Or maybe the `RemoteConfig` object, as you mention it in line 891?

lahirumaramba: Good catch. It should be `Remote Config template`. Updated in both places. Thanks!

PR #841, opened 2020-04-09
Update docs for RC publish template options

lahirumaramba:
- Add a more descriptive comment explaining what `force` updates are in `publishTempalte()` `options`.
- Add the new types to `toc.yaml`

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -979,7 +979,13 @@ declare namespace admin.remoteConfig {
      * Publishes a Remote Config template.
      *
      * @param template The Remote Config template to be published.
-     * @param options Optional options object when publishing a Remote Config template.
+     * @param options Optional options object when publishing a Remote Config template:
+     *    - {boolean} `force` Setting this to `true` forces the Remote Config template to 
+     *      be updated and circumvent the ETag. This approach is not recommended 
+     *      because it risks causing the loss of updates to your Remote Config 
+     *      template if multiple clients are updating the Remote Config template.
+     *      See {@link https://firebase.google.com/docs/remote-config/use-config-rest#etag_usage_and_forced_updates 
+     *      ETag usage and forced updates}.
      *
      * @return A Promise that fulfills with the published `RemoteConfigTemplate`.
      */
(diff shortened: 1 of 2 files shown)


src/index.d.ts
```diff
@@ -979,7 +979,13 @@ declare namespace admin.remoteConfig {
      * Publishes a Remote Config template.
      *
      * @param template The Remote Config template to be published.
-     * @param options Optional options object when publishing a Remote Config template.
+     * @param options Optional options object when publishing a Remote Config template:
+     *    - {boolean} `force` Setting this to `true` forces the Remote Config template to 
+     *      be updated and circumvent the ETag. This approach is not recommended 
+     *      because it risks causing the loss of updates to your Remote Config 
+     *      template if multiple clients are updating the Remote Config template.
```
egilmorez, line 986: Do we want to say "simultaneously updating?"  Or does that not really matter?

src/index.d.ts
```diff
@@ -979,7 +979,13 @@ declare namespace admin.remoteConfig {
      * Publishes a Remote Config template.
      *
      * @param template The Remote Config template to be published.
-     * @param options Optional options object when publishing a Remote Config template.
+     * @param options Optional options object when publishing a Remote Config template:
+     *    - {boolean} `force` Setting this to `true` forces the Remote Config template to
```
egilmorez, line 983: Does this render as a bullet list, or just literally as a dash?  

If it's a list, I guess the idea is that the object may later contain more than just one option?

lahirumaramba: This will render as a list and yes, we wanted to keep it open for possible future options. :)
Following is how the list would look like:
image

src/index.d.ts
lahirumaramba: I shamelessly copied it from: https://firebase.google.com/docs/remote-config/use-config-rest#etag_usage_and_forced_updates

I think it doesn't matter if the updates are simultaneous if clients are force publishing templates. I think the concern here is:
- Client A updates the template (not forced)
- Client B now has an outdated template
- Client B force updates the template (without getting the latest changes)
- The project loses updates made by Client A

lahirumaramba:
I forgot to update the `toc.yaml` file for the new types. Updated the PR.
Thank you for your patience! :)

src/index.d.ts
egilmorez: LG, thanks!

PR #848, opened 2020-04-15
Docstring fixes and additions

kevinthecheung:


src/index.d.ts
```diff
@@ -5208,52 +5208,123 @@ declare namespace admin.machineLearning {
...
 
   /**
-   * A TFLite Model output object
+   * A TensorFlow Lite Model output object
    */
   interface TFLiteModel {
+    /** The size of the model. */
     readonly sizeBytes: number;
 
+    /** The URI of the model in Cloud Storage. */
```
ifielker, line 5253: This is true, but there is a bit of subtlety here too. Maybe we should mention something? I mean: The URI is more properly characterized as original source URI. This is the uri that the user supplied for create or update API calls, but once those calls have completed, the model file at that URI location can be modified or deleted without affecting the Firebase model. (We make our own copy of it during create/update).

src/index.d.ts
```diff
@@ -5208,52 +5208,123 @@ declare namespace admin.machineLearning {
...
+    /** The model's tags. */
     readonly tags?: string[];
+
+    /** The timestamp of the model's creation. */
     readonly createTime: string;
+
+    /** The timestamp of the model's most recent update. */
     readonly updateTime: string;
+
     readonly validationError?: string;
```
ifielker, line 5294: Do we need a docstring for this?

src/index.d.ts
```diff
@@ -5208,52 +5208,123 @@ declare namespace admin.machineLearning {
...
+     * Wait for the model to be unlocked.
+     *
+     * @param {number} maxTimeSeconds The maximum time in seconds to wait.
+     *
+     * @return {Promise<void>} A promise that resolves when the model is unlocked
+     *   or the maximum wait time has passed.
+     */
     waitForUnlocked(maxTimeSeconds?: number): Promise<void>;
 
+    /** The model's TensorFlow Lite model file. */
```
ifielker, line 5327: Sort of. It's a TFLiteModel object which contains reference to the original TFLite file and metadata about it (sizeBytes)

Issue #846, opened 2020-04-15
DecodedIDToken interface is missing fields


hixus:
#602 # [REQUIRED] Step 2: Describe your environment

  * Operating System version: OSX
  * Firebase SDK version: 8.10
  * Firebase Product: auth (auth, database, storage, etc)
  * Node.js version: v12.16.1
  * NPM version: 6.13.4

### [REQUIRED] Step 3: Describe the problem

Typescript type of DecodedIdToken seem to miss email and email_verified which are part of at least my decoded id token.

{
    "type": "user",
    "iss": "https://securetoken.google.com/foobar",
    "aud": "foobar",
    "auth_time": 1586957764,
    "user_id": "<userid>",
    "sub": "<userid>",
    "iat": 1586957764,
    "exp": 1586961364,
    "email": "foo@bar.com",
    "email_verified": true,
    "firebase": {
        "identities": {
            "email": [
                "foo@bar.com"
            ]
        },
        "sign_in_provider": "password"
    }
}

#### Steps to reproduce:

console.log the output of admin.auth().verifyIdToken function with valid id token

PR #852, opened 2020-04-17
fix(auth): Defines missing `DecodedIdToken` types.

bojeil-google:
Fixes https://github.com/firebase/firebase-admin-node/issues/846

RELEASE NOTE: Exposed `email`, `email_verified`, `phone_number` and `picture` fields from the `DecodedIdToken` type.

### DIFF
--- a/src/auth.d.ts
+++ b/src/auth.d.ts
@@ -421,6 +421,17 @@ export namespace admin.auth {
     auth_time: number;
 
     /**
+     * The email corresponding to the user who the ID token belonged to, if available.
+     */
+    email?: string;
+
+    /**
+     * Whether the email corresponding to the user who the ID token belonged to is
+     * verified or not, provided the user has an email.
+     */
+    email_verified?: boolean;
+
+    /**
      * The ID token's expiration time, in seconds since the Unix epoch. That is, the
      * time at which this ID token expires and should no longer be considered valid.
      *
@@ -493,6 +504,18 @@ export namespace admin.auth {
      * same project ID specified in the [`aud`](#aud) property.
      */
     iss: string;
+
+    /**
+     * The phone number corresponding to the user who the ID token belonged to,
+     * if available.
+     */
+    phone_number?: string;
+
+    /**
+     * The photo URL corresponding to the user who the ID token belonged to,
+     * if available.
+     */
+    picture?: string;
 
     /**
      * The `uid` corresponding to the user who the ID token belonged to.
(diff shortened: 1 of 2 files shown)


src/auth.d.ts
```diff
@@ -420,6 +420,17 @@ export namespace admin.auth {
      */
     auth_time: number;
 
+    /**
+     * The email corresponding to the user who the ID token belonged to, if available.
```
egilmorez, line 424: Suggest "The email of the user to whom the ID token belonged, if available."

egilmorez, line 424: The email of the user to whom the ID token belonged, if available.

src/auth.d.ts
```diff
@@ -420,6 +420,17 @@ export namespace admin.auth {
...
+     */
+    email?: string;
+
+    /**
+     * Whether the email corresponding to the user who the ID token belonged to is
```
egilmorez, line 429: Suggest "Whether or not the email of the user to whom the ID token belonged is verified, provided the user has an email."

Are the valid values true and false?

src/auth.d.ts
```diff
@@ -494,6 +505,18 @@ export namespace admin.auth {
      */
     iss: string;
 
+    /**
+     * The phone number corresponding to the user who the ID token belonged to,
```
egilmorez, line 509: Suggest "The phone number of the user to whom the ID token belonged, if available."

Or -- can these be present tense "belongs?"  That would be preferable if it is accurate.

src/auth.d.ts
```diff
@@ -494,6 +505,18 @@ export namespace admin.auth {
...
+     * if available.
+     */
+    phone_number?: string;
+
+    /**
+     * The photo URL corresponding to the user who the ID token belonged to,
```
egilmorez, line 515: Suggest "The photo URL for the user to whom the ID token belonged..."

src/auth.d.ts
bojeil-google: Done

src/auth.d.ts
bojeil-google: Yes, true or false. Done.

src/auth.d.ts
bojeil-google: Done on both.

src/auth.d.ts
bojeil-google: done

PR #857, opened 2020-04-18
docs(readme): tag realtime db + add firestore db

rtpHarry:
The docs for the database point to the realtime database. There doesn't seem to be any mention of Firestore in the docs, but looking in the source code, it exists and just passes it through to the `@google-cloud/firestore` package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,8 @@ environments.
 ## Documentation
 
 * [Setup Guide](https://firebase.google.com/docs/admin/setup/)
-* [Database Guide](https://firebase.google.com/docs/database/admin/start/)
+* [Realtime Database Guide](https://firebase.google.com/docs/database/admin/start/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,7 +67,8 @@ environments.
 ## Documentation
 
 * [Setup Guide](https://firebase.google.com/docs/admin/setup/)
-* [Database Guide](https://firebase.google.com/docs/database/admin/start/)
+* [Realtime Database Guide](https://firebase.google.com/docs/database/admin/start/)
+* [Firestore Guide](https://googleapis.dev/nodejs/firestore/latest/index.html)
```
hiranya911, line 71: I think this should point to https://firebase.google.com/docs/firestore

rtpHarry: The reason that I submitted this pr was because I was trying to get started with coding and the api for this library is different to the api used on the main docs. The PR is editable if you want to change the url.

PR #860, opened 2020-04-21
Update toc.yaml

lahirumaramba:
- Add `admin.` prefix to `machinelearning` module title

### DIFF
--- a/docgen/content-sources/node/toc.yaml
+++ b/docgen/content-sources/node/toc.yaml
@@ -132,7 +132,7 @@ toc:
   - title: "InstanceId"
     path: /docs/reference/admin/node/admin.instanceId.InstanceId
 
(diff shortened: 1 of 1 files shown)


lahirumaramba:
Looks like there are a few unused variables in the new machine learning files. I could clean them up as part of the same PR or we should probably so that in a follow up.

docgen/content-sources/node/toc.yaml
```diff
@@ -132,7 +132,7 @@ toc:
   - title: "InstanceId"
     path: /docs/reference/admin/node/admin.instanceId.InstanceId
 
-- title: "machinelearning"
+- title: "admin.machinelearning"
```
hiranya911, line 135: This should be `admin.machineLearning`

Issue #864, opened 2020-04-23
Inconvenient type of UserRecord.customClaims


anantakrishna:
### Environment

  * Operating System version: Windows
  * Firebase SDK version: 8.11.0
  * Firebase Product: auth
  * Node.js version: 10.20.1
  * NPM version: 6.14.4

### The problem

Typing of `UserRecord.customClaims` is `Object`. 
This is inconvenient to use [typescript optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) in this case.

```
const user = await admin.auth().getUser(uid);
if (user.customClaims?.roles?.coordinator)
  …
```

Produces the following error:

```
Property 'roles' does not exist on type 'Object'.ts(2339)
```

This contrasts to the `DecodedIdToken` which has `[key: string]: any;` indexer.

PR #866, opened 2020-04-23
fix(auth): Refines `UserRecord.customClaims` type.

bojeil-google:
Fixes https://github.com/firebase/firebase-admin-node/issues/864

RELEASE NOTE: Updated the typings of the `customClaims` attribute on `UserRecord` and `UserImportRecord` interfaces.


anantakrishna:
Super fast, guys, thank you.

PR #874, opened 2020-05-04
Get remote config template by version number

lahirumaramba:
- Add support to get remote config template by version number

NOTE: API proposal is still under review. Getting a head start

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -994,10 +994,12 @@ declare namespace admin.remoteConfig {
     /**
      * Gets the current active version of the {@link admin.remoteConfig.RemoteConfigTemplate
      * `RemoteConfigTemplate`} of the project.
-     *
+     * 
+     * @param versionNumber Version number of the Remote Config template to look up.
+     *    If not specified, the latest Remote Config template will be returned. Optional.
      * @return A promise that fulfills with a `RemoteConfigTemplate`.
      */
-    getTemplate(): Promise<RemoteConfigTemplate>;
+    getTemplate(versionNumber?: number | string): Promise<RemoteConfigTemplate>;
 
     /**
      * Validates a {@link admin.remoteConfig.RemoteConfigTemplate `RemoteConfigTemplate`}.
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -996,7 +996,7 @@ declare namespace admin.remoteConfig {
      * `RemoteConfigTemplate`} of the project.
      * 
      * @param versionNumber Version number of the Remote Config template to look up.
-     *    If not specified, the latest Remote Config template will be returned. Optional.
+     *    If not specified, the latest Remote Config template is returned. Optional.
      * @return A promise that fulfills with a `RemoteConfigTemplate`.
      */
     getTemplate(versionNumber?: number | string): Promise<RemoteConfigTemplate>;
--- a/src/remote-config/remote-config-api-client.ts
+++ b/src/remote-config/remote-config-api-client.ts
@@ -107,10 +107,10 @@ export class RemoteConfigApiClient {
   }
 
   public getTemplate(versionNumber?: number | string): Promise<RemoteConfigTemplate> {
-    let requestData: any;
+    const requestData: { [k: string]: any } = {};
     if (typeof versionNumber !== 'undefined') {
       this.validateVersionNumber(versionNumber);
-      requestData = { versionNumber: `${versionNumber}` };
+      requestData['versionNumber'] = versionNumber;
     }
     return this.getUrl()
       .then((url) => {
@@ -118,9 +118,7 @@ export class RemoteConfigApiClient {
           method: 'GET',
           url: `${url}/remoteConfig`,
           headers: FIREBASE_REMOTE_CONFIG_HEADERS,
-          ...requestData
-            ? { data: requestData }
-            : {}
+          data: requestData
         };
         return this.httpClient.send(request);
       })
@@ -309,9 +307,7 @@ export class RemoteConfigApiClient {
         'invalid-argument',
         'versionNumber must be a non-empty string in int64 format or a number');
     }
-    const version = validator.isNumber(versionNumber) ?
-      versionNumber : parseFloat(versionNumber as string);
-    if (!Number.isInteger(version as number)) {
+    if (!Number.isInteger(Number(versionNumber))) {
       throw new FirebaseRemoteConfigError(
         'invalid-argument',
         'versionNumber must be an integer or a string in int64 format');
--- a/src/remote-config/remote-config-api-client.ts
+++ b/src/remote-config/remote-config-api-client.ts
@@ -107,16 +107,14 @@ export class RemoteConfigApiClient {
   }
 
   public getTemplate(versionNumber?: number | string): Promise<RemoteConfigTemplate> {
-    let path = 'remoteConfig';
-    if (versionNumber) {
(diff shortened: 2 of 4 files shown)


src/remote-config/remote-config-api-client.ts
```diff
@@ -106,12 +106,16 @@ export class RemoteConfigApiClient {
     this.httpClient = new AuthorizedHttpClient(app);
   }
 
-  public getTemplate(): Promise<RemoteConfigTemplate> {
+  public getTemplate(versionNumber?: number | string): Promise<RemoteConfigTemplate> {
+    let path = 'remoteConfig';
+    if (versionNumber) {
```
lahirumaramba, line 111: If the `versionNumber` is an invalid string (NaN)the backend API will return an error similar to the following:
```
errorInfo:
   { code: 'remote-config/invalid-argument',
     message: 'Invalid value at \'version_number\' (TYPE_INT64), "abc"' },
  codePrefix: 'remote-config' }
```

We could add a check here before making the request, but converting a `string` to a `number` to check `isNaN` will be tricky as the `string` is in `int64 format`. Appreciate any thoughts on this! :)

lahirumaramba: We could also run a RegEx to check if the `string` only contains numbers @hiranya911

src/remote-config/remote-config-api-client.ts
```diff
@@ -106,12 +106,16 @@ export class RemoteConfigApiClient {
...
+      path += '?versionNumber=' + versionNumber;
```
hiranya911, line 112: Use string interpolation.

src/remote-config/remote-config-api-client.ts
hiranya911, line 111: `typeof versionNumber !== 'undefined'`

hiranya911: How about:

```
const version = validators.isNumber(versionNumber) ? versionNumber : parseFloat(versionNumber);
if (!Number.isInteger(version)) {
  throw
}
```

src/remote-config/remote-config-api-client.ts
```diff
@@ -106,12 +106,16 @@ export class RemoteConfigApiClient {
...
+    }
     return this.getUrl()
       .then((url) => {
         const request: HttpRequestConfig = {
           method: 'GET',
-          url: `${url}/remoteConfig`,
+          url: `${url}/${path}`,
```
hiranya911, line 118: Instead of appending to URL, try passing: 

```
data: {versionNumber: `${version}`},
```

src/remote-config/remote-config-api-client.ts
```diff
@@ -303,6 +307,7 @@ const ERROR_CODE_MAPPING: { [key: string]: RemoteConfigErrorCode } = {
   ABORTED: 'aborted',
   ALREADY_EXISTS: `already-exists`,
   INVALID_ARGUMENT: 'invalid-argument',
+  INTERNAL_ERROR: 'internal-error',
```
hiranya911, line 310: INTERNAL

src/remote-config/remote-config-api-client.ts
lahirumaramba: Thanks! I am fine with this. If the string prefixes with numbers and followed by characters, `parseFloat` ignores the characters and parses the number in string.

ex: `70abc` becomes `70`

I believe this is okay, unless we would like further validate with a regex

lahirumaramba:
Updated and added a new test case for version number validation

src/remote-config/remote-config-api-client.ts
```diff
@@ -107,16 +107,20 @@ export class RemoteConfigApiClient {
   }
 
   public getTemplate(versionNumber?: number | string): Promise<RemoteConfigTemplate> {
-    let path = 'remoteConfig';
-    if (versionNumber) {
-      path += '?versionNumber=' + versionNumber;
+    let requestData: any;
```
hiranya911, line 110: `const requestData = {};`

src/remote-config/remote-config-api-client.ts
```diff
@@ -107,16 +107,20 @@ export class RemoteConfigApiClient {
...
     }
     return this.getUrl()
       .then((url) => {
         const request: HttpRequestConfig = {
           method: 'GET',
-          url: `${url}/${path}`,
-          headers: FIREBASE_REMOTE_CONFIG_HEADERS
+          url: `${url}/remoteConfig`,
+          headers: FIREBASE_REMOTE_CONFIG_HEADERS,
+          ...requestData
```
hiranya911, line 121: This is not very readable. Change to `data: requestData` once the above changes are made.

src/remote-config/remote-config-api-client.ts
```diff
@@ -107,16 +107,20 @@ export class RemoteConfigApiClient {
   }
 
   public getTemplate(versionNumber?: number | string): Promise<RemoteConfigTemplate> {
-    let path = 'remoteConfig';
-    if (versionNumber) {
-      path += '?versionNumber=' + versionNumber;
+    let requestData: any;
+    if (typeof versionNumber !== 'undefined') {
+      this.validateVersionNumber(versionNumber);
+      requestData = { versionNumber: `${versionNumber}` };
```
hiranya911, line 113:
```
requestData['versionNumber'] = `${versionNumber}`
```

src/remote-config/remote-config-api-client.ts
```diff
@@ -291,6 +295,28 @@ export class RemoteConfigApiClient {
...
+   */
+  private validateVersionNumber(versionNumber: string | number): void {
+    if (!validator.isNonEmptyString(versionNumber) &&
+      !validator.isNumber(versionNumber)) {
+      throw new FirebaseRemoteConfigError(
+        'invalid-argument',
+        'versionNumber must be a non-empty string in int64 format or a number');
+    }
+    const version = validator.isNumber(versionNumber) ?
+      versionNumber : parseFloat(versionNumber as string);
```
hiranya911, line 313: It looks like `Number(str)` provides stricter parsing.

src/remote-config/remote-config-api-client.ts
lahirumaramba: I used the spread syntax here to avoid sending an empty object for data when no version number is provided. Will update the code to be more readable :)

src/remote-config/remote-config-api-client.ts
lahirumaramba: `Number` is `64 bit double` `(2^53 - 1)` so I was concerned at first as we cannot represent `int 64` `(2^63 - 1)`. But it looks like `Number.isInteger(Number(version))` should work here.

lahirumaramba:
Hi @egilmorez! Please review the docs for the new changes. Thanks!

src/index.d.ts
```diff
@@ -994,10 +994,12 @@ declare namespace admin.remoteConfig {
     /**
      * Gets the current active version of the {@link admin.remoteConfig.RemoteConfigTemplate
      * `RemoteConfigTemplate`} of the project.
-     *
+     * 
+     * @param versionNumber Version number of the Remote Config template to look up.
+     *    If not specified, the latest Remote Config template will be returned. Optional.
```
egilmorez, line 999: Always prefer "is" to "will be" if it works.  It does here!  :)

lahirumaramba: Good catch! Thank you. Updated!

PR #876, opened 2020-05-06
Allow FIREBASE_EMULATOR_CREDENTIALS

samtstern:
See: https://github.com/firebase/firebase-tools/pull/2214

This change will allow the Firebase emulators to pass the credentials from `firebase login` to the Admin SDK as a JSON environment variable.

Currently the emulators rely on picking up `gcloud auth application-default login` credentials on the machine which is at best annoying and at worst means that the emulators are authorized as the wrong user.

### DIFF
--- a/src/auth/credential.ts
+++ b/src/auth/credential.ts
@@ -258,7 +258,7 @@ export class ComputeEngineCredential implements Credential {
  */
 export class RefreshTokenCredential implements Credential {
 
-  private readonly refreshToken: RefreshToken;
+  readonly refreshToken: RefreshToken;
   private readonly httpClient: HttpClient;
 
   /**
@@ -350,6 +350,21 @@ class RefreshToken {
 export function getApplicationDefault(httpAgent?: Agent): Credential {
   if (process.env.GOOGLE_APPLICATION_CREDENTIALS) {
     return credentialFromFile(process.env.GOOGLE_APPLICATION_CREDENTIALS, httpAgent);
+  }
+
+  // Allow the Emulator Suite to pass an authorized_user credential as an environment
+  // variable. This is similar to the gcloud's application default credentials but
+  // not saved to the file system.
+  if (process.env.FIREBASE_EMULATOR_CREDENTIALS) {
+    try {
+      const creds = JSON.parse(process.env.FIREBASE_EMULATOR_CREDENTIALS);
+      return new RefreshTokenCredential(creds, httpAgent, true);
+    } catch (error) {
+      throw new FirebaseAppError(
+        AppErrorCodes.INVALID_CREDENTIAL, 
+        "FIREBASE_EMULATOR_CREDENTIALS set to an invalid JSON string"
+      );
+    }
   }
 
   // It is OK to not have this file. If it is present, it must be valid.
--- a/src/auth/credential.ts
+++ b/src/auth/credential.ts
@@ -358,7 +358,7 @@ export function getApplicationDefault(httpAgent?: Agent): Credential {
   if (process.env.FIREBASE_EMULATOR_CREDENTIALS) {
     try {
       const creds = JSON.parse(process.env.FIREBASE_EMULATOR_CREDENTIALS);
-      return new RefreshTokenCredential(creds, httpAgent, false);
+      return new RefreshTokenCredential(creds, httpAgent, true);
     } catch (error) {
       throw new FirebaseAppError(
         AppErrorCodes.INVALID_CREDENTIAL, 
--- a/src/auth/credential.ts
+++ b/src/auth/credential.ts
@@ -258,7 +258,7 @@ export class ComputeEngineCredential implements Credential {
  */
 export class RefreshTokenCredential implements Credential {
 
-  readonly refreshToken: RefreshToken;
+  private readonly refreshToken: RefreshToken;
   private readonly httpClient: HttpClient;
 
   /**
@@ -358,7 +358,7 @@ export function getApplicationDefault(httpAgent?: Agent): Credential {
   if (process.env.FIREBASE_EMULATOR_CREDENTIALS) {
     try {
       const creds = JSON.parse(process.env.FIREBASE_EMULATOR_CREDENTIALS);
-      return new RefreshTokenCredential(creds, httpAgent, true);
+      return new RefreshTokenCredential(creds, httpAgent, false);
     } catch (error) {
       throw new FirebaseAppError(
         AppErrorCodes.INVALID_CREDENTIAL, 
--- a/test/unit/auth/credential.spec.ts
+++ b/test/unit/auth/credential.spec.ts
@@ -584,7 +584,7 @@ describe('Credential', () => {
       expect(isApplicationDefault(c)).to.be.true;
     });
 
-    describe("FIREBASE_EMULATOR_CREDENTIALS", () => {
+    describe('FIREBASE_EMULATOR_CREDENTIALS', () => {
 
       let oldEnv: any;
 
@@ -596,7 +596,7 @@ describe('Credential', () => {
         process.env = oldEnv;
       });
 
-      it('should return false for RefreshTokenCredential loaded from FIREBASE_EMULATOR_CREDENTIALS', () => {
+      it('should return true for RefreshTokenCredential loaded from FIREBASE_EMULATOR_CREDENTIALS', () => {
         const firebaseEmulatorCred = {
           client_id: 'fakeclientid',
           client_secret: 'fakeclientsecret',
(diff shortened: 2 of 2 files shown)


samtstern:
@hiranya911 open to suggestions of better ways to do this.  We want to pass a credential from the Firebase CLI (and Emulators) into the Admin SDK that comes from `firebase login`.

test/unit/auth/credential.spec.ts
```diff
@@ -440,6 +441,49 @@ describe('Credential', () => {
...
+        refresh_token: "fakerefreshtoken",
+        type: "authorized_user",
+      };
+      process.env.FIREBASE_EMULATOR_CREDENTIALS = JSON.stringify(firebaseEmulatorCred);
+
+      const defaultCred = getApplicationDefault();
+      expect(defaultCred).to.be.an.instanceof(RefreshTokenCredential);
+      expect((defaultCred as RefreshTokenCredential).refreshToken.refreshToken).to.eql(firebaseEmulatorCred.refresh_token);
+    });
+
```
hiranya911, line 486: Also add a test case for `isApplicationDefault()`.

test/unit/auth/credential.spec.ts
```diff
@@ -440,6 +441,49 @@ describe('Credential', () => {
...
+        client_secret: "fakeclientsecret",
+        refresh_token: "fakerefreshtoken",
+        type: "authorized_user",
+      };
+      delete process.env.GOOGLE_APPLICATION_CREDENTIALS;
+      process.env.FIREBASE_EMULATOR_CREDENTIALS = JSON.stringify(firebaseEmulatorCred);
+
+      const defaultCred = getApplicationDefault();
+      expect(defaultCred).to.be.an.instanceof(RefreshTokenCredential);
+      expect((defaultCred as RefreshTokenCredential).refreshToken.refreshToken).to.eql(firebaseEmulatorCred.refresh_token);
```
hiranya911, line 456:
```
expect(defaultCred).to.have.property('refreshToken').that.includes(firebaseEmulatorCred);
```

Then you can don't have to make `refreshToken` public either.

test/unit/auth/credential.spec.ts
```diff
@@ -440,6 +441,49 @@ describe('Credential', () => {
...
     });
 
+    it('should return a RefreshTokenCredential with FIREBASE_EMULATOR_CREDENTIALS', () => {
+      const firebaseEmulatorCred = {
+        client_id: "fakeclientid",
+        client_secret: "fakeclientsecret",
+        refresh_token: "fakerefreshtoken",
+        type: "authorized_user",
+      };
+      delete process.env.GOOGLE_APPLICATION_CREDENTIALS;
```
hiranya911, line 451: Not strictly needed because `FIREBASE_EMULATOR_CREDENTIALS` has precedence?

src/auth/credential.ts
```diff
@@ -352,6 +352,21 @@ export function getApplicationDefault(httpAgent?: Agent): Credential {
     return credentialFromFile(process.env.GOOGLE_APPLICATION_CREDENTIALS, httpAgent);
   }
 
+  // Allow the Emulator Suite to pass an authorized_user credential as an environment
+  // variable. This is similar to the gcloud's application default credentials but
+  // not saved to the file system.
+  if (process.env.FIREBASE_EMULATOR_CREDENTIALS) {
+    try {
+      const creds = JSON.parse(process.env.FIREBASE_EMULATOR_CREDENTIALS);
```
hiranya911, line 360: Do we also need to support specifying a path to a json file instead of the json string itself?

samtstern: No not yet, there is no such file it's something we construct at runtime inside th Functions emulator.

samtstern:
@hiranya911 to answer your questions

> Will the emulator tooling take care of setting the new environment variable or is that something developers ought to do?

This is something that will only be done inside the emulator, we won't document this for external developers.

> Also how would this generalize to other languages where we don't have access to the credentials implementation (or is this only an issue for Functions Emulator)?

This will be Node-only since it's just to make the emulator suite work better.

test/unit/auth/credential.spec.ts
samtstern: Oh thanks!

test/unit/auth/credential.spec.ts
samtstern: Done.  I made `implicit` false but I wasn't sure if that was the correct choice.  Does that seem right to you?

test/unit/auth/credential.spec.ts
```diff
@@ -563,6 +572,20 @@ describe('Credential', () => {
       expect(isApplicationDefault(c)).to.be.true;
     });
 
+    it('should return false for RefreshTokenCredential loaded from FIREBASE_EMULATOR_CREDENTIALS', () => {
```
hiranya911: I think this ought to return true (which means setting implicit=true in the ctor). Otherwise calls like `admin.firestore()` will fail with an invalid-credential error. See https://github.com/firebase/firebase-admin-node/blob/d98560260b66435255e928778c7b5845926204d7/src/firestore/firestore.ts#L66-L102

Also note that Firestore and Storage run their own credential discovery routine, which might fallback to gcloud credentials. Is that the behavior we want? Unfortunately those packages don't support passing explicit refresh token credentials.

test/unit/auth/credential.spec.ts
```diff
@@ -440,6 +440,56 @@ describe('Credential', () => {
       expect((getApplicationDefault())).to.be.an.instanceof(RefreshTokenCredential);
     });
 
+    it('should return a RefreshTokenCredential with FIREBASE_EMULATOR_CREDENTIALS', () => {
+      const firebaseEmulatorCred = {
+        client_id: "fakeclientid",
```
hiranya911, line 445: Nit: Use single-quotes for string literal. Throughout the PR.

test/unit/auth/credential.spec.ts
```diff
@@ -440,6 +440,56 @@ describe('Credential', () => {
...
+      expect(defaultCred).to.be.an.instanceof(RefreshTokenCredential);
+      expect(defaultCred).to.have.property('refreshToken').that.includes({ 
+        clientId: firebaseEmulatorCred.client_id,
+        clientSecret: firebaseEmulatorCred.client_secret,
+        refreshToken: firebaseEmulatorCred.refresh_token,
+      });
+    });
+
+    it('should throw if FIREBASE_EMULATOR_CREDENTIALS is invalid', () => {
+      process.env.FIREBASE_EMULATOR_CREDENTIALS = "{{}}";
```
hiranya911, line 462: I believe you need that `after()` block to unset this. Otherwise all subsequent tests see this invalid value and fail.

test/unit/auth/credential.spec.ts
```diff
@@ -519,6 +569,36 @@ describe('Credential', () => {
       expect(isApplicationDefault(c)).to.be.true;
     });
 
+    describe("FIREBASE_EMULATOR_CREDENTIALS", () => {
+
+      let googleCreds: any = undefined;
+
+      before(() => {
+        googleCreds = process.env.GOOGLE_APPLICATION_CREDENTIALS;
+        delete process.env.GOOGLE_APPLICATION_CREDENTIALS;
```
hiranya911, line 578: I think just unsetting this variable within the test case should be enough. We used to do this trick of restoring the env variables in the past, but that's not really needed for our tests.

samtstern: My concern is that if a test fails before we get to the line where it unsets, it can cause downstream failures.

test/unit/auth/credential.spec.ts
samtstern: So I have tested this with the Firestore SDK and it works ... maybe that's only the case for Storage?  I thought all of those SDKs shared credential discovery.

samtstern:
@hiranya911 thanks for the review, seems like we might need to talk this one through over GVC at some point as I really don't want to mess anything up in something as central as `credential`.

test/unit/auth/credential.spec.ts
hiranya911: Unset at the beginning of the test case, which is what `before()` would do anyway. Just don't worry about restoring the old value.

test/unit/auth/credential.spec.ts
hiranya911: `@google-cloud/*` libraries have their own credentials implementation and lookup. We can pass a service account, but not a refresh token. So in this case they will silently fallback to using the local gcloud sdk credentials.

samtstern:
After meeting with @hiranya911 going to pursue a different route.

PR #877, opened 2020-05-06
Remove type aliases from toc.yaml

lahirumaramba:
- Remove type aliases from toc.yaml

Type doc complaints about missing files when generating docs. This PR removes the type aliases in `toc.yaml` causing the missing file paths.


lahirumaramba:
Type aliases will still show up on the namespace root page:
image

PR #820, opened 2020-03-16
Fixing Android notification options descriptions

egilmorez:
Per internal bug report 149696213.

Staged internally at https://firebase.devsite.corp.google.com/docs/reference/admin/node/admin.messaging.NotificationMessagePayload#sound

Thanks!

PR #879, opened 2020-05-07
Rectified issues in comments for NotificationMessagePayload

VPanjeta:
The comments regarding the `NotificationMessagePayload` were not correct. They were placed with different options.  
The comment related to `tag` was attached to `sound` and `tag` did not have any comment attached.  

Had an issue understanding the payload the first time I opened the installed directory to go through the payload of `sendToDevice`

### DIFF
--- a/src/messaging.d.ts
+++ b/src/messaging.d.ts
@@ -684,8 +684,12 @@ export namespace admin.messaging {
     color?: string;
 
     /**
-     * The sound to be played when the device receives the
-     * notification.
+     * The sound to be played when device receives a notification. Supports
+     * "default" for default notification sounf of device or the filename of a 
+     * sound resource bundled in the app. 
+     * Sound files must reside in `/res/raw/`.
+     * 
+     * **Platforms:** Android
      */
     sound?: string;
 
--- a/src/messaging.d.ts
+++ b/src/messaging.d.ts
@@ -685,7 +685,7 @@ export namespace admin.messaging {
 
     /**
      * The sound to be played when device receives a notification. Supports
-     * "default" for default notification sounf of device or the filename of a 
+     * "default" for default notification sound of device or the filename of a 
      * sound resource bundled in the app. 
      * Sound files must reside in `/res/raw/`.
      * 
--- a/src/messaging.d.ts
+++ b/src/messaging.d.ts
@@ -632,6 +632,17 @@ export namespace admin.messaging {
    * for code samples and detailed documentation.
    */
   interface NotificationMessagePayload {
+    
+    /**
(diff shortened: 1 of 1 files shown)


VPanjeta:
@googlebot I signed it!

lahirumaramba:
Thank you @VPanjeta !

Looks like this was previously addressed in #820, but was broken by #751 right after.

src/messaging.d.ts
```diff
@@ -673,14 +684,8 @@ export namespace admin.messaging {
...
-     * Identifier used to replace existing notifications in the notification drawer.
-     *
-     * If not specified, each request creates a new notification.
-     *
-     * If specified and a notification with the same tag is already being shown,
-     * the new notification replaces the existing one in the notification drawer.
-     *
-     * **Platforms:** Android
+     * The sound to be played when the device receives the
+     * notification.
```
lahirumaramba, line 688:
```
/**
     * The sound to play when the device receives the notification. Supports
     * "default" or the filename of a sound resource bundled in the app. Sound files
     * must reside in `/res/raw/`.
     * 
     * **Platforms:** Android
     */
```
I think the docs from #820 might be more suitable here. I will leave @egilmorez to confirm.

VPanjeta:
Ok since it is already present in #820  . Closing this for now.

lahirumaramba:
It was addressed in #820 but was broken again in #751. Therefore we could still use this PR.
Please update the comments (docs) to the suggested text in my review and we should be good to merge this. Thanks!

VPanjeta:
@lahirumaramba Made the changes as expected. Do let me know if anything else is needed.

src/messaging.d.ts
```diff
@@ -673,13 +684,11 @@ export namespace admin.messaging {
...
 
     /**
-     * Identifier used to replace existing notifications in the notification drawer.
-     *
-     * If not specified, each request creates a new notification.
-     *
-     * If specified and a notification with the same tag is already being shown,
-     * the new notification replaces the existing one in the notification drawer.
-     *
+     * The sound to be played when device receives a notification. Supports
```
egilmorez, line 687: Since we are using articles everywhere else (instead of the extra-concise style that omits them), we should probably use them here too:

"The sound to be played when the device receives a notification. Supports
     *	     * "default" for the default notification sound of the device"

VPanjeta:
@egilmorez Sorry for having some semantic/grammatical errors. English isn't my native language. Pushed a commit after adding articles to the comments. Hopefully that fixes your concerns.  
Do let me know if I'm still missing something.

PR #885, opened 2020-05-11
Add Remote Config Rollback operation

lahirumaramba:
- Add `rollback` operation to the RC API
- Add unit tests

Note: The API proposal is still under review. Getting a head start here.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1026,12 +1026,12 @@ declare namespace admin.remoteConfig {
     publishTemplate(template: RemoteConfigTemplate, options?: { force: boolean }): Promise<RemoteConfigTemplate>;
 
     /**
-    * Rollbacks a project's published Remote Config template to the specified version.
+    * Rolls back a project's published Remote Config template to the specified version.
     * A rollback is equivalent to getting a previously published Remote Config
-    * template, and re-publishing it using a force update.
+    * template and re-publishing it using a force update.
     * 
     * @param versionNumber The version number of the Remote Config template to rollback to.
-    *    The specified version number must be less than the current version number, and not have
+    *    The specified version number must be lower than the current version number, and not have
     *    been deleted due to staleness.
     * @return A promise that fulfills with the published `RemoteConfigTemplate`.
     */
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1026,6 +1026,18 @@ declare namespace admin.remoteConfig {
     publishTemplate(template: RemoteConfigTemplate, options?: { force: boolean }): Promise<RemoteConfigTemplate>;
 
     /**
+    * Rollbacks a project's published Remote Config template to the one specified by the provided
+    * version number. A rollback is equivalent to getting a previously published Remote Config
+    * template, and re-publishing it using a force update.
+    * 
+    * @param versionNumber The version number of the Remote Config template to rollback to.
+    *    The specified version number must be less than the current version number, and not have
+    *    been deleted due to staleness.
+    * @return A promise that fulfills with the published `RemoteConfigTemplate`.
+    */
+    rollback(versionNumber: string | number): Promise<RemoteConfigTemplate>;
+
+    /**
      * Creates and returns a new Remote Config template from a JSON string.
      *
      * @param json The JSON string to populate a Remote Config template.
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1026,8 +1026,8 @@ declare namespace admin.remoteConfig {
     publishTemplate(template: RemoteConfigTemplate, options?: { force: boolean }): Promise<RemoteConfigTemplate>;
 
     /**
-    * Rollbacks a project's published Remote Config template to the one specified by the provided
-    * version number. A rollback is equivalent to getting a previously published Remote Config
+    * Rollbacks a project's published Remote Config template to the specified version.
+    * A rollback is equivalent to getting a previously published Remote Config
     * template, and re-publishing it using a force update.
     * 
     * @param versionNumber The version number of the Remote Config template to rollback to.
--- a/src/remote-config/remote-config-api-client.ts
+++ b/src/remote-config/remote-config-api-client.ts
@@ -123,11 +123,7 @@ export class RemoteConfigApiClient {
         return this.httpClient.send(request);
       })
       .then((resp) => {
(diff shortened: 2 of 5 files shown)


src/index.d.ts
```diff
@@ -1025,6 +1025,18 @@ declare namespace admin.remoteConfig {
      */
     publishTemplate(template: RemoteConfigTemplate, options?: { force: boolean }): Promise<RemoteConfigTemplate>;
 
+    /**
+    * Rollbacks a project's published Remote Config template to the one specified by the provided
```
hiranya911, line 1029: "... to the specified version."

src/remote-config/remote-config-api-client.ts
```diff
@@ -191,6 +191,36 @@ export class RemoteConfigApiClient {
...
+        const request: HttpRequestConfig = {
+          method: 'POST',
+          url: `${url}/remoteConfig:rollback`,
+          headers: FIREBASE_REMOTE_CONFIG_HEADERS,
+          data: { versionNumber: versionNumber }
+        };
+        return this.httpClient.send(request);
+      })
+      .then((resp) => {
+        if (!validator.isNonEmptyString(resp.headers['etag'])) {
```
hiranya911, line 207: I'd assume this logic is used by at least one other method (getTemplate). Extract into a helper method?

lahirumaramba: Good catch! Moved to a helper method. Thanks!

lahirumaramba:
Hi @egilmorez ! Please review the docstrings in `index.d.ts`. Thank you!

src/index.d.ts
```diff
@@ -1025,6 +1025,18 @@ declare namespace admin.remoteConfig {
      */
     publishTemplate(template: RemoteConfigTemplate, options?: { force: boolean }): Promise<RemoteConfigTemplate>;
 
+    /**
+    * Rollbacks a project's published Remote Config template to the specified version.
```
egilmorez, line 1029: Despite the noun "rollback" we still would way "Rolls back a project's . . . "

src/index.d.ts
```diff
@@ -1025,6 +1025,18 @@ declare namespace admin.remoteConfig {
...
+    * A rollback is equivalent to getting a previously published Remote Config
+    * template, and re-publishing it using a force update.
```
egilmorez, line 1031: Suggest omitting the comma here.

src/index.d.ts
```diff
@@ -1025,6 +1025,18 @@ declare namespace admin.remoteConfig {
...
+    * 
+    * @param versionNumber The version number of the Remote Config template to rollback to.
+    *    The specified version number must be less than the current version number, and not have
```
egilmorez, line 1034: Suggest "lower than"

lahirumaramba:
Thanks @egilmorez ! Updated.

src/remote-config/remote-config-api-client.ts
```diff
@@ -174,11 +166,33 @@ export class RemoteConfigApiClient {
     }
     return this.sendPutRequest(template, ifMatch)
       .then((resp) => {
-        if (!validator.isNonEmptyString(resp.headers['etag'])) {
-          throw new FirebaseRemoteConfigError(
-            'invalid-argument',
-            'ETag header is not present in the server response.');
-        }
+        this.validateResponseEtag(resp.headers['etag']);
+        return {
```
hiranya911, line 170: This can also go into a util right? Something like:

```
return this.createRemoteConfigTemplate(resp);
```

The util can do the validation and then extract the right properties from the response.

lahirumaramba: Great point! Moved the creation to a new util. Thanks!

src/remote-config/remote-config-api-client.ts
```diff
@@ -274,6 +251,24 @@ export class RemoteConfigApiClient {
...
+   * @param {string} customEtag A custom etag to replace the etag fom the API response (Optional).
+   */
+  private toRemoteConfigTemplate(resp: HttpResponse, customEtag?: string): RemoteConfigTemplate {
+    const etag = (typeof customEtag == 'undefined') ? resp.headers['etag'] : customEtag;
+    this.validateResponseEtag(etag);
+    return {
+      conditions: resp.data.conditions,
+      parameters: resp.data.parameters,
+      parameterGroups: resp.data.parameterGroups,
+      etag: etag,
```
hiranya911, line 268: Nit: You can just do `etag,` here.

src/remote-config/remote-config-api-client.ts
```diff
@@ -140,16 +134,11 @@ export class RemoteConfigApiClient {
     this.validateRemoteConfigTemplate(template);
     return this.sendPutRequest(template, template.etag, true)
       .then((resp) => {
+        // validating a template returns an etag with the suffix -0 means that your update
```
hiranya911, line 137: Do we ever check for this `-0` suffix?

src/remote-config/remote-config-api-client.ts
```diff
@@ -274,6 +251,24 @@ export class RemoteConfigApiClient {
...
+  /**
+   * Creates a RemoteConfigTemplate from the API response.
+   * If provided, customEtag is used instead of the etag returned in the API response.
+   *
+   * @param {HttpResponse} resp API response object.
+   * @param {string} customEtag A custom etag to replace the etag fom the API response (Optional).
+   */
+  private toRemoteConfigTemplate(resp: HttpResponse, customEtag?: string): RemoteConfigTemplate {
+    const etag = (typeof customEtag == 'undefined') ? resp.headers['etag'] : customEtag;
+    this.validateResponseEtag(etag);
```
hiranya911, line 263: Nit: This should be called something like `validateEtag()` now.

src/remote-config/remote-config-api-client.ts
lahirumaramba: Thanks!

src/remote-config/remote-config-api-client.ts
lahirumaramba: We don't necessarily check for `-0` suffix here. If the response is not an error (200 OK) and contains a valid etag, we can safely assume that the template was successfully validated. All other cases should return an error code. I will add an additional check for the `-0` suffix in a separate PR for good measure :)

PR #896, opened 2020-05-20
Remote Config Add list versions operation

lahirumaramba:
- Add list versions operation
- Add unit tests

NOTE: Integration tests for all the new operations will be added in a separate PR.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -983,6 +983,121 @@ declare namespace admin.remoteConfig {
   type RemoteConfigParameterValue = ExplicitParameterValue | InAppDefaultValue;
 
   /**
+   * Interface representing a Remote Config template version.
+   * Contains all metadata about a particular version of the Remote Config template.
+   * All fields are set at the time the specified Remote Config template was written.
+   */
+  export interface Version {
+    /**
+     * The version number of the version's corresponding Remote Config template. Output only.
+     */
+    versionNumber?: string | number;
+    
+    /**
+     * The Remote Config template creation timestamp in RFC3339 UTC "Zulu" format,
+     * accurate to nanoseconds. Output only.
+     */
+    updateTime?: string;
+
+    /**
+     * The source of origin of the template update action. Output only.
+     */
+    updateOrigin?: ('REMOTE_CONFIG_UPDATE_ORIGIN_UNSPECIFIED' | 'CONSOLE' |
+      'REST_API' | 'ADMIN_SDK_NODE');
+
+    /**
+     * The type of the template update action. Output only.
+     */
+    updateType?: ('REMOTE_CONFIG_UPDATE_TYPE_UNSPECIFIED' |
+      'INCREMENTAL_UPDATE' | 'FORCED_UPDATE' | 'ROLLBACK');
+
+    /**
+     * Aggregation of all metadata fields about the account that performed the update. Output only.
+     */
+    updateUser?: RemoteConfigUser;
+
+    /**
+     * The user-provided description of the corresponding Remote Config template. Optional.
+     */
+    description?: string;
+
+    /**
+     * Only present if this version is the result of a rollback, and will be the version number of
+     * the Remote Config template that was rolled-back to. Output only.
+     */
+    rollbackSource?: string;
+    
+    /**
+     * Indicates weather this Remote Config template was published before version history was
+     * supported. Output only.
+     */
+    isLegacy?: boolean;
+  }
+
+  /** Interface representing a Remote Config list version result. */
+  export interface ListVersionsResult {
+    /**
+     * A list of version metadata objects, sorted in reverse chronological order.
+     */
+    versions: Version[];
+
+    /**
+     * Token to retrieve the next page of results, or empty if there are no more results
+     * in the list.
+     */
+    nextPageToken?: string;
+  }
+
+  /** Interface representing a Remote Config list version options. */
+  export interface ListVersionsOptions {
+    /**
+     * The maximum number of items to return per page.
+     */
+    pageSize?: number;
+
+    /**
+     * The `nextPageToken` value returned from a previous list versions request, if any.
+     */
+    pageToken?: string;
+    
+    /** 
+     * Specify the newest version number to include in the results.
+     * If specified, must be greater than zero. Defaults to the newest version.
+     */
+    endVersionNumber?: string | number;
+
+    /**
+     * Specify the earliest update time to include in the results. Any entries updated before this
+     * time are omitted.
+     */
+    startTime?: Date;
+
+    /**
+     * Specify the latest update time to include in the results. Any entries updated on or after
+     * this time are omitted.
+     */
+    endTime?: Date;
+  }
+
+  /** Interface representing a Remote Config user.*/
+  export interface RemoteConfigUser {
+    /**
+     * Email address. Output only.
+     */
+    email: string;
+
+    /**
+     * Display name. Output only.
+     */
+    name?: string;
+
+    /**
+     * Image URL. Output only.
+     */
+    imageUrl?: string;
+  }
+
+  /**
    * The Firebase `RemoteConfig` service interface.
    *
    * Do not call this constructor directly. Instead, use
@@ -1036,6 +1151,26 @@ declare namespace admin.remoteConfig {
     * @return A promise that fulfills with the published `RemoteConfigTemplate`.
     */
     rollback(versionNumber: string | number): Promise<RemoteConfigTemplate>;
+
+    /**
+    * Gets a list of Remote Config template versions that have been published, sorted in reverse 
+    * chronological order. Only the last 300 versions are stored.
+    * All versions that correspond to non-active Remote Config templates (i.e., all except the 
+    * template that is being fetched by clients) are also deleted if they are older than 90 days.
+    * 
+    * @param options Optional options object for getting a list of tempalte versions:
+    *    - {number} `pageSize` The maximum number of items to return per page.
+    *    - {string} `pageToken` The `nextPageToken` value returned from a previous list versions
+    *      request, if any.
+    *    - {string | number} `endVersionNumber` Specify the newest version number to include in the
+    *      results. If specified, must be greater than zero. Defaults to the newest version.
+    *    - {Date} `startTime` Specify the earliest update time to include in the results. Any
+    *      entries updated before this time are omitted.
+    *    - {Date} `endTime` Specify the latest update time to include in the results. Any entries
+    *      updated on or after this time are omitted.
+    * @return A promise that fulfills with a `ListVersionsResult`.
+    */
+    listVersions(options?: ListVersionsOptions): Promise<ListVersionsResult>;
 
     /**
      * Creates and returns a new Remote Config template from a JSON string.
@@ -1122,7 +1257,7 @@ declare namespace admin.machineLearning {
      *
      * Example: `tfliteModel: { gcsTfliteUri: 'gs://your-bucket/your-model.tflite' }`
      */
-    tfliteModel?: {gcsTfliteUri: string};
+    tfliteModel?: { gcsTfliteUri: string };
   }
 
   /**
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -996,8 +996,8 @@ declare namespace admin.remoteConfig {
     versionNumber?: string;
 
     /**
-     * The timestamp of when the Remote Config template was written to the Remote Config
-     * server.
+     * The timestamp of when this version of the Remote Config template was written to the
+     * Remote Config server.
      */
     updateTime?: string;
 
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -996,7 +996,8 @@ declare namespace admin.remoteConfig {
     versionNumber?: string;
 
     /**
-     * The Remote Config template creation timestamp in UTC.
+     * The timestamp of when the Remote Config template was written to the Remote Config
+     * server.
      */
     updateTime?: string;
 
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -984,57 +984,58 @@ declare namespace admin.remoteConfig {
 
   /**
    * Interface representing a Remote Config template version.
-   * Contains all metadata about a particular version of the Remote Config template.
-   * All fields are set at the time the specified Remote Config template was written.
+   * Output only, except for the version description. Contains metadata about a particular
+   * version of the Remote Config template. All fields are set at the time the specified Remote
+   * Config template was published. A version's description field may be specified in
+   * `publishTemplate` calls.
    */
   export interface Version {
     /**
-     * The version number of the version's corresponding Remote Config template. Output only.
+     * The version number of a Remote Config template.
      */
-    versionNumber?: string | number;
-    
+    versionNumber?: string;
+
     /**
-     * The Remote Config template creation timestamp in RFC3339 UTC "Zulu" format,
-     * accurate to nanoseconds. Output only.
+     * The Remote Config template creation timestamp in UTC.
      */
     updateTime?: string;
 
     /**
-     * The source of origin of the template update action. Output only.
+     * The source of origin of the template update action.
      */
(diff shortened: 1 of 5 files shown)


test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -361,7 +480,7 @@ describe('RemoteConfigApiClient', () => {
     });
   }
 
-  function runErrorResponseTests(rcOperation: () => Promise<RemoteConfigTemplate>): void {
+  function runEtagHeaderTests(rcOperation: () => Promise<RemoteConfigTemplate>): void {
```
lahirumaramba, line 483: Added this function to extract etag header validation tests from `runErrorResponseTests` function. That way I could re-use `runErrorResponseTests` for `list versions` operation. :)

src/index.d.ts
```diff
@@ -982,6 +982,121 @@ declare namespace admin.remoteConfig {
...
+   * All fields are set at the time the specified Remote Config template was written.
+   */
+  export interface Version {
+    /**
+     * The version number of the version's corresponding Remote Config template. Output only.
+     */
+    versionNumber?: string | number;
+    
+    /**
+     * The Remote Config template creation timestamp in RFC3339 UTC "Zulu" format,
```
hiranya911, line 997: We generally expose timestamps in UTC string format in this library. We may want to stick to that convention here also.

src/index.d.ts
```diff
@@ -982,6 +982,121 @@ declare namespace admin.remoteConfig {
...
+     * accurate to nanoseconds. Output only.
+     */
+    updateTime?: string;
+
+    /**
+     * The source of origin of the template update action. Output only.
```
hiranya911, line 1003: Instead of repeating `Output only`, let's just clarify the semantic once at the type header description.

src/index.d.ts
```diff
@@ -982,6 +982,121 @@ declare namespace admin.remoteConfig {
...
+      'REST_API' | 'ADMIN_SDK_NODE');
+
+    /**
+     * The type of the template update action. Output only.
+     */
+    updateType?: ('REMOTE_CONFIG_UPDATE_TYPE_UNSPECIFIED' |
+      'INCREMENTAL_UPDATE' | 'FORCED_UPDATE' | 'ROLLBACK');
+
+    /**
+     * Aggregation of all metadata fields about the account that performed the update. Output only.
```
hiranya911, line 1015: `Metadata about the user that published this version.`

Just curious. What does this get set to for versions published by the Admin SDK?

src/index.d.ts
```diff
@@ -982,6 +982,121 @@ declare namespace admin.remoteConfig {
...
   type RemoteConfigParameterValue = ExplicitParameterValue | InAppDefaultValue;
 
+  /**
+   * Interface representing a Remote Config template version.
+   * Contains all metadata about a particular version of the Remote Config template.
+   * All fields are set at the time the specified Remote Config template was written.
+   */
+  export interface Version {
+    /**
+     * The version number of the version's corresponding Remote Config template. Output only.
```
hiranya911, line 992: `The version number of a Remote Config template.`

src/index.d.ts
```diff
@@ -982,6 +982,121 @@ declare namespace admin.remoteConfig {
...
+    rollbackSource?: string;
+    
+    /**
+     * Indicates weather this Remote Config template was published before version history was
+     * supported. Output only.
+     */
+    isLegacy?: boolean;
+  }
+
+  /** Interface representing a Remote Config list version result. */
```
hiranya911, line 1037: `representing a list of Remote Config template versions.`

src/index.d.ts
```diff
@@ -982,6 +982,121 @@ declare namespace admin.remoteConfig {
...
+  /**
+   * Interface representing a Remote Config template version.
+   * Contains all metadata about a particular version of the Remote Config template.
+   * All fields are set at the time the specified Remote Config template was written.
+   */
+  export interface Version {
+    /**
+     * The version number of the version's corresponding Remote Config template. Output only.
+     */
+    versionNumber?: string | number;
```
hiranya911, line 994: If this is output only let's default to a more specific type. Pick either `string` or `number`.

src/index.d.ts
```diff
@@ -1037,6 +1152,26 @@ declare namespace admin.remoteConfig {
...
     rollback(versionNumber: string | number): Promise<RemoteConfigTemplate>;
 
+    /**
+    * Gets a list of Remote Config template versions that have been published, sorted in reverse 
+    * chronological order. Only the last 300 versions are stored.
+    * All versions that correspond to non-active Remote Config templates (i.e., all except the 
+    * template that is being fetched by clients) are also deleted if they are older than 90 days.
+    * 
+    * @param options Optional options object for getting a list of tempalte versions:
+    *    - {number} `pageSize` The maximum number of items to return per page.
```
hiranya911, line 1162: I don't think we do this anywhere else. Let TSDoc just link this to `ListVersionsOptions`.

src/remote-config/remote-config-api-client.ts
```diff
@@ -86,6 +86,42 @@ export interface RemoteConfigTemplate {
   readonly etag: string;
 }
 
+/** Interface representing a Remote Config version. */
+export interface Version {
+  versionNumber?: string | number; // int64 format
```
hiranya911, line 91: Use a specific type for output fields.

src/index.d.ts
```diff
@@ -982,6 +982,121 @@ declare namespace admin.remoteConfig {
...
+     * Specify the newest version number to include in the results.
+     * If specified, must be greater than zero. Defaults to the newest version.
+     */
+    endVersionNumber?: string | number;
+
+    /**
+     * Specify the earliest update time to include in the results. Any entries updated before this
+     * time are omitted.
+     */
+    startTime?: Date;
```
hiranya911, line 1073: Should we accept `Date | string` here? Our existing APIs (auth.UserMetadataRequest) accepts UTC formatted strings.

src/remote-config/remote-config.ts
```diff
@@ -116,6 +118,28 @@ export class RemoteConfig implements FirebaseServiceInterface {
...
   }
 
+  /**
+  * Gets a list of Remote Config template versions that have been published, sorted in reverse 
+  * chronological order. Only the last 300 versions are stored.
+  * All versions that correspond to non-active Remote Config templates (i.e., all except the 
+  * template that is being fetched by clients) are also deleted if they are older than 90 days.
+  * 
+  * @param {ListVersionsOptions} options Optional options object for getting a list of versions:
+  *    - {number} `pageSize` The maximum number of items to return per page.
```
hiranya911, line 128: This kind of documentation is hard to maintain. Only indicating the type and a summary for the enclosing type is enough.

src/remote-config/remote-config-api-client.ts
```diff
@@ -330,6 +389,49 @@ export class RemoteConfigApiClient {
...
+   * 
+   * @param {ListVersionsOptions} options An options object to be validated.
+   */
+  private validateListVersionsOptions(options: ListVersionsOptions): void {
+    if (!validator.isNonNullObject(options)) {
+      throw new FirebaseRemoteConfigError(
+        'invalid-argument',
+        'ListVersionsOptions must be a non-null object');
+    }
+    if (typeof options.pageSize !== 'undefined' && !validator.isNumber(options.pageSize)) {
```
hiranya911, line 405: Also check for bounds. I guess it should be greater than 0 and less than or eq to 300?

src/index.d.ts
lahirumaramba: It is set to something similar to:
```
{
  email:
   'firebase-adminsdk-vs7zk@PROJECT_ID.iam.gserviceaccount.com',
  imageUrl:
   'https://coded_url/photo.jpg'
}
```

src/index.d.ts
lahirumaramba: Good point! Will update!

src/remote-config/remote-config-api-client.ts
lahirumaramba: The REST API ignores 0 and values greater than 300 and returns all the versions. I added a check anyway :)

src/index.d.ts
lahirumaramba: We did something similar for `publishTemplate` `options` parameter. I agree with you though that we have a lot more parameters here. Will just use a link here.

test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -62,6 +63,33 @@ describe('RemoteConfigApiClient', () => {
     version: {},
   };
 
+  const TEST_VERSIONS_RESULT: ListVersionsResult = {
+    versions:
```
hiranya911, line 67:
```
versions: [
  {
    // ...
  },
]
```

test/unit/remote-config/remote-config.spec.ts
```diff
@@ -193,6 +221,143 @@ describe('RemoteConfig', () => {
...
+      stubs.push(stub);
+      return remoteConfig.listVersions({
+        pageSize: 2
+      })
+        .then((response) => {
+          expect(response.versions.length).to.equal(2);
+          expect(response.versions[0].updateTime).equals('Thu, 07 May 2020 18:46:09 GMT');
+          expect(response.versions[1].updateTime).equals('Thu, 07 May 2020 18:44:41 GMT');
+        });
+    });
```
hiranya911, line 358: Add test cases for input timestamp conversion (Date to ISO and UTC to ISO) and version number conversion (75 to '75').

src/remote-config/remote-config-api-client.ts
```diff
@@ -330,6 +389,61 @@ export class RemoteConfigApiClient {
...
+    if (typeof options.pageSize !== 'undefined' && (options.pageSize < 1 || options.pageSize > 300)) {
+      throw new FirebaseRemoteConfigError(
+        'invalid-argument', 'pageSize must be a number between 1 and 300 (inclusive).');
+    }
+    if (typeof options.pageToken !== 'undefined' && !validator.isNonEmptyString(options.pageToken)) {
+      throw new FirebaseRemoteConfigError(
+        'invalid-argument', 'pageToken must be a string value.');
+    }
+    if (typeof options.endVersionNumber !== 'undefined') {
+      this.validateVersionNumber(options.endVersionNumber, 'endVersionNumber');
```
hiranya911, line 418: If the version is given as a number shouldn't we convert it into string?

src/index.d.ts
```diff
@@ -982,6 +982,122 @@ declare namespace admin.remoteConfig {
...
+   * `publishTemplate` calls.
+   */
+  export interface Version {
+    /**
+     * The version number of a Remote Config template.
+     */
+    versionNumber?: string;
+
+    /**
+     * The Remote Config template creation timestamp in UTC.
```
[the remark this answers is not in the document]
lahirumaramba: I think the name of the parameter used here is a little misguiding. This represents when the Remote Config template was written to the Remote Config server through updateTemplate in the backend API, hence the updateTime. I changed the text to be more descriptive. Let me know what you think.

src/remote-config/remote-config-api-client.ts
```diff
@@ -182,6 +218,29 @@ export class RemoteConfigApiClient {
       });
   }
 
+  public listVersions(options?: ListVersionsOptions): Promise<ListVersionsResult> {
+    if (typeof options !== 'undefined') {
+      // validate options and convert Date objects to UTC strings.
+      this.validateListVersionsOptions(options);
```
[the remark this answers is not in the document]
lahirumaramba: Ah great point! Thanks!

src/remote-config/remote-config-api-client.ts
lahirumaramba: The backend API responds to both numbers and strings :)
Do you think we should always convert to string here for good measure?

src/remote-config/remote-config-api-client.ts
```diff
@@ -330,6 +389,61 @@ export class RemoteConfigApiClient {
...
+      this.validateVersionNumber(options.endVersionNumber, 'endVersionNumber');
+    }
+    if (typeof options.startTime !== 'undefined') {
+      if (!(options.startTime instanceof Date) && !validator.isUTCDateString(options.startTime)) {
+        throw new FirebaseRemoteConfigError(
+          'invalid-argument', 'startTime must be a valid Date object or a UTC date string.');
+      }
+      // Convert startTime to RFC3339 UTC "Zulu" format.
+      if (options.startTime instanceof Date) {
+        options.startTime = options.startTime.toISOString();
```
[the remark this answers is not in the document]
lahirumaramba: Thanks! Fixed.

test/unit/remote-config/remote-config.spec.ts
lahirumaramba: Input timestamp conversion tests are here https://github.com/firebase/firebase-admin-node/pull/896/files#diff-e21807cc3d787a27c233f0b4c5cfc719R442
Do you think we need specific ones for each timestamp?

Also, the API responds to both number and string version numbers... do you think we should always convert them to string?

src/index.d.ts
```diff
@@ -996,7 +996,8 @@ declare namespace admin.remoteConfig {
     versionNumber?: string;
 
     /**
-     * The Remote Config template creation timestamp in UTC.
+     * The timestamp of when the Remote Config template was written to the Remote Config
```
hiranya911, line 999: when this version of the template was...

src/remote-config/remote-config-api-client.ts
```diff
@@ -219,17 +220,17 @@ export class RemoteConfigApiClient {
   }
 
   public listVersions(options?: ListVersionsOptions): Promise<ListVersionsResult> {
+    let optionsCopy: ListVersionsOptions;
     if (typeof options !== 'undefined') {
-      // validate options and convert Date objects to UTC strings.
-      this.validateListVersionsOptions(options);
+      optionsCopy = this.validateListVersionsOptions(options);
```
hiranya911, line 225: can we do `options = ...` here and avoid another variable?

src/remote-config/remote-config-api-client.ts
```diff
@@ -368,7 +369,7 @@ export class RemoteConfigApiClient {
    *
    * @param {string|number} versionNumber A version number to be validated.
    */
-  private validateVersionNumber(versionNumber: string | number, propertyName: string): void {
+  private validateVersionNumber(versionNumber: string | number, propertyName = 'versionNumber'): void {
```
hiranya911, line 372: Callers can now just do: `validateVersionNumber(versionNumber)`

src/remote-config/remote-config-api-client.ts
```diff
@@ -391,58 +392,64 @@ export class RemoteConfigApiClient {
...
    * 
    * @param {ListVersionsOptions} options An options object to be validated.
+   * 
+   * @return {ListVersionsOptions} A copy of the provided options object with timestamps converted
+   * to UTC Zulu format.
    */
-  private validateListVersionsOptions(options: ListVersionsOptions): void {
-    if (!validator.isNonNullObject(options)) {
+  private validateListVersionsOptions(options: ListVersionsOptions): ListVersionsOptions {
+    const optionsCopy: ListVersionsOptions = deepCopy(options);
```
hiranya911, line 404: It seems TS best practice is to avoid explicit type annotations whenever possible. We haven't really followed this guidance in the past, but we should probably get into that habit.

src/remote-config/remote-config-api-client.ts
hiranya911: > The backend API responds to both numbers and strings

🤯 

> Do you think we should always convert to string here for good measure?

Feels like that's the right thing to do, but I'm not sure. Your call :)

test/unit/remote-config/remote-config.spec.ts
hiranya911: Ah. It's a little buried in the code. Can we make it a bit more explicit?

PR #906, opened 2020-06-15
Add version meta data to RC templates

lahirumaramba:
- Add Version Metadata to Remote Config templates
- Exclude output only parameters from metadata in input templates
- Add unit tests

### DIFF
--- a/test/unit/remote-config/remote-config.spec.ts
+++ b/test/unit/remote-config/remote-config.spec.ts
@@ -28,7 +28,6 @@ import {
   RemoteConfigCondition,
   TagColor,
   ListVersionsResult,
-  Version,
 } from '../../../src/remote-config/remote-config-api-client';
 import { FirebaseRemoteConfigError } from '../../../src/remote-config/remote-config-utils';
 import { deepCopy } from '../../../src/utils/deep-copy';
@@ -54,6 +53,7 @@ describe('RemoteConfig', () => {
       },
     },
   };
+
   const VERSION_INFO = {
     versionNumber: '86',
     updateOrigin: 'ADMIN_SDK_NODE',
@@ -614,9 +614,7 @@ describe('RemoteConfig', () => {
           }).to.throw(
             'Cannot set property etag of #<RemoteConfigTemplateImpl> which has only a getter');
 
-          const v = template.version;
-          expect(v).to.be.not.undefined;
-          const version = v as Version;
+          const version = template.version!;
           expect(version.versionNumber).to.equal('86');
           expect(version.updateOrigin).to.equal('ADMIN_SDK_NODE');
           expect(version.updateType).to.equal('INCREMENTAL_UPDATE');
--- a/test/unit/remote-config/remote-config.spec.ts
+++ b/test/unit/remote-config/remote-config.spec.ts
@@ -28,6 +28,7 @@ import {
   RemoteConfigCondition,
   TagColor,
   ListVersionsResult,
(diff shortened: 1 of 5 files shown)


test/unit/remote-config/remote-config.spec.ts
```diff
@@ -596,6 +614,18 @@ describe('RemoteConfig', () => {
           }).to.throw(
             'Cannot set property etag of #<RemoteConfigTemplateImpl> which has only a getter');
 
+          const v = template.version;
+          expect(v).to.be.not.undefined;
+          const version = v as Version;
```
hiranya911, line 619: Why is this needed? Shouldn't this already be typed `Version`?

test/unit/remote-config/remote-config.spec.ts
```diff
@@ -596,6 +614,18 @@ describe('RemoteConfig', () => {
...
+          expect(version.versionNumber).to.equal('86');
```
hiranya911, line 620: Can we do this:

```
expect(version).to.deep.equal(VERSION_INFO);
```

test/unit/remote-config/remote-config.spec.ts
```diff
@@ -612,7 +642,11 @@ describe('RemoteConfig', () => {
           expect(cond.tagColor).to.equal(TagColor.BLUE);
 
           const parsed = JSON.parse(JSON.stringify(template));
-          expect(parsed).deep.equals(REMOTE_CONFIG_RESPONSE);
+          const templateCopy = deepCopy(REMOTE_CONFIG_RESPONSE);
```
hiranya911, line 645: I had a little trouble following the logic here. May be rename the variables to `expectedTemplate` and `expectedVersion`.

test/unit/remote-config/remote-config.spec.ts
lahirumaramba: If I do not cast it here the compiler complains that the object might be undefined. I think it is because of version is optional. We do the same for conditions below...
`const cond = c as RemoteConfigCondition;` any thoughts?

test/unit/remote-config/remote-config.spec.ts
lahirumaramba: `version` is a `Version` type and `VERSION_INFO` is a js object so deep equal didn't work unfortunately...

test/unit/remote-config/remote-config.spec.ts
lahirumaramba: Ah good call! Thanks!

test/unit/remote-config/remote-config.spec.ts
lahirumaramba: An alternative is to define a global `Version` const an use that here

test/unit/remote-config/remote-config.spec.ts
hiranya911: Add an assertion.

```
const version = template.version!;
```

test/unit/remote-config/remote-config.spec.ts
hiranya911: If that's feasible let's do that. Otherwise the way it is now is not terrible either.

lahirumaramba: Looks like it wouldn't work unless we use an instance of the internal `VersionImpl`, which I do not think we can use here.

PR #909, opened 2020-06-17
fix(auth): Add interfaces for UserProviderRequest and Metadata used in UserImportRecord

MathBunny:
I think there is an inconsistency with the defined `auth.d.ts` file [here](https://github.com/firebase/firebase-admin-node/blob/master/src/auth.d.ts#L802). This aims to add the interface to be consistent and then have an array of `UserProviderRequest`.

Similar change applies also for [metadata](https://github.com/firebase/firebase-admin-node/blob/master/src/auth.d.ts#L696).

Not sure if this is the best approach.

### DIFF
--- a/src/auth/user-import-builder.ts
+++ b/src/auth/user-import-builder.ts
@@ -59,7 +59,6 @@ interface UserProviderRequest {
 interface UserMetadataRequest {
   lastSignInTime?: string;
   creationTime?: string;
-  lastRefreshTime?: string|null;
 }
 
 /** User import record as accepted from developer. */
--- a/src/auth/user-import-builder.ts
+++ b/src/auth/user-import-builder.ts
@@ -47,6 +47,14 @@ interface SecondFactor {
   factorId: string;
 }
 
+interface UserProviderRequest {
+  uid: string;
+  displayName?: string;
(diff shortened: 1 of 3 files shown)


src/auth/user-import-builder.ts
```diff
@@ -47,6 +47,20 @@ interface SecondFactor {
...
+  email?: string;
+  phoneNumber?: string;
+  photoURL?: string;
+  providerId: string;
+}
+
+interface UserMetadataRequest {
+  lastSignInTime?: string;
+  creationTime?: string;
+  lastRefreshTime?: string|null;
```
bojeil-google, line 62: I am not sure `lastRefreshTime` is supported for import operations. Please remove this one for now.

MathBunny: Thanks -- are you referring to the specific field or should I remove the entire interface?

bojeil-google: Just this specific field.

hiranya911: @bojeil-google we have the following in our typings for `UserMetadataRequest`:

```
lastRefreshTime?: string|null;
```

Should we remove that too?

PR #911, opened 2020-06-19
fix(auth): Make uid mandatory for UserImportRecord

MathBunny:
In the type signature for the equivalent of `UserImportRecord`'s `multiFactor.enrolledFactors`, the `uid` is an optional field, however in the implementation it is mandatory. I am not certain if it should be mandatory or not, but I think it should be mandatory.

To account for this, originally I added an extra interface in the hierarchy of `UpdateRequest` and `CreateRequest` that will mandate the `uid` is provided, because otherwise the type-checker would raise a warning as `UpdateMultiFactorInfoRequest` cannot be extended by `CreateMultiFactorInfoRequest`.

However I realized that instead a better approach might be to simply bring the `SecondFactor` interface to the typing so that the other hierarchy doesn't change.

Summary of observation:

## Source
- `UserImportRecord` expects `SecondFactor[]` for property `multiFactor`: https://github.com/firebase/firebase-admin-node/blob/b5909906bd4b73630f2a6311ac06b9e65be2d621/src/auth/user-import-builder.ts#L52-L73
- `SecondFactor` (UID not optional): https://github.com/firebase/firebase-admin-node/blob/bed1e5ec6c3e515b037b993cdbbeb1f5b585f56a/src/auth/user-import-builder.ts#L42-L48
 
## Typing
- `UserImportRecord` expects `MultiFactorUpdateSettings`: https://github.com/firebase/firebase-admin-node/blob/bed1e5ec6c3e515b037b993cdbbeb1f5b585f56a/src/auth.d.ts#L828
- `MultiFactorUpdateSettings` yields `UpdateMultiFactorInfoRequest`:  https://github.com/firebase/firebase-admin-node/blob/bed1e5ec6c3e515b037b993cdbbeb1f5b585f56a/src/auth.d.ts#L255-L262
- `UpdateMultiFactorInfoRequest` (UID as optional): 
https://github.com/firebase/firebase-admin-node/blob/bed1e5ec6c3e515b037b993cdbbeb1f5b585f56a/src/auth.d.ts#L269-L275

## Other SDKs
- It appears that in the Java SDK the UID is mandatory: [link](https://github.com/firebase/firebase-admin-java/blob/master/src/main/java/com/google/firebase/auth/ImportUserRecord.java#L254)

### DIFF
--- a/src/auth.d.ts
+++ b/src/auth.d.ts
@@ -825,7 +825,20 @@ export namespace admin.auth {
     /**
      * The user's multi-factor related properties.
      */
-    multiFactor?: admin.auth.MultiFactorUpdateSettings;
+    multiFactor?: {
+      enrolledFactors: admin.auth.SecondFactor[];
+    };
+  }
+
+  /**
+   * Interface representing user's multi-factor related properties.
+   */
+  interface SecondFactor {
+    uid: string;
+    phoneNumber: string;
+    displayName?: string;
+    enrollmentTime?: string;
+    factorId: string;
   }
 
   /**
--- a/src/auth/user-import-builder.ts
+++ b/src/auth/user-import-builder.ts
@@ -39,7 +39,8 @@ export interface UserImportOptions {
   };
 }
 
-interface SecondFactor {
+/** Interface representing user's multi-factor related properties */
+export interface SecondFactor {
   uid: string;
   phoneNumber: string;
   displayName?: string;
--- a/src/auth/user-record.ts
+++ b/src/auth/user-record.ts
@@ -94,6 +94,7 @@ export interface UpdateRequest {
 
 /** Parameters for create user operation */
 export interface CreateRequest extends UpdateRequest {
+  uid?: string;
   multiFactor?: {
     enrolledFactors: CreateMultiFactorInfoRequest[];
   };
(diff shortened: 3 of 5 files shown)


MathBunny:
Actually, a disadvantage of this approach is that it's more of a breaking change than adding the hierarchy, since now it introduces a new interface to those types. Thoughts would be appreciated (or if there's a better approach or if it's better to not change it at all/there is no issue).

hiranya911:
Yes, this is a potentially breaking change, but it might be the right fix in the long run. I think our options are:

1. Wait to make this change until the next major release (we should create a bug to track this).
2. Make the uid required if it's actually required in all the possible flows (this may not be the case, and if so this is not really an option for us).

bojeil-google:
Some background on this. I intentionally didn't want to create multiple interfaces for this.
Different endpoints use this as optional, not needed or required.

1. When retrieving a user with a second factor, `uid` will always be there
2. When listing users with a second factor, `uid` will always be there
3. When creating a user with a second factor via `createUser`, `uid` is not needed and the Auth server will always provision one.
4. When updating an existing user second factors, `uid` is optional. If not provided, Auth server will provision one.
5. When importing users with second factors in bulk (usually from another system), `uid` is optional. If not provided, Auth server will provision one.

hiranya911:
Thanks @bojeil-google. So it looks like the public API surface (d.ts) files are correct. It's just the `SecondFactor` type in the implementation is used in use cases where the `uid` is optional too (a reasonable implementation choice). 

@MathBunny I feel like we can live with the current implementation. If it has any impact on your ongoing refactoring work, let's have a separate conversation about how to get around it.

MathBunny:
Sounds good, thank you for the explanation!

PR #912, opened 2020-06-19
fix: Upgraded node-forge and jsonwebtoken dependencies

### LINKED ISSUE #905: Update jsonwebtoken and node-forge
For some reason the version of `jsonwebtoken` is pinned to 8.1.0 (last updated 3 years ago) and `node-forge` is pinned to 0.7.4 (last updated 2 years ago).

Consider either bumping their versions or using the caret `^` if there's no good reason to have them pinned.

Note that both packages have received several updates including bug fixes and security updates, `jsonwebtoken` latest version is 8.5.1 and `node-forge` 0.9.1.

Note that pinning the versions mean that `npm` won't deduplicated the dependencies and multiple versions may be present in a project, making deployments slower and using more memory.

hiranya911:
Resolves #905

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,8 +57,8 @@
     "@firebase/database": "^0.6.0",
     "@types/node": "^8.10.59",
     "dicer": "^0.3.0",
-    "jsonwebtoken": "8.1.0",
-    "node-forge": "0.7.4"
+    "jsonwebtoken": "^8.5.1",
+    "node-forge": "^0.7.6"
   },
   "optionalDependencies": {
     "@google-cloud/firestore": "^3.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -57,8 +57,8 @@
     "@firebase/database": "^0.6.0",
     "@types/node": "^8.10.59",
     "dicer": "^0.3.0",
-    "jsonwebtoken": "8.1.0",
-    "node-forge": "0.7.4"
+    "jsonwebtoken": "^8.5.1",
+    "node-forge": "^0.7.6"
```
lahirumaramba, line 61: Any reason why we wouldn't update this to most recent`0.9.1`?

hiranya911: I upgraded via `npm update` command, and this is what was automatically picked by NPM (possibly because 0.x versions are allowed to make breaking changes between minor releases). I will test 0.9.1 and upgrade in a separate PR.

lahirumaramba: Sounds good. Thanks for checking!

PR #914, opened 2020-06-23
Add integration tests for RC manage version operations

lahirumaramba:
- Add integration tests for `listVersions`, `getTemplateAtVersion`, and `rollback` operations
- Add integration tests for `versions` metadata in Remote Config templates
- Remove undefined values in list versions options, and add unit tests
- Check for empty versions lists in `listVersions`, and add unit tests

Note: Added `release:stage` to trigger integration tests. Merging to ~`remote-config-vc`~ `master`.

### DIFF
--- a/src/remote-config/remote-config-api-client.ts
+++ b/src/remote-config/remote-config-api-client.ts
@@ -237,6 +237,9 @@ export class RemoteConfigApiClient {
   public listVersions(options?: ListVersionsOptions): Promise<ListVersionsResult> {
     if (typeof options !== 'undefined') {
       options = this.validateListVersionsOptions(options);
+      // Remove undefined fields from options
+      Object.keys(options).forEach(key =>
+        (options as any)[key] === undefined && delete (options as any)[key])
     }
     return this.getUrl()
       .then((url) => {
--- a/src/remote-config/remote-config-api-client.ts
+++ b/src/remote-config/remote-config-api-client.ts
@@ -237,9 +237,6 @@ export class RemoteConfigApiClient {
   public listVersions(options?: ListVersionsOptions): Promise<ListVersionsResult> {
     if (typeof options !== 'undefined') {
       options = this.validateListVersionsOptions(options);
-      // Remove undefined fields from options
-      Object.keys(options).forEach(key =>
-        (options as any)[key] === undefined && delete (options as any)[key])
     }
     return this.getUrl()
       .then((url) => {
@@ -480,6 +477,10 @@ export class RemoteConfigApiClient {
         optionsCopy.endTime = new Date(optionsCopy.endTime).toISOString();
       }
     }
+    // Remove undefined fields from optionsCopy
+    Object.keys(optionsCopy).forEach(key =>
+      (typeof (optionsCopy as any)[key] === 'undefined') && delete (optionsCopy as any)[key]
+    );
     return optionsCopy;
   }
 }
--- a/src/remote-config/remote-config.ts
+++ b/src/remote-config/remote-config.ts
@@ -144,8 +144,10 @@ export class RemoteConfig implements FirebaseServiceInterface {
   public listVersions(options?: ListVersionsOptions): Promise<ListVersionsResult> {
     return this.client.listVersions(options)
       .then((listVersionsResponse) => {
+        const versions = (typeof listVersionsResponse.versions == 'undefined') ? [] :
+          listVersionsResponse.versions.map(version => new VersionImpl(version));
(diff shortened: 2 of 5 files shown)


lahirumaramba:
Note: That failing integration test is from ML. Fix is in `master`

src/remote-config/remote-config.ts
```diff
@@ -144,8 +144,10 @@ export class RemoteConfig implements FirebaseServiceInterface {
   public listVersions(options?: ListVersionsOptions): Promise<ListVersionsResult> {
     return this.client.listVersions(options)
       .then((listVersionsResponse) => {
+        const versions = (typeof listVersionsResponse.versions == 'undefined') ? [] :
```
hiranya911, line 147:
```
const versions = listVersionsResponse.versions?.map(...) ?? [];
```

test/integration/remote-config.spec.ts
```diff
@@ -57,16 +57,22 @@ const VALID_PARAMETER_GROUPS = {
...
+    tagColor: 'INDIGO',
+  },
+  {
+    name: 'android',
+    expression: 'device.os == \'android\'',
+    tagColor: 'GREEN',
+  },
+];
+
+const VALID_VERSION: admin.remoteConfig.Version = {
```
hiranya911, line 73: Just wondering. Are these explicit type annotations really required?

src/remote-config/remote-config-api-client.ts
```diff
@@ -237,6 +237,9 @@ export class RemoteConfigApiClient {
   public listVersions(options?: ListVersionsOptions): Promise<ListVersionsResult> {
     if (typeof options !== 'undefined') {
       options = this.validateListVersionsOptions(options);
+      // Remove undefined fields from options
+      Object.keys(options).forEach(key =>
+        (options as any)[key] === undefined && delete (options as any)[key])
```
hiranya911, line 242: I feel like this logic belongs in the `validateListVersionsOptions` method.

Also `typeof ... === 'undefined'`

test/integration/remote-config.spec.ts
```diff
@@ -143,6 +157,87 @@ describe('admin.remoteConfig', () => {
...
+      versionOneNumber = activeTemplate.version!.versionNumber!;
+
+      // publish another template to create a second version number
+      activeTemplate.version = { description: versionTwoDescription };
+      activeTemplate = await admin.remoteConfig().publishTemplate(activeTemplate)
+      expect(activeTemplate.version).to.be.not.undefined;
+      versionTwoNumber = activeTemplate.version!.versionNumber!;
+    });
+
+    it('verfy that getTemplateAtVersion() returns the requested template version v1', () => {
```
hiranya911, line 189: Use the familiar wording style. `should return....`

test/integration/remote-config.spec.ts
```diff
@@ -143,6 +157,87 @@ describe('admin.remoteConfig', () => {
...
+      return admin.remoteConfig().getTemplateAtVersion(versionOneNumber)
+        .then((template) => {
+          expect(template.etag).matches(/^etag-[0-9]*-[0-9]*$/);
+          expect(template.version).to.be.not.undefined;
+          expect(template.version!.versionNumber).equals(versionOneNumber);
+          expect(template.version!.description).equals(versionOneDescription);
+        });
+    });
+
+    it('verfy that getTemplateAtVersion() returns the requested template version v2', () => {
```
hiranya911, line 199: Seems redundant.

src/remote-config/remote-config-api-client.ts
lahirumaramba: Good call. Moved to `validateListVersionsOptions`

src/remote-config/remote-config.ts
lahirumaramba: Thanks!

test/integration/remote-config.spec.ts
lahirumaramba: I removed the type annotations for `VALID_VERSION`. It looks like we need the type annotation for `VALID_CONDITIONS` unless I use a `const INDIGO: admin.remoteConfig.TagColor = 'INDIGO'` for tag color.

test/integration/remote-config.spec.ts
lahirumaramba: This was to confirm that we are not getting the most recent template and having two versions confirms it. I see your point though :)
Will remove the `expect`s here, but leave the templates for `listVersions` tests.

PR #916, opened 2020-06-23
fix(auth): Migrated IAM sign endpoint to iamcredentials.googleapis.com

hiranya911:
The iam.googleapis.com API that we use for signing JWTs is going away and being replaced by a new IAM service accounts API: https://cloud.google.com/iam/docs/migrating-to-credentials-api#blob

This PR migrates our existing code that relies on this endpoint.


hiranya911:
I'm seeing unrelated integration test failures due to `deleteUsers()` exceeding a quota. This will be addressed in a separate PR.

PR #926, opened 2020-07-03
Add null to databaseAuthVariableOverride

MathBunny:
I noticed that the `databaseAuthVariableOverride` does not accept null in the source:

https://github.com/firebase/firebase-admin-node/blob/488f9318350c6b46af2e93b99907b9a02f170029/src/firebase-app.ts#L48-L56

Meanwhile, in typings it accepts null:

https://github.com/firebase/firebase-admin-node/blob/488f9318350c6b46af2e93b99907b9a02f170029/src/index.d.ts#L138-L163

Another observation is `object` vs `Object` discrepancy, from my understanding `object` is generally preferred but I don't know if it's safe or worth changing the typings for this.

### DIFF
--- a/src/firebase-app.ts
+++ b/src/firebase-app.ts
@@ -47,7 +47,7 @@ export type AppHook = (event: string, app: FirebaseApp) => void;
  */
 export interface FirebaseAppOptions {
   credential?: Credential;
-  databaseAuthVariableOverride?: object;
+  databaseAuthVariableOverride?: object | null;
   databaseURL?: string;
   serviceAccountId?: string;
   storageBucket?: string;


lahirumaramba:
It makes sense to me. I am adding @rsgowman for his input as the `null` checks were added as part of #676 

I think you are right. We probably should use `object` instead of `Object`. I see multiple places where we have used `Object` in code :)
https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

lahirumaramba:
Thank you, @rsgowman !

hiranya911:
@schmidt-sebastian can you also take a quick look?

MathBunny:
Thanks @schmidt-sebastian - will merge this now and then will discuss with Hiranya and Lahiru next steps for the `Object` vs `object` discrepancy in another PR.

PR #928, opened 2020-07-06
Prototype for moving internally used types to dedicated internal classes in Auth

MathBunny:
When using autogenerated typing, some types/classes are exported but not intended to be used publicly. Formerly this was implemented by not including the types in the `d.ts` file, but now since they're autogenerated we have the option of either:

1. Denoting the type using the _undocumented_ strip-internal `@internal` annotation
2. Split it off into another file that is not exported via the barrel export

This PR explores the second option for the Auth module. While the typings are still generated, users will not be able to import them regularly via the `firebase-admin/auth` import.


MathBunny:
Finished the entire `auth` service. I didn't run into any issues when doing the conversion, except functions with default parameters. For those I wasn't able to set the default value inside the interface, and instead overloaded the function. I'm going to look into alternatives now.

MathBunny:
Okay, turns out in TypeScript you cannot have default parameter values for declared functions. Instead, I denoted the parameter as optional using `?` which eliminates the need to overload.

PR #739, opened 2019-12-21
Adds scrypt support in node 12

bojeil-google:
Replaces scrypt npm module (no longer maintained and
doesn't work for node 12) for testing standard scrypt with
`crypto.scryptSync` which has been supported since node v10.5.

PR #934, opened 2020-07-08
change: Dropped Node 8 support and upgraded Firestore/Storage dependencies

hiranya911:
Building on top of #739 this PR:

1. Upgrades `@google-cloud/firestore` to latest major version.
2. Adds Node 12 and 14 to CI matrix.
3. Bumped some typing dependencies to make the compiler happy.

API CHANGE: Admin SDK now requires Node.js 10 or higher. Node.js 8 support has been terminated.
API CHANGE: Upgraded dependency on the `@google-cloud/firestore` package to v4.
API CHANGE: Upgraded dependency on the `@google-cloud/storage` package to v5.

Resolves #934

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "node-forge": "^0.7.6"
   },
   "optionalDependencies": {
-    "@google-cloud/firestore": "^3.0.0",
+    "@google-cloud/firestore": "^4.0.0",
     "@google-cloud/storage": "^4.1.2"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -61,15 +61,15 @@
     "node-forge": "^0.7.6"
   },
   "optionalDependencies": {
-    "@google-cloud/firestore": "^3.0.0",
+    "@google-cloud/firestore": "^4.0.0",
     "@google-cloud/storage": "^4.1.2"
```
schmidt-sebastian, line 65: Please update to 5.x as well.

hiranya911:
Engines field was updated in #739

package.json
hiranya911: Done

package.json
```diff
@@ -61,15 +61,15 @@
     "node-forge": "^0.7.6"
```
sk-, line 61: @hiranya911 Given that you are already updating the dependencies, could you update this one to 0.9.1. Given that it is a 0.x version, npm considers it to be a breaking change, and hence not compatible with the current requirement.

`node-forge` is also transitively being used by firestore (direct dependency is google-p12-pem), and requiring version ^0.9.0, which means two copies of the library are currently needed. Switching to just one dependency would save 1.7Mib.

hiranya911:
Good point @sk-. I'll send a separate PR for that. Feel free to beat me to it :)

PR #937, opened 2020-07-08
[Pending] Specifying hashInputOrder

xil222:
waiting for requirement of md5 confirmation
add functionality and tests for specifying inputHashOrder for algorithms:
hmac{md5, sha1, sha256, sha512} and {sha1, sha256, sha512}

### DIFF
--- a/src/auth.d.ts
+++ b/src/auth.d.ts
@@ -570,6 +570,8 @@ export namespace admin.auth {
     pageToken?: string;
   }
 
+  type HashInputOrderType = 'SALT_FIRST' | 'PASSWORD_FIRST';
+
   type HashAlgorithmType = 'SCRYPT' | 'STANDARD_SCRYPT' | 'HMAC_SHA512' |
     'HMAC_SHA256' | 'HMAC_SHA1' | 'HMAC_MD5' | 'MD5' | 'PBKDF_SHA1' | 'BCRYPT' |
     'PBKDF2_SHA256' | 'SHA512' | 'SHA256' | 'SHA1';
@@ -632,11 +634,21 @@ export namespace admin.auth {
        * `STANDARD_SCRYPT` algorithm.
        */
       blockSize?: number;
+
       /**
        * The derived key length of the hashing algorithm. Required for the
        * `STANDARD_SCRYPT` algorithm.
        */
       derivedKeyLength?: number;
+
+      /**
+       * The hash input order for certain hashing algorithms. 
+       * orders supported:
+       * 'SALT_FIRST', 'PASSWORD_FIRST'.
+       * Optional for `HMAC_SHA512`, `HMAC_SHA256`, `HMAC_SHA1`,
+       * `HMAC_MD5`, `MD5`, `SHA512`, `SHA256` and `SHA1`.
+       */
+      inputOrder?: HashInputOrderType;
     };
   }
 
--- a/src/auth/user-import-builder.ts
+++ b/src/auth/user-import-builder.ts
@@ -20,6 +20,10 @@ import * as validator from '../utils/validator';
 import {AuthClientErrorCode, FirebaseAuthError, FirebaseArrayIndexError} from '../utils/error';
 
 /** Firebase Auth supported hashing algorithms for import operations. */
+export type HashInputOrderType = 'SALT_FIRST' | 'PASSWORD_FIRST';
+
+export type PasswordHashOrderType = 'SALT_AND_PASSWORD' | 'PASSWORD_AND_SALT';
+
 export type HashAlgorithmType = 'SCRYPT' | 'STANDARD_SCRYPT' | 'HMAC_SHA512' |
     'HMAC_SHA256' | 'HMAC_SHA1' | 'HMAC_MD5' | 'MD5' | 'PBKDF_SHA1' | 'BCRYPT' |
     'PBKDF2_SHA256' | 'SHA512' | 'SHA256' | 'SHA1';
@@ -36,6 +40,7 @@ export interface UserImportOptions {
     parallelization?: number;
     blockSize?: number;
     derivedKeyLength?: number;
+    inputOrder?: HashInputOrderType;
   };
 }
 
@@ -129,6 +134,7 @@ export interface UploadAccountOptions {
   parallelization?: number;
   blockSize?: number;
   dkLen?: number;
+  passwordHashOrder?: PasswordHashOrderType;
 }
 
 
@@ -371,6 +377,26 @@ export class UserImportBuilder {
   }
 
   /**
+   * Checks whether is a hashInputOrder-supported hashing algorithm.
+   * @param {HashAlgorithmType | undefined} hashAlgorithmType Hash algorithm type.
+   * @return {boolean} Whether is hashInputOrderSupported.
+   */
+  private hashInputOrderSupported(hashAlgorithmType: HashAlgorithmType): boolean {
+    switch(hashAlgorithmType) {
+    case 'HMAC_SHA512':
+    case 'HMAC_SHA256':
+    case 'HMAC_SHA1':
+    case 'HMAC_MD5':
+    case 'MD5':
+    case 'SHA512':
+    case 'SHA256':
+    case 'SHA1':
+      return true;
+    }
+    return false;
+  }
+
+  /**
    * Validates and returns the hashing options of the uploadAccount request.
    * Throws an error whenever an invalid or missing options is detected.
    * @param {UserImportOptions} options The UserImportOptions.
@@ -552,8 +578,39 @@ export class UserImportBuilder {
         `Unsupported hash algorithm provider "${options.hash.algorithm}".`,
       );
     }
+
+    if (this.hashInputOrderSupported(options.hash.algorithm)) {
+      populatedOptions = this.populateInputOrderOption(populatedOptions, options.hash.inputOrder);
+    }
+
     return populatedOptions;
   }
+
+  /**
+   * Adds a new hash order field to the populated account options, if provided.
+   * @param {UploadAccountOptions} populatedOptions The already populated account options.
+   * @param {HashInputOrderType=} inputOrder The optional hash input order.
+   * @return {UploadAccountOptions} A copy of populated options with a new field if the provided
+   *     input order was provided.
+   */
+  private populateInputOrderOption(
+    populatedOptions: UploadAccountOptions,
+    inputOrder?: HashInputOrderType,
+  ): UploadAccountOptions {
+    const populatedOptionsCopy = {...populatedOptions};
+
+    if (typeof inputOrder !== 'undefined') {
+      if (inputOrder !== 'SALT_FIRST' && inputOrder !== 'PASSWORD_FIRST') {
+        throw new FirebaseAuthError(AuthClientErrorCode.INVALID_HASH_INPUT_ORDER);
+      }
+
+      populatedOptionsCopy.passwordHashOrder =
+        inputOrder === 'SALT_FIRST' ? 'SALT_AND_PASSWORD' : 'PASSWORD_AND_SALT';
+    }
+
+    return populatedOptionsCopy;
+  }
+
 
   /**
    * Validates and returns the users list of the uploadAccount request.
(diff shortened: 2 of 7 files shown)


hiranya911:
This implementation makes sense to me. It also happens to match an internally approved API proposal. @bojeil-google WDYT?

test/unit/auth/user-import-builder.spec.ts
```diff
@@ -230,22 +229,58 @@ describe('UserImportBuilder', () => {
...
+          expect(() =>  {
+            return new UserImportBuilder(users, invalidOptions as any, userRequestValidator);
+          }).to.throw(expectedError.message);
+        });
+
+        it('should not throw with valid PASSWORD_FIRST options and should generate expected request', () => {
+          const validOptions = {
+            hash: {
+              algorithm,
+              inputOrder: "PASSWORD_FIRST",
```
bojeil-google, line 250: Use single quotes: 'PASSWORD_FIRST'

test/unit/auth/user-import-builder.spec.ts
```diff
@@ -230,22 +229,58 @@ describe('UserImportBuilder', () => {
...
+          const userImportBuilder =
+              new UserImportBuilder(users, validOptions as any, userRequestValidator);
+          expect(userImportBuilder.buildRequest()).to.deep.equal(expectedRequest);
+        });
+
+        it('should not throw with valid SALT_FIRST options and should generate expected request', () => {
           const validOptions = {
             hash: {
               algorithm,
+              inputOrder: "SALT_FIRST",
```
bojeil-google, line 269: Use single quotes: 'SALT_FIRST'

test/unit/auth/user-import-builder.spec.ts
```diff
@@ -257,7 +292,7 @@ describe('UserImportBuilder', () => {
         case 'MD5':
           minRounds = 0;
           maxRounds = 8192;
-          break;
+          break;
```
bojeil-google, line 295: Remove trailing spaces.

test/unit/auth/user-import-builder.spec.ts
```diff
@@ -230,22 +229,58 @@ describe('UserImportBuilder', () => {
           });
         });
 
-        it('should not throw with valid options and should generate expected request', () => {
+        it('should throw when an invalid hash input order option is provided', () => {
```
bojeil-google, line 232: Add test when hash input order is specified with algorithm that doesn't support it.

test/unit/auth/user-import-builder.spec.ts
```diff
@@ -293,23 +328,59 @@ describe('UserImportBuilder', () => {
...
+            expect(() =>  {
+              return new UserImportBuilder(users, invalidOptions as any, userRequestValidator);
+            }).to.throw(expectedError.message);
+          });
+
+          it('should not throw with valid SALT_FIRST options and should generate expected request', () => {
+            const validOptions = {
+              hash: {
+                algorithm,
+                inputOrder: "SALT_FIRST",
```
bojeil-google, line 350: Use single quotes: 'SALT_FIRST'

test/unit/auth/user-import-builder.spec.ts
```diff
@@ -293,23 +328,59 @@ describe('UserImportBuilder', () => {
...
+            const userImportBuilder =
+                new UserImportBuilder(users, validOptions as any, userRequestValidator);
+            expect(userImportBuilder.buildRequest()).to.deep.equal(expectedRequest);
+          });
 
+          it('should not throw with valid PASSWORD_FIRST options and should generate expected request', () => {
+            const validOptions = {
+              hash: {
+                algorithm,
+                inputOrder: "PASSWORD_FIRST",
```
bojeil-google, line 369: Use single quotes: 'PASSWORD_FIRST'

test/unit/auth/user-import-builder.spec.ts
```diff
@@ -293,23 +328,59 @@ describe('UserImportBuilder', () => {
           });
         });
 
-        it('should not throw with valid options and should generate expected request', () => {
-          const validOptions = {
-            hash: {
-              algorithm,
+        if (algorithm !== 'PBKDF_SHA1' && algorithm !== 'PBKDF2_SHA256' && algorithm !== 'MD5') {
```
bojeil-google, line 331: Why are you excluding these algorithms specifically?

test/integration/auth.spec.ts
```diff
@@ -1652,6 +1655,58 @@ describe('admin.auth', () => {
...
+      {
+        name: 'HMAC_MD5',
+        importOptions: {
+          hash: {
+            algorithm: 'HMAC_MD5',
+            key: Buffer.from('secret'),
+          },
+        } as any,
+        computeSaltFirstPasswordHash: (userImportTest: UserImportTest): Buffer => {
+          return userImportTest.computePasswordHash(userImportTest, 'SALT_FIRST');
```
bojeil-google, line 1667: I think this is better done as an option to `importOptions`. After all, this is where this parameter is expected:
```
      importOptions: {
          hash: {
            algorithm: 'HMAC_MD5',
            key: Buffer.from('secret'),
            inputOrder: 'SALT_FIRST',
          },
        } as any,
```

You can then check the order in `computePasswordHash`

bojeil-google (changes requested):

src/auth/user-import-builder.ts
```diff
@@ -370,6 +376,25 @@ export class UserImportBuilder {
     return importResult;
   }
 
+  /**
+   * Checks whether is a hashInputOrder-supported hashing algorithm.
+   * @param {HashAlgorithmType | undefined} hashAlgorithmType Hash algorithm type.
+   * @return {boolean} Whether is hashInputOrderSupported.
```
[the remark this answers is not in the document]
xil222: Do you mean this? 
Whether hash input order is supported.

test/unit/auth/user-import-builder.spec.ts
xil222: We don't support specifying hashing order for 'PBKDF_SHA1', 'MD5', and 'PBKDF2_SHA256' algorithms. Inside the if statement we're only testing special cases for password hash order supported algorithms.

src/auth/user-import-builder.ts
```diff
@@ -20,6 +20,10 @@ import * as validator from '../utils/validator';
 import {AuthClientErrorCode, FirebaseAuthError, FirebaseArrayIndexError} from '../utils/error';
 
 /** Firebase Auth supported hashing algorithms for import operations. */
+export type HashInputOrderType = 'SALT_FIRST' | 'PASSWORD_FIRST';
+
+export type PasswordHashOrderType = 'SALT_AND_PASSWORD' | 'PASSWORD_AND_SALT';
```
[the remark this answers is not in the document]
xil222: I have a basic understanding of how server & client side work, but have some confusions here.
Which one are you referring to the previous one, src/auth.d.ts? It's exporting admin.auth, shouldn't it be the server side?

src/auth/user-import-builder.ts
```diff
@@ -370,6 +376,25 @@ export class UserImportBuilder {
     return importResult;
   }
 
+  /**
+   * Checks whether is a hashInputOrder-supported hashing algorithm.
+   * @param {HashAlgorithmType | undefined} hashAlgorithmType Hash algorithm type.
```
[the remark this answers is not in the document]
xil222: Do you mean for this function only? 
It's been a long time, but I remember I probably got an error for removing undefined or the whole thing.

src/auth/user-import-builder.ts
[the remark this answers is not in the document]
bojeil-google: `HashInputOrderType` is the client side value, whereas `PasswordHashOrderType` is the server side value. Maybe you can change to:

```
/** Server side values for password hash order types. **/
export type PasswordHashOrderServerType = 'SALT_AND_PASSWORD' | 'PASSWORD_AND_SALT';
```

src/auth/user-import-builder.ts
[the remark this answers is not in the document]
bojeil-google: Sorry, my suggestion was also incorrect. I am trying to be grammatically correct ("whether is" is not correct). Change to:
```
 * @return {boolean} Whether hash input order is supported.
```

src/auth/user-import-builder.ts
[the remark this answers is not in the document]
bojeil-google: Do this for all the functions you introduced. It won't complain about javadocs but it will complain if the type definitions are incorrect. The explicit types were needed when we couldn't easily determine them. In typescript, they are explicitly defined in the function definition.
The TS style guide is a [helpful reference](https://g3doc.corp.google.com/javascript/typescript/g3doc/styleguide.md?cl=head).

test/unit/auth/user-import-builder.spec.ts
bojeil-google: Hashing order is not supported for other algorithms too (eg. bcrypt, scrypt, etc)

xil222: This condition is based on the forloop for md5ShaPbkdfAlgorithms where 
const md5ShaPbkdfAlgorithms = ['MD5', 'SHA1', 'SHA256', 'SHA512', 'PBKDF_SHA1', 'PBKDF2_SHA256'];
In the origin code structure, algorithms like bcrypt and scrypt have their own defined sections for testing because they differ a lot from others. Maybe I'll try add one extra test each for these individual algorithms.

xil222: for your previous comment "test when hash input order is specified with algorithm that doesn't support it."

bojeil-google: Ah ok. The GitHub review UI was not helping. Can you just add a comment that hash input order is not supported for these algorithms?

xil222: sure.

test/unit/auth/user-import-builder.spec.ts
xil222: Sure, to be clear I am putting tests starting at line 383.
Since algorithms in hmacAlgorithms ['HMAC_SHA512', 'HMAC_SHA256', 'HMAC_SHA1', 'HMAC_MD5'] all support specifying hash order.

src/auth.d.ts
```diff
@@ -570,6 +570,8 @@ export namespace admin.auth {
     pageToken?: string;
   }
 
+  type HashInputOrderType = 'SALT_FIRST' | 'PASSWORD_FIRST';
```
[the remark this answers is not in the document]
xil222: I don't quite understand here. You're referring to the usage of HashInputOrderType but not this line right?
Cause what I saw is like removing "?" or "| null" along with the type.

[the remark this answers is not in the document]
bojeil-google: Hey @xil222, I don't refer specifically to this line, but in general, they are moving away from `auth.d.ts` and now are auto-generating the type definitions from the code. This PR may be helpful to understand these changes:
https://github.com/firebase/firebase-admin-node/pull/1009/files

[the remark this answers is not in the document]
xil222: It took me sometime but I am not sure I understand it correctly or PR did help.

1. I didn't see how it resolved auto-generating **type** definition. In the link above, the only reference to type is line 79 in src/auth/user-import-builder-internal.ts:
export type ValidatorFunction = (data: UploadAccountUser) => void;
It's exactly the same implementation as line 158 in src/auth/user-import-builder.ts
There's no sign of auto-generating type definition here.
What I can see in src/auth/user-import-builder-internal.ts doing is importing few interfaces exported in src/auth/user-import-builder.ts, but I didn't see how it do with autogenerating the type definitions.

2. In the link above, src/auth/user-import-builder-internal.ts was implemented in another branch which hasn't been merged into master branch yet. Even if my first point was wrong, that few lines of code edition could realize the auto-generating of type definitions, real implementation should be done only after that branch has been merged. I can't remove or take off the definition of this line.

3. I need more clarification on auto-generating type definitions. I can't see how the implementation could help with that.

Therefore, maybe it's better merge this PR into master for now.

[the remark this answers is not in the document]
bojeil-google: Ok it is fine to stick with the current implementation in master. Though I think if this ends up landing after the other branch lands, you may have to make the changes then.

@hiranya911 just a heads on this.

PR #938, opened 2020-07-09
fix: Upgraded Chai, Sinon and other test dependencies

hiranya911:
We are using several heavily outdated test dependencies, and this PR upgrades them to the latest versions:

* Chai 3.5 --> 4.2
* Sinon 4.5 --> 9.0
* Sinon-Chai 2.14 --> 3.0
* Chai-as-promised 6.0 --> 7.0

The biggest change stems from the Chai upgrade, where the latest versions of Chai do not allow `deep.equals` assertions on `Error` types. See https://github.com/chaijs/chai/issues/1009. Therefore I had to use `deep.include` in all the places where we directly compare `Error` objects.

The latest versions of Sinon also have better TS support which caused several TS compilation errors in tests. I've fixed those instances too.


test/unit/auth/auth.spec.ts
```diff
@@ -378,14 +378,16 @@ AUTH_CONFIGS.forEach((testConfig) => {
       });
 
       it('should be fulfilled given an app which returns null access tokens', () => {
-        getTokenStub = sinon.stub(ServiceAccountCredential.prototype, 'getAccessToken').resolves(null);
+        getTokenStub = sinon.stub(ServiceAccountCredential.prototype, 'getAccessToken')
+          .resolves(null as any);
```
nbegley, line 382: Are you able to use a specific type rather than `any` for these changes?

test/unit/auth/user-import-builder.spec.ts
```diff
@@ -861,9 +864,21 @@ describe('UserImportBuilder', () => {
...
+    function expectUserImportResult(result: UserImportResult, expected: UserImportResult): void {
+      expect(result.successCount).to.equal(expected.successCount);
+      expect(result.failureCount).to.equal(expected.failureCount);
+      expect(result.errors.length).to.equal(expected.errors.length);
+      result.errors.forEach((err, idx) => {
+        const want = expected.errors[idx];
+        expect(err.index).to.equal(want.index);
+        expect(err.error).to.deep.include(want.error);
+      });
+    }
```
nbegley, line 881: Consider extracting this to a common utility, since it's used in multiple files

test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -344,7 +344,8 @@ describe('RemoteConfigApiClient', () => {
         stubs.push(stub);
         const expected = new FirebaseRemoteConfigError('failed-precondition', message);
         return apiClient.validateTemplate(REMOTE_CONFIG_TEMPLATE)
-          .should.eventually.be.rejected.and.deep.equal(expected);
+          .should.eventually.be.rejectedWith(expected.message)
```
nbegley, line 347: It looks like this changes what's being tested - Why are we expecting a message to be thrown now, rather than an Error?

test/unit/auth/auth.spec.ts
hiranya911: We can but I don't see what benefit we get from such a change. Since we are basically testing for an invalid return value here, I think `any` is appropriate.

test/unit/remote-config/remote-config-api-client.spec.ts
hiranya911: `rejectedWith()` and `throws()` always test for error messages (or string patterns) only. Given that Chai can no longer compare `Error` objects (see the description of the PR), we have to assert on message and code properties separately now.

test/unit/auth/user-import-builder.spec.ts
hiranya911: Done. Exported from this module, and imported into the other.

test/unit/auth/auth-api-request.spec.ts
```diff
@@ -1761,12 +1764,23 @@ AUTH_REQUEST_HANDLER_TESTS.forEach((handler) => {
...
             throw new Error('Unexpected success');
           }, (error) => {
-            expect(error).to.deep.equal(expectedError);
+            expect(error).to.deep.include(expectedError);
             expect(stub).to.have.been.calledOnce.and.calledWith(
               callParams(path, method, userImportBuilder.buildRequest()));
           });
       });
 
+      function expectUserImportResult(result: UserImportResult, expected: UserImportResult): void {
```
lahirumaramba, line 1773: Should this be the function exported from `user-import-builder.spec.ts`?

test/unit/remote-config/remote-config-api-client.spec.ts
lahirumaramba: Just curious. Do you think we can use `.should.eventually.be.rejected.and.deep.include(error)` here?

test/unit/auth/auth-api-request.spec.ts
hiranya911: I think I forgot to push a local change. This should be gone now.

test/unit/remote-config/remote-config-api-client.spec.ts
hiranya911: Yes, that actually works since `expected` does not contain a stack trace. Updated this file.

PR #941, opened 2020-07-10
chore: update node-forge to ^0.9.1

sk-:
As commnented on https://github.com/firebase/firebase-admin-node/pull/934#discussion_r452226525

Given that node-forge is a 0.x version, npm considers it to be a breaking change, and hence not compatible with the current requirement.

node-forge is also transitively being used by firestore (direct dependency is google-p12-pem), and requiring version ^0.9.0, which means two copies of the library are currently needed. Switching to just one dependency would save 1.7Mib.

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3441,14 +3441,6 @@
       "optional": true,
       "requires": {
         "node-forge": "^0.9.0"
-      },
-      "dependencies": {
-        "node-forge": {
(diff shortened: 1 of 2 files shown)


sk-:
@hiranya911 Here is the PR updating `node-forge`

PR #944, opened 2020-07-14
Fix url fragment

matamatanot:
https://firebase.google.com/docs/reference/admin/node/admin.auth.Auth#verifyIdToken
https://firebase.google.com/docs/reference/admin/node/admin.auth.Auth#verifyidtoken


matamatanot:
@googlebot I signed it!

PR #948, opened 2020-07-15
Add prototype for Firestore module

MathBunny:
- Adds prototype for a Firestore module
   - Unfortunately, it requires manually re-exporting each type from Firestore individually

Next steps: explore if CI can determine when a type is not exported


MathBunny:
Not sure why GitHub isn't updating the PR with the new commit but I since added two things:

1. Invoked the unit tests by adding `firestore.spec.ts` to `index.ts` (they actually weren't running before :) )

2. Unit test for verifying if all the props from the cloudFirestore (@google-cloud/firestore) exist in the admin firestore (/src/firestore), with the exception of two whitelisted props.

```
describe('typing re-exports', () => {
    const whitelistedProps = new Set(['DocumentChange', 'default']);
    const adminProps = new Set();

    for (const key in adminFirestore) {
      adminProps.add(key);
    }

    for (const key in cloudFirestore) {
      expect(adminProps.has(key) || whitelistedProps.has(key.toString())).to.be.true;
    }
  });
```

PR #952, opened 2020-07-16
chore: Remove instances of XXX_SDK_VERSION_XXX by reading version from package.json

MathBunny:
### Summary
Removes instances of `XXX_SDK_VERSION_XXX` which were substituted by gulp at the build stage, and instead reads from `package.json` the value of `version`. 

### Test Plan
Added unit test to verify that the value is read from `package.json`.

Edit: I had a thought of perhaps adding an integration test to it or for the `example.test.ts`, however this is a util function which isn't supposed to be exposed to developers hence my initial reaction was against that.

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -19,8 +19,15 @@ import { ServiceAccountCredential, ComputeEngineCredential } from '../auth/crede
 
 import * as validator from './validator';
 
-const { version } = require('../../package.json'); // eslint-disable-line @typescript-eslint/no-var-requires
-export const SDK_VERSION = version;
+let sdkVersion: string;
+
+export function getSdkVersion(): string {
+  if (!sdkVersion) {
+    const { version } = require('../../package.json'); // eslint-disable-line @typescript-eslint/no-var-requires
+    sdkVersion = version;
+  }
+  return sdkVersion;
+}
 
 /**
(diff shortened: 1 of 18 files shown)


src/utils/index.ts
```diff
@@ -19,6 +19,9 @@ import { ServiceAccountCredential, ComputeEngineCredential } from '../auth/crede
 
 import * as validator from './validator';
 
+const { version } = require('../../package.json'); // eslint-disable-line @typescript-eslint/no-var-requires
```
hiranya911, line 22: Let's expose this as a function. That's more future proof, and also defers loading the `package.json` until absolutely necessary.

```
let sdkVersion: string;

export function getSdkVersion() {
  if (!sdkVersion) {
    // load from file
  }

  return sdkVersion;
}
```

test/unit/utils/index.spec.ts
```diff
@@ -34,6 +34,23 @@ interface Obj {
   [key: string]: any;
 }
 
+describe('SDK_VERSION', () => {
+  it('utils index should retrieve the SDK_VERSION from package.json', () => {
+    const mockVersionNumber = "MockVersion";
```
hiranya911, line 39: May be this test case can just read the version from `../../../package.json` and compare that to what's returned by `utils`.

PR #961, opened 2020-07-21
fix: Use object instead of Object in typings

MathBunny:
It is best-practice in TypeScript to use `object` instead of `Object`. While moving towards auto-generated typings, we will get this for free as the source was written correctly, however in the meantime this would make the move more transparent.

From my understanding this should be a safe move so-long as none of the developers use primitive types in place of Object (for example `Number`), but I don't think that is the case here. 

I provided more details in an internal document: go/firebase-admin-node-object

### DIFF
--- a/src/database.d.ts
+++ b/src/database.d.ts
@@ -422,7 +422,7 @@ export namespace admin.database {
     /**
      * @return A JSON-serializable representation of this object.
      */
-    toJSON(): Object | null;
+    toJSON(): object | null;
 
     /**
      * Extracts a JavaScript value from a `DataSnapshot`.
@@ -604,7 +604,7 @@ export namespace admin.database {
      * });
      * ```
      *
-     * @param values Object containing multiple values.
+     * @param values object containing multiple values.
      * @param onComplete An optional callback function that will
      *   be called when synchronization to the server has completed. The
      *   callback will be passed a single parameter: null for success, or an Error
@@ -612,7 +612,7 @@ export namespace admin.database {
      * @return Resolves when synchronization to the
      *   Database is complete.
      */
-    update(values: Object, onComplete?: (a: Error | null) => any): Promise<void>;
+    update(values: object, onComplete?: (a: Error | null) => any): Promise<void>;
   }
 
   type EventType = 'value' | 'child_added' | 'child_changed' | 'child_moved' | 'child_removed';
@@ -865,7 +865,7 @@ export namespace admin.database {
     off(
       eventType?: admin.database.EventType,
       callback?: (a: admin.database.DataSnapshot, b?: string | null) => any,
-      context?: Object | null
+      context?: object | null
     ): void;
 
     /**
@@ -987,8 +987,8 @@ export namespace admin.database {
     on(
       eventType: admin.database.EventType,
       callback: (a: admin.database.DataSnapshot, b?: string | null) => any,
-      cancelCallbackOrContext?: ((a: Error) => any) | Object | null,
-      context?: Object | null
+      cancelCallbackOrContext?: ((a: Error) => any) | object | null,
+      context?: object | null
     ): (a: admin.database.DataSnapshot | null, b?: string) => any;
 
     /**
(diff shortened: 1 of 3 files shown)


schmidt-sebastian:
I think we need actual TypeScript expertise here. I worry that returning `object` instead of `Object` is a breaking change: https://subscription.packtpub.com/book/application_development/9781789345575/3/ch03lvl1sec22/the-differences-between-object-and-object

See also https://www.typescriptlang.org/play/#code/FAYw9gdgzgLgBASygfQPICMBWBTEMBccGOecAvHBAK4A2NcAhlIxAJ4DcwSaWuMAdAAsmqAO4QACgCcwAB2xSYrABQBydAymqAlJ1CRYiFGF55CJkvArU6jZgzaduFvkJHjpchUrUatuuAB6QLgAYTAAW1kEGmwoABo4dCp4ABMwOIhVeChBMFE4KllECDhwVOwyyNlYmARIYGAAMyoIPHrSmAYAazieS2U-QmI+bTgAbzgAX2Au3pQRvGVnUxgAoJDwqJi4xKgwOBhBJCM4JoQIbEaWtrrIQ56+lyWholWxyZm5p9Xlhff2HANmFqjsEnB9odjswTudLsAgA

hiranya911:
From what I can surmise, the compiler is pretty lenient about this and it's generally a non breaking change (unless the implementation actually uses `Object` type).

## Case 1

```
function foo(): Object {
  return {};
}

let x: Object = foo();
```

Here the return type of foo can be changed to `object` with no compiler errors (Type annotation on `x` doesn't have to change).

## Case 2

```
function foo(param: Object) {

}

let x: Object = {};
foo(x);
``` 

Here the argument `param` type can be changed to `object` with no compiler errors (Type annotation on `x` doesn't have to change).

However, if somebody has passed a primitive value into `foo()` that would break:

```
foo('foo'); // will not compile if 'param' is changed to 'object`
```

But I think for the APIs in question, passing primitive values is an invalid usage anyway. So we can argue that this is bug fix for out typings?

schmidt-sebastian:
I am only worried about Case 1. Case 2 (changing the input type) seems like the right thing to do.

My only concern is that the code completion on `object` is empty, but on `Object` it is not (shows toString() and others). That being said, it compiles fine.

hiranya911:
The `Object` interface defines `hasOwnProperty()`, `toString()` and other methods:

```
interface Object {
    constructor: Function;
    toString(): string;
    toLocaleString(): string;
    valueOf(): Object;
    hasOwnProperty(v: PropertyKey): boolean;
    isPrototypeOf(v: Object): boolean;
    propertyIsEnumerable(v: PropertyKey): boolean;
}
```

That explains why the code completion is working. The `object` type is not an interface but the compiler does seem to allow calling `toString()` and other `Object` interface methods on it.

```
function foo(): object {
  return {};
}

// This compiles
let x = foo();
if (x.hasOwnProperty('bar')) {
  console.log(x.toString());
}
```

MathBunny:
Thanks for the concerns @schmidt-sebastian  and @hiranya911. I think the findings outlined so far align with what I found throughout my research shown the doc too (primarily focused on primitive types and conversion between `object` <-> `Object` return and parameter types), which is why I leaned towards this being a safe change.

-- -
The only other case I could find was where the `object` for some reason overrode a method defined in the `Object` interface and used a different return type (ex: `toString` and returns a `number`). In that case, you could not assign the now-valid `object` to an `Object`; however previously you couldn't return it anyway as an `Object` so I don't think there's any point considering this case.

bojeil-google:
The change looks fine. I just share the same concern as @schmidt-sebastian on the risk of introducing unexpected breaking changes to developers. It would be safer to rollout as a major version change but I leave it to @hiranya911 to make the call.

MathBunny:
To confirm the changes are non-breaking I asked on the internal TypeScript group and unfortunately received feedback that indeed it can be a breaking change for those who leverage type inferences. For example, performing:

```
function foo(): Object {
  // ...
}

let x = foo();
x = 3; 
```

If the return signature of `foo()` is `Object`, then the above would work since `x` would be implicitly receive the `Object` typing, however if we change `foo()` to return `object` then this no longer compiles. As such, I'm going to amend the PR to only change function parameters to `object`.

src/database.d.ts
```diff
@@ -1651,7 +1651,7 @@ export namespace admin.database.ServerValue {
    * });
    * ```
    */
-  const TIMESTAMP: Object;
+  const TIMESTAMP: object;
```
schmidt-sebastian, line 1654: We probably have to undo this too since it can be used in assignments.

PR #963, opened 2020-07-22
Allow RTDB to auto-generate typings, separate internal vs external APIs

MathBunny:
### Context
The goal is to bridge the gap between today's Admin SDK (manually curated `d.ts` files, nested namespaces) and a modularized SDK that has auto-generated typings.

This is the first milestone from: `go/firebase-node-auto-typing`.

### Goals for this PR
1. Allow for auto-generated typings for RTDB
2. Split the internal and external APIs
3. Still use the previous manually curated typings
4. Be a non-breaking change

To verify that the typings work as expected:
> export TYPE_GENERATION_MODE=auto && npm test && npm run integration && npm pack && ./.github/scripts/verify_package.sh firebase-admin-* 

~I also manually ran `npm pack` and `verify_script.sh`, however for `npm pack` it requires manual modification of the `package.json` so that it invokes the proper gulp build script.~

Several concerns:
1. Should I make the `FirebaseApp` split into an interface and class (similar to `App` today)?
2. I added an extra `admin` in the `index.ts`'s declared namespace for `database` (as compared to the Auth POC) because otherwise when installing `firebase-admin` the current `index.d.ts` wouldn't work. For example:

```
declare namespace admin.database {
  export import Database = _database.admin.database.Database;
  // this would have to be _database.database.Database; but right now it's _database.admin.database.Database
  // ...
}
```

3. ~It shouldn't be necessary to change the integration tests at all; this is because RTDB typings are using 'unknown' for `val()`'s return type right now. I asked Sebastian about his input on this.~

I imported the database types for the `DataSnapshot`, is that ok?


gulpfile.js
```diff
@@ -50,15 +51,19 @@ var paths = {
   ],
 
   build: 'lib/',
+
+  autoGeneratedTypings: ['lib/database.d.ts'],
```
hiranya911, line 55: This name is confusing. Perhaps `curatedTypings`?

gulpfile.js
```diff
@@ -50,15 +51,19 @@ var paths = {
...
 
 // Create a separate project for buildProject that overrides the rootDir
 // This ensures that the generated production files are in their own root
 // rather than including both src and test in the lib dir.
-var buildProject = ts.createProject('tsconfig.json', {rootDir: 'src'});
+var buildProject = ts.createProject('tsconfig.json', { rootDir: 'src' });
 
 var buildTest = ts.createProject('tsconfig.json');
 
+var buildAutogenTyping = ts.createProject('tsconfig.json', { declaration: true });
```
hiranya911, line 65: Pass `rootDir` option too.

gulpfile.js
```diff
@@ -93,6 +98,39 @@ gulp.task('compile_test', function() {
...
+ */
+gulp.task('compile_autogen_typing', function() {
+  return gulp.src(paths.src)
+    // Compile Typescript into .js and .d.ts files
+    .pipe(buildAutogenTyping())
+
+    // Exclude typings that are unintended (only RTDB and FirebaseApp are
+    // expected to produce typings).
+    .pipe(filter([
+      'lib/src/**/*.js',
```
hiranya911, line 112: If you pass `rootDir` option to the compiler `lib/src` can be shortened to `lib/`

gulpfile.js
```diff
@@ -93,6 +98,39 @@ gulp.task('compile_test', function() {
     .pipe(buildTest())
 });
 
+/**
```
hiranya911, line 101: Please add more details here. Describe what this task does end-to-end.

gulpfile.js
```diff
@@ -93,6 +98,39 @@ gulp.task('compile_test', function() {
...
+    // Exclude typings that are unintended (only RTDB and FirebaseApp are
+    // expected to produce typings).
+    .pipe(filter([
+      'lib/src/**/*.js',
+      'lib/src/**/*.d.ts',
+      '!lib/src/index.d.ts',
+      '!lib/src/default-namespace.d.ts',
+      '!lib/src/firebase-namespace.d.ts',
+      '!lib/src/firebase-service.d.ts',
+      '!lib/**/*-internal.d.ts',
```
hiranya911, line 118: Let's add some comments here to provide more context. We remove `-internal.d.ts` stuff because we don't ever want them exposed. We delete the other stuff *for now* since they are a work in progress.

src/firebase-app.ts
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { Credential, GoogleOAuthAccessToken, getApplicationDefault } from './auth/credential';
+import { getApplicationDefault } from './auth/credential';
```
hiranya911, line 17: Can we not touch the `firebase-app` module just yet? Is it possible to refactor only the database module in this PR and update the app module in a separate PR?

src/database/index.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { FirebaseApp } from '../firebase-app';
+import * as firebaseRtdbApi from '@firebase/database';
+import * as firebaseRtdbTypesApi from '@firebase/database-types';
```
hiranya911, line 19: We probably need to take an explicit dependency on this package.

src/database/index.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// We must define a namespace to make the typings work correctly. 
+// Otherwise `admin.database()` cannot be called like a function.
+/* eslint-disable @typescript-eslint/no-namespace */
+export namespace admin.database {
+  // See https://github.com/microsoft/TypeScript/issues/4336
+  /* eslint-disable @typescript-eslint/no-unused-vars */
+  // For context: github.com/typescript-eslint/typescript-eslint/issues/363
+  export import Database = firebaseRtdbApi.Database;
```
hiranya911, line 33: This should come from the database-types package as well I think. In fact all RTDB types should come from the database-types package.

gulpfile.js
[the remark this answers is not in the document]
hiranya911: Discussed offline.

src/database/index.ts
```diff
@@ -0,0 +1,42 @@
...
+import * as adminRtdbApi from './database';
+
+export function database(app: FirebaseApp): firebaseRtdbApi.Database {
+  return app.database();
+}
+
+// We must define a namespace to make the typings work correctly. 
+// Otherwise `admin.database()` cannot be called like a function.
+/* eslint-disable @typescript-eslint/no-namespace */
+export namespace admin.database {
```
hiranya911, line 29: Shouldn't this be just `export namespace database`? Or was this done as another temporary measure since we don't yet have the top-level barrel export?

src/database/database.ts
```diff
@@ -1,233 +1,60 @@
...
+import { Reference } from '@firebase/database';
 import { FirebaseApp } from '../firebase-app';
-import { FirebaseDatabaseError, AppErrorCodes, FirebaseAppError } from '../utils/error';
-import { FirebaseServiceInterface, FirebaseServiceInternalsInterface } from '../firebase-service';
 import { Database } from '@firebase/database';
 
-import * as validator from '../utils/validator';
-import { AuthorizedHttpClient, HttpRequestConfig, HttpError } from '../utils/api-request';
-import { getSdkVersion } from '../utils/index';
+export type EventType = 'value' | 'child_added' | 'child_changed' | 'child_moved' | 'child_removed';
```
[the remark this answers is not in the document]
MathBunny: Originally it was needed as part of the barrel exports in `database/index.ts`, however now that we are using `database-types` it is no longer needed as we can simply re-export it from there.

src/database/index.ts
MathBunny: Yep, the reason I did this is because we cannot perform a top-level barrel export yet and inside the existing `src/index.d.ts` file it expects an `admin.database` instead of simply `database`. If I change it to just `database`, then we'd need to change the `src/index.d.ts` as well to point to the correct namespace. 

I didn't want to change it for existing code as it reads from `src/database.d.ts`, though I considered adding a `.replace()` in the gulpfile for our tests to replace the code of `src/index.d.ts` at build time but it felt less clean than this solution.

So for example today's `src/index.d.ts` looks like this:

```
declare namespace admin.database {
  export import Database = _database.admin.database.Database;
  // this would have to be _database.database.Database; but right now it's _database.admin.database.Database
  // ... plus eventually we will just barrel export anyway
}
```

src/database/database.ts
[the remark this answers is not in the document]
MathBunny: Likewise we don't need the `ThenableReference` interface.

gulpfile.js
```diff
@@ -93,6 +98,47 @@ gulp.task('compile_test', function() {
     .pipe(buildTest())
 });
 
+/**
+ * Task used to verify TypeScript auto-generated typings. First, it builds
+ * the project generating both *.js and *.d.ts files. After, it removes
+ * declarations for all files terminating in -internal.d.ts because we do not
+ * want to expose internally exposed types to developers. As auto-generated
+ * typings are a work-in-progress, we remove the *.d.ts files for classes
```
hiranya911, line 106: s/classes/modules

src/database/database.ts
```diff
@@ -1,233 +1,75 @@
...
-   */
-  public delete(): Promise<void> {
-    for (const dbUrl of Object.keys(this.databases)) {
-      const db: Database = this.databases[dbUrl];
-      db.INTERNAL.delete();
-    }
-    return Promise.resolve(undefined);
-  }
-}
+import '@firebase/database';
```
hiranya911, line 16: Why was this needed?

src/database/index.ts
```diff
@@ -0,0 +1,52 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { FirebaseApp } from '../firebase-app';
+import * as firebaseAdmin from '../index';
+import * as firebaseRtdbApi from '@firebase/database';
+import * as firebaseRtdbTypesApi from '@firebase/database-types';
```
hiranya911, line 20: Ok, as per the recommendation from the JS SDK team, let's use `@firebase/database` for everything. Let's assume their implementation will sync up with the typings eventually (refer to the issues you've reported).

src/database/database.ts
```diff
@@ -1,233 +1,75 @@
...
-  private handleError(err: Error): Error {
-    if (err instanceof HttpError) {
-      return new FirebaseDatabaseError({
-        code: AppErrorCodes.INTERNAL_ERROR,
-        message: this.getErrorMessage(err),
-      });
-    }
-    return err;
-  }
+declare module '@firebase/database-types' {
```
hiranya911, line 48: See my other comment about loading everything from `@firebase/database`. In that case this won't be needed.

MathBunny:
Following up on our offline conversation, if the environment variable `TYPE_GENERATION_MODE` is set to `auto` then automatic type generation occurs.

src/database/database.ts
MathBunny: It's no longer needed but it might've been because I wasn't importing `@firebase/database` anymore in the `index.ts` file and I needed to perform the augmentation, so I did it mostly for the side-effects. I don't think it necessary

gulpfile.js
```diff
@@ -50,15 +51,19 @@ var paths = {
...
 
 // Create a separate project for buildProject that overrides the rootDir
 // This ensures that the generated production files are in their own root
 // rather than including both src and test in the lib dir.
-var buildProject = ts.createProject('tsconfig.json', {rootDir: 'src'});
+var buildProject = ts.createProject('tsconfig.json', { rootDir: 'src' });
 
 var buildTest = ts.createProject('tsconfig.json');
 
+var buildAutogenTyping = ts.createProject('tsconfig.json', { rootDir: 'src', declaration: true });
```
hiranya911, line 65: Let's use the same `buildProject` function in both modes:

```
const declaration = process.env.TYPE_GENERATION_MODE === 'auto';
var buildProject = ts.createProject('tsconfig.json', { rootDir: 'src', declaration });
```

gulpfile.js
```diff
@@ -93,6 +98,46 @@ gulp.task('compile_test', function() {
...
+/**
+ * Task used to verify TypeScript auto-generated typings. First, it builds
+ * the project generating both *.js and *.d.ts files. After, it removes
+ * declarations for all files terminating in -internal.d.ts because we do not
+ * want to expose internally exposed types to developers. As auto-generated
+ * typings are a work-in-progress, we remove the *.d.ts files for modules
+ * which we do not intend to auto-generate typings for yet.
+ */
+gulp.task('compile_autogen_typing', function() {
+  return gulp.src(paths.src)
```
hiranya911, line 110: Let's move this logic into the existing `compile` task.

```
let workflow = gulp.src(paths.src)
    // Compile Typescript into .js and .d.ts files
    .pipe(buildProject())

    // Add header
    .pipe(header(banner));

if (declaration) {
  workflow = workflow.pipe(filter(REDUNDANT_DECLARATION_FILES));
}

return workflow.pipe(gulp.dest(paths.build));
```

gulpfile.js
```diff
@@ -112,15 +157,28 @@ gulp.task('copyTypings', function() {
...
+
 gulp.task('compile_all', gulp.series('compile', 'copyDatabase', 'copyTypings', 'compile_test'));
 
 // Regenerates js every time a source file changes
 gulp.task('watch', function() {
-  gulp.watch(paths.src.concat(paths.test), {ignoreInitial: false}, gulp.series('compile_all'));
+  gulp.watch(paths.src.concat(paths.test), { ignoreInitial: false }, gulp.series('compile_all'));
 });
 
+// If the environment variable TYPE_GENERATION_MODE is set to auto then the
```
hiranya911, line 171: Once the above 2 changes are made, this part can remain unchanged.

src/database/database.ts
```diff
@@ -1,233 +1,56 @@
...
-        return `${intro}: ${body.error.trim()}`;
-      }
-    } catch {
-      // Ignore parsing errors
-    }
-
-    return `${intro}: ${err.response.text}`;
-  }
-}
+export interface ThenableReference extends Reference, Promise<Reference> { }
```
hiranya911, line 56: I don't think this makes much sense. The `push()` method is declared to return `Reference` in `@firebase/database`. We need to get that declaration updated. Trying to add it here doesn't really do anything.

MathBunny:
Putting this on pause for now until the JS SDK team follows up with some suggested changes to `@firebase/database-types`.

src/database/database.ts
```diff
@@ -1,233 +1,47 @@
...
-      db.INTERNAL.delete();
-    }
-    return Promise.resolve(undefined);
-  }
-}
-
-declare module '@firebase/database' {
-  interface Database {
+// Required to perform module augmentation to FirebaseDatabase interface.
+import '@firebase/database';
```
MathBunny, line 18: I couldn't fully reason about why this works.

If I do `import '@firebase/database-types'` then `npm test` fails:

```
Error: Cannot find module '@firebase/database-types'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
// ...
```

If I do no imports here, then `npm run integration` fails:
```
[10:41:29] Using gulpfile ~/Desktop/Programming-Projects/google-internship/firebase-admin-node/gulpfile.js
[10:41:29] Starting 'build'...
[10:41:29] Starting 'cleanup'...
[10:41:29] Finished 'cleanup' after 33 ms
[10:41:29] Starting 'compile'...
/Users/horatiulazu/Desktop/Programming-Projects/google-internship/firebase-admin-node/node_modules/@firebase/database/dist/index.d.ts(25,41): error TS2339: Property 'enableLogging' does not exist on type 'typeof import("@firebase/database-types")'.
/Users/horatiulazu/Desktop/Programming-Projects/google-internship/firebase-admin-node/node_modules/@firebase/database/dist/index.d.ts(26,32): error TS2694: Namespace '"@firebase/database-types"' has no exported member 'ServerValue'.
/Users/horatiulazu/Desktop/Programming-Projects/google-internship/firebase-admin-node/node_modules/@firebase/database/dist/index.d.ts(27,36): error TS2339: Property 'FirebaseDatabase' does not exist on type 'typeof import("@firebase/database-types")'.
src/database/index.ts(41,53): error TS2339: Property 'DataSnapshot' does not exist on type 'typeof import("@firebase/database-types")'.
src/database/index.ts(41,53): error TS2694: Namespace '"@firebase/database-types"' has no exported member 'DataSnapshot'.
src/database/index.ts(42,53): error TS2339: Property 'OnDisconnect' does not exist on type 'typeof import("@firebase/database-types")'.
src/database/index.ts(42,53): error TS2694: Namespace '"@firebase/database-types"' has no exported member 'OnDisconnect'.
// ...
```

When importing `@firebase/database-types`, I am not sure why this doesn't work.  

When performing no imports, my suspicion is that the module augmentation / declaration merging isn't working as expected and instead actually using it as a definition the `FirebaseDatabase` type (as-per the errors reported).

src/database/index.ts
MathBunny: Commenting for visibility, we're switching back now to `@firebase/database-types`.

src/database/database.ts
hiranya911: Discussed offline. Related fix in development at https://github.com/firebase/firebase-js-sdk/pull/3511

src/firebase-namespace.ts
```diff
@@ -31,7 +31,7 @@ import { Auth } from './auth/auth';
 import { MachineLearning } from './machine-learning/machine-learning';
 import { Messaging } from './messaging/messaging';
 import { Storage } from './storage/storage';
-import { Database } from '@firebase/database';
+import { FirebaseDatabase } from '@firebase/database-types';
```
hiranya911, line 34: Can we import this from `./database/database`?

test/unit/firebase-app.spec.ts
```diff
@@ -35,7 +35,7 @@ import { Messaging } from '../../src/messaging/messaging';
 import { MachineLearning } from '../../src/machine-learning/machine-learning';
 import { Storage } from '../../src/storage/storage';
 import { Firestore } from '@google-cloud/firestore';
-import { Database } from '@firebase/database';
+import { FirebaseDatabase } from '@firebase/database-types';
```
hiranya911, line 38: Import `Database` from `./database/database`?

test/unit/firebase-namespace.spec.ts
```diff
@@ -36,6 +36,7 @@ import {
   Reference,
   ServerValue,
 } from '@firebase/database';
+import { FirebaseDatabase } from '@firebase/database-types';
```
hiranya911, line 39: Here too

src/database/index.ts
```diff
@@ -0,0 +1,54 @@
...
+  // See https://github.com/typescript-eslint/typescript-eslint/issues/363
+  export import DataSnapshot = firebaseDbTypesApi.DataSnapshot;
+  export import OnDisconnect = firebaseDbTypesApi.OnDisconnect;
+  export import EventType = firebaseDbTypesApi.EventType;
+  export import Query = firebaseDbTypesApi.Query;
+  export import Reference = firebaseDbTypesApi.Reference;
+  export import ThenableReference = firebaseDbTypesApi.ThenableReference;
+  export import enableLogging = firebaseDbTypesApi.enableLogging;
+  export const ServerValue: firebaseDbTypesApi.ServerValue = sv;
+  // There is a known bug where @firebase/database-types FirebaseDatabase
```
hiranya911, line 50: Is this comment still relevant?

src/database/index.ts
```diff
@@ -0,0 +1,54 @@
...
+ * admin.database is used as the namespace name because we cannot barrel
+ * re-export the contents from @firebase/database-types, and we want it to
+ * match the namespacing in the re-export inside src/index.d.ts
+ */
+/* eslint-disable @typescript-eslint/no-namespace */
+export namespace admin.database {
+  // See https://github.com/microsoft/TypeScript/issues/4336
+  /* eslint-disable @typescript-eslint/no-unused-vars */
+  // See https://github.com/typescript-eslint/typescript-eslint/issues/363
+  export import DataSnapshot = firebaseDbTypesApi.DataSnapshot;
```
hiranya911, line 42: Use some kind of a logical ordering for these (alphabetical or similar).

MathBunny: Reordered based on ASCII, and kept the constant at the bottom

src/database/index.ts
MathBunny: No, thanks!

MathBunny:
Thanks for the feedback. Regarding the `interop-types`, turns out I don't actually need it, thanks for pointing that out.

PR #966, opened 2020-07-27
fix(auth): Fix several typing inconsistencies

MathBunny:
While working on the modularization of `instance-id` I came across several typing inconsistencies that I thought I'd bring up. Not sure if this is the best approach for all of them though.

##### 1) `auth-config.ts`
Observation: The `enabled` field should not be optional in `auth-config.ts`.

Source
https://github.com/firebase/firebase-admin-node/blob/54a58348ab7d8ee5e1ef4fd90cb68110a3e41391/src/auth/auth-config.ts#L168-L170

Typing

https://github.com/firebase/firebase-admin-node/blob/54a58348ab7d8ee5e1ef4fd90cb68110a3e41391/src/auth.d.ts#L960-L972

##### 2) `auth.ts`
Observation: There is a missing `uid` field in `DecodedIdToken`.

Source
https://github.com/firebase/firebase-admin-node/blob/54a58348ab7d8ee5e1ef4fd90cb68110a3e41391/src/auth/auth.ts#L108-L111

Typing
https://github.com/firebase/firebase-admin-node/blob/54a58348ab7d8ee5e1ef4fd90cb68110a3e41391/src/auth.d.ts#L531-L538

##### 3) `user-record.ts`
Observation: `displayName` in `MultiFactorInfo` doesn't have a `null` option for type.

Source
https://github.com/firebase/firebase-admin-node/blob/54a58348ab7d8ee5e1ef4fd90cb68110a3e41391/src/auth/user-record.ts#L149-L152

Typing
https://github.com/firebase/firebase-admin-node/blob/54a58348ab7d8ee5e1ef4fd90cb68110a3e41391/src/auth.d.ts#L80-L90

##### 4) `user-record.ts`
Observation: `enrolledFactors ` in `MultiFactor ` isn't read-only in types.

Source
https://github.com/firebase/firebase-admin-node/blob/54a58348ab7d8ee5e1ef4fd90cb68110a3e41391/src/auth/user-record.ts#L267-L268

Typing
https://github.com/firebase/firebase-admin-node/blob/54a58348ab7d8ee5e1ef4fd90cb68110a3e41391/src/auth.d.ts#L228-L233

### DIFF
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -108,6 +108,7 @@ export interface DecodedIdToken {
   sub: string;
   tenant?: string;
   [key: string]: any;
+  uid: string;
 }
 
 
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -107,8 +107,8 @@ export interface DecodedIdToken {
   picture?: string;
   sub: string;
   tenant?: string;
-  [key: string]: any;
   uid: string;
+  [key: string]: any;
 }
 
 
--- a/src/auth/user-record.ts
+++ b/src/auth/user-record.ts
@@ -213,7 +213,7 @@ export abstract class MultiFactorInfo {
     }
     utils.addReadonlyGetter(this, 'uid', response.mfaEnrollmentId);
     utils.addReadonlyGetter(this, 'factorId', factorId);
-    utils.addReadonlyGetter(this, 'displayName', response.displayName || null);
+    utils.addReadonlyGetter(this, 'displayName', response.displayName);
     // Encoded using [RFC 3339](https://www.ietf.org/rfc/rfc3339.txt) format.
     // For example, "2017-01-15T01:30:15.01Z".
     // This can be parsed directly via Date constructor.
--- a/src/auth/user-record.ts
+++ b/src/auth/user-record.ts
@@ -148,7 +148,7 @@ export enum MultiFactorId {
  */
 export abstract class MultiFactorInfo {
   public readonly uid: string;
-  public readonly displayName: string | null;
+  public readonly displayName: string;
   public readonly factorId: MultiFactorId;
   public readonly enrollmentTime: string;
 
@@ -265,7 +265,7 @@ export class PhoneMultiFactorInfo extends MultiFactorInfo {
 
 /** Class representing multi-factor related properties of a user. */
 export class MultiFactor {
-  public readonly enrolledFactors: ReadonlyArray<MultiFactorInfo>;
+  public enrolledFactors: MultiFactorInfo[];
 
   /**
    * Initializes the MultiFactor object using the server side or JWT format response.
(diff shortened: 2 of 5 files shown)


hiranya911:
Haven't dug into the details (I'll let @bojeil-google make a more informed assessment). My gut feeling is:

* 1-2 are trivial changes that we can do right away.
* 3 possibly requires updating the d.ts file to make `displayName` nullable instead of optional.
* 4 we should probably change the implementation to use a regular array. I don't know if we want to expose `ReadonlyArray` in our public typings at the moment.

MathBunny:
Thanks @hiranya911. Regarding the fourth point, indeed my approach right now is to change the implementation to use a regular array.

test/unit/auth/auth.spec.ts
```diff
@@ -161,6 +161,7 @@ function getDecodedIdToken(uid: string, authTime: Date, tenantId?: string): Deco
       sign_in_provider: 'custom', // eslint-disable-line @typescript-eslint/camelcase
       tenant: tenantId,
     },
+    uid: '123',
```
bojeil-google, line 164: This should match the `sub` field, eg. the value should be `uid`

test/unit/auth/auth.spec.ts
```diff
@@ -186,6 +187,7 @@ function getDecodedSessionCookie(uid: string, authTime: Date, tenantId?: string)
...
```
bojeil-google, line 190: Same here. This should match the `sub` field, eg. the value should be `uid`

src/auth/user-record.ts
```diff
@@ -148,7 +148,7 @@ export enum MultiFactorId {
  */
 export abstract class MultiFactorInfo {
   public readonly uid: string;
-  public readonly displayName: string | null;
+  public readonly displayName: string;
```
bojeil-google, line 151: Hmm, actually this is nullable. Check below: 
`utils.addReadonlyGetter(this, 'displayName', response.displayName || null);`

If we can't fix in the type definitions, then we should modify below to not be nullable.

MathBunny:
Thanks for the feedback @bojeil-google, I addressed your comments, PTAL :) 

Not sure about the unit tests and if the `displayName` change is a good idea.

bojeil-google:
@hiranya911 what are your thoughts on `displayName`? I personally prefer changing the d.ts to make this nullable, but we can also update the implementation as is currently done to not be nullable.

src/auth/auth.ts
```diff
@@ -108,6 +108,7 @@ export interface DecodedIdToken {
   sub: string;
   tenant?: string;
   [key: string]: any;
+  uid: string;
```
hiranya911, line 111: Move this up, so the `[key: string]` becomes the last entry.

src/auth/user-record.ts
hiranya911: I think changing typings from optional to nullable can be a breaking change. I think fixing the implementation to match the public API declaration is the right move here.

MathBunny:
Just as a note, I'm actually a fan of making it nullable, I just always steer more towards not changing typings. If possible I'd go that route

PR #967, opened 2020-07-28
Modify gulpfile to allow for autogenerated types per-service

MathBunny:
## Context
The goal is to bridge the gap between today's Admin SDK (manually curated d.ts files, nested namespaces) and a modularized SDK that has auto-generated typings. To accomplish this, we can go forward with gradually refactoring the codebase to allow for auto-generated typings.

The first step is to modify the gulpfile to accommodate toggling between auto-generated types and the manually curated ones in the builds. For more information, view the internal document: go/firebase-node-auto-typing.

## Overview
Depending on the value of `TYPE_GENERATION_MODE`:
- `TYPE_GENERATION_MODE = auto`: For all services that support auto-generated typings (i.e., have been refactored) the associated `d.ts` files are emitted. However, we automatically strip all files that have a suffix of `-internal.d.ts`, which is the now-canonical file representation for files containing internally used exported types.
- `TYPE_GENERATION_MODE != auto`: The former behavior (i.e., do not emit `d.ts` files and copy the manually curated typings).

## Test Plan
- Manual inspection
   - Ran it in the current state with `TYPE_GENERATION_MODE != auto`
       - All manually curated `d.ts` files were successfully copied
   - Added `!src/database.d.ts` to `paths.curatedTypings`, and removed `'!lib/database/*.d.ts',` from the compile step, and re-ran with `TYPE_GENERATION_MODE=auto`.
       - All manually curated `d.ts` files were copied except for `database.d.ts`, and inside `./database/` there was a `d.ts` file generated
- Unit, integration and sanity tests pass

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -55,9 +55,11 @@ var paths = {
   curatedTypings: ['src/**/*.d.ts'],
 };
 
-// Create a separate project for buildProject that overrides the rootDir
+// Create a separate project for buildProject that overrides the rootDir.
 // This ensures that the generated production files are in their own root
-// rather than including both src and test in the lib dir.
+// rather than including both src and test in the lib dir. Declaration
+// is used by TypeScript to determine if auto-generated typings should be
+// emitted.
 const declaration = process.env.TYPE_GENERATION_MODE === 'auto';
 var buildProject = ts.createProject('tsconfig.json', { rootDir: 'src', declaration });
 
@@ -91,7 +93,7 @@ gulp.task('compile', function() {
     // Add header
     .pipe(header(banner));
   
-  // Exclude typings that are unintended (currently includes all auto-generated
+  // Exclude typings that are unintended (currently excludes all auto-generated
   // typings, but as services are refactored to auto-generate typings this will
   // change). Moreover, all *-internal.d.ts typings should not be exposed to 
   // developers as it denotes internally used types.
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -55,6 +55,24 @@ var paths = {
   curatedTypings: ['src/*.d.ts'],
 };
 
+const TEMPORARY_TYPING_EXCLUDES = [
+  '!lib/default-namespace.d.ts',
+  '!lib/firebase-namespace.d.ts',
+  '!lib/firebase-app.d.ts',
+  '!lib/firebase-service.d.ts',
+  '!lib/auth/*.d.ts',
+  '!lib/database/*.d.ts',
+  '!lib/firestore/*.d.ts',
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -71,16 +77,50 @@ gulp.task('cleanup', function() {
   ]);
 });
 
+/**
```
MathBunny, line 80: Is there a preference for range comments (`/* */ vs //`)?

gulpfile.js
```diff
@@ -50,12 +51,17 @@ var paths = {
   ],
 
   build: 'lib/',
+
+  curatedTypings: ['src/**/*.d.ts'],
```
hiranya911, line 55: Just `src/*.d.ts` since curated typings are all directly in the src directory.

gulpfile.js
hiranya911: This file already seems to use `// comments`. Lets stick to the same convention.

gulpfile.js
```diff
@@ -71,16 +77,50 @@ gulp.task('cleanup', function() {
...
-    // Write to build directory
-    .pipe(gulp.dest(paths.build))
+    .pipe(header(banner));
+  
+  // Exclude typings that are unintended (currently excludes all auto-generated
+  // typings, but as services are refactored to auto-generate typings this will
+  // change). Moreover, all *-internal.d.ts typings should not be exposed to 
+  // developers as it denotes internally used types.
+  if (declaration) {
+    workflow = workflow.pipe(filter([
```
hiranya911, line 101: Let's make this a bit more readable:

```
const TEMPORARY_TYPING_EXCLUDES = [
  '!lib/default-namespace.d.ts',
  '!lib/firebase-namespace.d.ts',
  // other entries
];

const configuration = [
  'lib/**/*.js',
  'lib/**/*.d.ts',
  '!lib/**/*-internal.d.ts',
].concat(TEMPORARY_TYPING_EXCLUDES);
workflow = workflow.pipe(filter(configuration));
```

gulpfile.js
```diff
@@ -98,10 +96,7 @@ gulp.task('compile', function() {
   // change). Moreover, all *-internal.d.ts typings should not be exposed to 
   // developers as it denotes internally used types.
   if (declaration) {
-    workflow = workflow.pipe(filter([
-      'lib/**/*.js',
-      'lib/**/*.d.ts',
-      '!lib/**/*-internal.d.ts',
+    const TEMPORARY_TYPING_EXCLUDES = [
```
hiranya911, line 99: Move this closer to the top of the file. Around where curatedTypings is. As this effort moves forward we will be adding things to curatedTypings and removing things from this list.

PR #970, opened 2020-07-30
fix(auth): Make displayName optional for AuthProviderConfig typings

MathBunny:
### Observation
There is a discrepancy regarding `displayName` being a mandatory field in `AuthProviderConfig`.

Source
https://github.com/firebase/firebase-admin-node/blob/031edc2c3eb3b676b95063ce33bfc5673d7dc2bc/src/auth/auth-config.ts#L29-L34

Typings
https://github.com/firebase/firebase-admin-node/blob/031edc2c3eb3b676b95063ce33bfc5673d7dc2bc/src/auth.d.ts#L1062-L1082

### Solution

Assuming the typings are the source of truth I made it a mandatory field.

### DIFF
--- a/src/auth.d.ts
+++ b/src/auth.d.ts
@@ -1072,7 +1072,7 @@ export namespace admin.auth {
      * The user-friendly display name to the current configuration. This name is
      * also used as the provider label in the Cloud Console.
      */
-    displayName: string;
+    displayName?: string;
 
     /**
      * Whether the provider configuration is enabled or disabled. A user
--- a/src/auth/auth-config.ts
+++ b/src/auth/auth-config.ts
@@ -29,7 +29,7 @@ export interface AuthProviderConfigFilter {
 /** The base Auth provider configuration interface. */
 export interface AuthProviderConfig {
   providerId: string;
-  displayName?: string;
+  displayName: string;
   enabled: boolean;
 }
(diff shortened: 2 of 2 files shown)


hiranya911:
This might be a case of `AuthProviderConfig` type internally used for create/update requests as well. If not, LGTM.

MathBunny:
That's true ... thinking out loud, in that case I could set the `displayName` attribute of `AuthProviderConfig` to `string | undefined` and then in sub-classes such as `SAMLConfig` to `string`, maybe that would work?

bojeil-google:
I think this was a mistake. `displayName` is not required on a provider. It is definitely not required for update requests.

MathBunny:
Do you suggest I change the `d.ts` file then, and wait for a major version change? Or leave it as is?

hiranya911:
If the displayName is allowed to be `undefined` on some of the objects, then our current typings are incorrect. It should be corrected and the change can be treated as a bug fix.

bojeil-google:
Yeah, that would be great. I think we should update it in the typings to be optional. Thank you.

MathBunny:
Thanks for the responses, @bojeil-google PTAL :)

bojeil-google:
I already approved :)

PR #975, opened 2020-07-31
chore(rc): Move manual typings remote-config typings to separate folder

MathBunny:
## Context
As part of the initiative to auto-generate typings, it would be easier to verify that they are correctly generated if the remote-config manually curated typings were in another file. As part of the gulpfile build process, we could simply ignore the `d.ts` file.

## Solution
Moved the contents to a separate file, following the practice used by many of the other services today.

## Test Plan
1. Ran unit, integration and sanity tests.
2. Manually generated the docs, performed quick manual inspection, links work locally (excluding the index).


lahirumaramba:
Adding `release:stage` label to run integration tests.

MathBunny:
Thank you!

PR #981, opened 2020-08-06
chore: Adopt no-unused-vars-experimental for eslint to prevent incorrect linting errors

MathBunny:
## Context
There are several occasions where eslint incorrectly flags code claiming it violates the `@typescript-eslint/no-unused-vars` rule, for example while exporting namespaces that don't have any self-referencing in the typings files.

We can adopt [no-unused-vars-experimental](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unused-vars-experimental.md), which "leverages the TypeScript compiler's unused variable checks to report", and seemingly doesn't fall victim to the previous incorrect claims.

## Approach
Simply enable the experimental rule and disable the recommended rule. However, once doing that the following error would appear:

```
Error: Error while loading rule '@typescript-eslint/no-unused-vars-experimental':
 You have used a rule which requires parserServices to be generated. 
You must therefore provide a value for the "parserOptions.project" property 
for @typescript-eslint/parser.
```

To resolve this error, the following can be added:
```
"parserOptions": {
    "project": "./tsconfig.json"
  }
```

However, I realized later that not all files are mentioned in the current `tsconfig.json`. Apart from adding an `include` with a glob adding all src/test files, another option is to create a dedicated config just for linting, which is the approach taken here.

Another option was `createDefaultProgram: true`, but that has complexity `O(n^m)`, `n` = # files not included in config, `m` = average depth of the dependency tree of those respective files. I ran it with this setting and found it took ~1 minute to lint on my local computer so far from ideal.


.eslintrc.js
```diff
@@ -46,5 +47,10 @@ module.exports = {
         'allowHigherOrderFunctions': true
       }
     ],
-  }
+    '@typescript-eslint/no-unused-vars-experimental': 2,
+  },
+  // See github.com/typescript-eslint/typescript-eslint/issues/1320
```
hiranya911, line 52: Let's add a bit more details here.

```
// Required by the @typescript-eslint/no-unused-vars-experimental rule.
// We use a separate tsconfig file for linting to reduce the time complexity of the operation.
// See github.com/typescript-eslint/typescript-eslint/issues/1320
```

MathBunny:
Commenting for visibility the performance impact this has even without including `createDefaultProgram`:

```
Rule                                           | Time (ms) | Relative
:----------------------------------------------|----------:|--------:
@typescript-eslint/no-unused-vars-experimental |  3164.744 |    57.1%
indent                                         |  1709.654 |    30.8%
keyword-spacing                                |   137.947 |     2.5%
@typescript-eslint/camelcase                   |   120.434 |     2.2%
@typescript-eslint/type-annotation-spacing     |    49.622 |     0.9%
@typescript-eslint/no-empty-function           |    49.448 |     0.9%
no-regex-spaces                                |    25.818 |     0.5%
no-unexpected-multiline                        |    22.458 |     0.4%
no-mixed-spaces-and-tabs                       |    21.964 |     0.4%
object-curly-spacing                           |    17.590 |     0.3%
```
```
Rule                                           | Time (ms) | Relative
:----------------------------------------------|----------:|--------:
@typescript-eslint/no-unused-vars-experimental |  7122.536 |    60.3%
indent                                         |  3529.941 |    29.9%
@typescript-eslint/camelcase                   |   208.376 |     1.8%
@typescript-eslint/no-empty-function           |   205.101 |     1.7%
keyword-spacing                                |   128.697 |     1.1%
no-unexpected-multiline                        |   103.078 |     0.9%
no-regex-spaces                                |    72.967 |     0.6%
object-curly-spacing                           |    53.535 |     0.5%
no-useless-escape                              |    34.520 |     0.3%
no-control-regex                               |    29.842 |     0.3%
```

This is for `src`/`test` respectively. Moreover, we have the following benchmarks for former/PR/createDefaultProgram:

- `13.43 real        26.92 user         1.75 sys`
- `24.21 real        55.85 user         3.47 sys`
- `111.13 real       190.30 user        13.43 sys`

PR #982, opened 2020-08-06
Allow FCM to auto-generate typings, separate internal vs external APIs

MathBunny:
## Context
The goal is to bridge the gap between today's Admin SDK (manually curated d.ts files, nested namespaces) and a modularized SDK that has auto-generated typings.

This is the sixth milestone from: `go/firebase-node-auto-typing`.

## Goals for this PR
1. Allow to auto-generated typings for FCM
2. Split the internal and external APIs
3. Still use the previous manually curated typings
4. Be a non-breaking change

## Prerequisites
None.

## Test Plan
1. Unit tests: 
> `npm test`
2. Integration tests: 
> `export TYPE_GENERATION_MODE=nonauto && npm run integration && export TYPE_GENERATION_MODE=auto && npm run integration`
3. Sanity tests:  
> `export TYPE_GENERATION_MODE=nonauto && npm pack && ./.github/scripts/verify_package.sh firebase-* && export TYPE_GENERATION_MODE=auto && npm pack && ./.github/scripts/verify_package.sh firebase-*`


src/messaging/messaging-types.ts
```diff
@@ -142,107 +332,636 @@ export interface ApsAlert {
...
+
+  /**
+   * The package name of the application which the registration tokens must match
+   * in order to receive the message.
+   *
+   * **Default value:** None
+   */
   restrictedPackageName?: string;
-  [other: string]: any;
+  [key: string]: any | undefined;
```
MathBunny, line 934: Inconsistent here

src/messaging/index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * We must define a namespace to make the typings work correctly. Otherwise
+ * `admin.messaging()` cannot be called like a function. Temporarily,
+ * admin.messaging is used as the namespace name because we cannot barrel 
+ * re-export the contents from messsaging, and we want it to
+ * match the namespacing in the re-export inside src/index.d.ts
+ */
+/* eslint-disable @typescript-eslint/no-namespace */
+export namespace admin.messaging {
```
MathBunny, line 36: I missed exporting a bunch of interfaces here, will populate now

src/messaging/messaging-types-internal.ts
```diff
@@ -0,0 +1,600 @@
+/*!
```
hiranya911, line 1: Let's combine this into messaging-internal. We should look to keep *-internal files to a minimum.

MathBunny:
Thanks for the feedback! Yes, I missed adding it to the gulpfile, that is required. Moreover, after running the sanity tests with the updated gulpfile it also uncovered a missing interface that wasn't exported 👍🏼

src/messaging/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  // See https://github.com/microsoft/TypeScript/issues/4336
+  /* eslint-disable @typescript-eslint/no-unused-vars */
+  // See https://github.com/typescript-eslint/typescript-eslint/issues/363
+  export import AndroidConfig = messagingTypesApi.AndroidConfig;
```
MathBunny, line 41: The current `index.d.ts` splits this based on legacy types, should I do that here?

https://github.com/firebase/firebase-admin-node/blob/1984f1fd592e3cddbc3c5b27655628dd878aaebc/src/index.d.ts#L779-L781

hiranya911: Yes, that's a good idea.

gulpfile.js
```diff
@@ -56,7 +56,8 @@ var paths = {
     'src/*.d.ts',
     '!src/instance-id.d.ts',
     '!src/security-rules.d.ts',
-    '!src/project-management.d.ts'
+    '!src/project-management.d.ts',
+    '!src/messaging.d.ts'
```
hiranya911, line 60: Nit: Add trailing comma

PR #983, opened 2020-08-06
Firebase ML - adding AutoML support

ifielker:
Adding AutoML support

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1229,9 +1229,10 @@ declare namespace admin.machineLearning {
     /** The size of the model. */
     readonly sizeBytes: number;
 
-    // At most one of these two will be specified.
+    /** At most one of these two will be specified. */
     /** The URI from which the model was originally provided to Firebase. */
     readonly gcsTfliteUri?: string;
+    /** The AutoML URI from which the model was originally provided to Firebase. */
     readonly automlModel?: string;
   }
 
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1229,7 +1229,6 @@ declare namespace admin.machineLearning {
     /** The size of the model. */
     readonly sizeBytes: number;
 
-    /** At most one of these two will be specified. */
     /** The URI from which the model was originally provided to Firebase. */
     readonly gcsTfliteUri?: string;
     /** The AutoML URI from which the model was originally provided to Firebase. */
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1204,43 +1204,23 @@ declare namespace admin.securityRules {
 
 declare namespace admin.machineLearning {
   /**
-   * Interface representing options for listing Models.
+   * Firebase ML Model input objects
    */
-  interface ListModelsOptions {
-    /**
-     * An expression that specifies how to filter the results.
-     *
-     * Examples:
-     *
-     * ```
-     * display_name = your_model
-     * display_name : experimental_*
-     * tags: face_detector AND tags: experimental
-     * state.published = true
-     * ```
-     *
-     * See https://firebase.google.com/docs/ml-kit/manage-hosted-models#list_your_projects_models
-     */
-    filter?: string;
-
-    /** The number of results to return in each page. */
-    pageSize?: number;
-
-    /** A token that specifies the result page to return. */
-    pageToken?: string;
+  interface ModelOptionsBase {
+    displayName?: string;
+    tags?: string[];
   }
-
-  /** Response object for a listModels operation. */
-  interface ListModelsResult {
-    /** A list of models in your project. */
-    readonly models: Model[];
-
-    /**
-     * A token you can use to retrieve the next page of results. If null, the
-     * current page is the final page.
-     */
-    readonly pageToken?: string;
+  interface GcsTfliteModelOptions extends ModelOptionsBase {
+    tfliteModel: {
+      gcsTfliteUri: string;
+    };
   }
+  interface AutoMLTfliteModelOptions extends ModelOptionsBase {
+    tfliteModel: {
+      automlModel: string;
+    };
+  }
+  type ModelOptions = ModelOptionsBase | GcsTfliteModelOptions | AutoMLTfliteModelOptions;
 
   /**
    * A TensorFlow Lite Model output object
@@ -1249,26 +1229,10 @@ declare namespace admin.machineLearning {
     /** The size of the model. */
     readonly sizeBytes: number;
 
+    // At most one of these two will be specified.
     /** The URI from which the model was originally provided to Firebase. */
     readonly gcsTfliteUri?: string;
-  }
-
-  /**
-   * A Firebase ML Model input object
-   */
-  interface ModelOptions {
-    /** A name for the model. This is the name you use from your app to load the model. */
-    displayName?: string;
-
-    /** Tags for easier model management. */
-    tags?: string[];
-
-    /**
-     * An object containing the URI of the model in Cloud Storage.
-     *
-     * Example: `tfliteModel: { gcsTfliteUri: 'gs://your-bucket/your-model.tflite' }`
-     */
-    tfliteModel?: { gcsTfliteUri: string };
+    readonly automlModel?: string;
   }
 
   /**
@@ -1333,6 +1297,46 @@ declare namespace admin.machineLearning {
     /** Metadata about the model's TensorFlow Lite model file. */
     readonly tfliteModel?: TFLiteModel;
   }
+
+  /**
+   * Interface representing options for listing Models.
+   */
+  interface ListModelsOptions {
+    /**
+     * An expression that specifies how to filter the results.
+     *
+     * Examples:
+     *
+     * ```
+     * display_name = your_model
+     * display_name : experimental_*
+     * tags: face_detector AND tags: experimental
+     * state.published = true
+     * ```
+     *
+     * See https://firebase.google.com/docs/ml-kit/manage-hosted-models#list_your_projects_models
+     */
+    filter?: string;
+
+    /** The number of results to return in each page. */
+    pageSize?: number;
+
+    /** A token that specifies the result page to return. */
+    pageToken?: string;
+  }
+
+  /** Response object for a listModels operation. */
+  interface ListModelsResult {
+    /** A list of models in your project. */
+    readonly models: Model[];
+
+    /**
+     * A token you can use to retrieve the next page of results. If null, the
+     * current page is the final page.
+     */
+    readonly pageToken?: string;
+  }
+
 
   /**
    * The Firebase `MachineLearning` service interface.
--- a/src/machine-learning/machine-learning-api-client.ts
+++ b/src/machine-learning/machine-learning-api-client.ts
@@ -37,17 +37,28 @@ export interface StatusErrorResponse {
 }
 
 /**
- * A Firebase ML Model input object
+ * Firebase ML Model input objects
  */
-export interface ModelOptions {
+export interface ModelOptionsBase {
   displayName?: string;
   tags?: string[];
-
-  tfliteModel?: { gcsTfliteUri: string };
 }
+export interface GcsTfliteModelOptions extends ModelOptionsBase {
+  tfliteModel: {
+    gcsTfliteUri: string;
+  };
+}
+export interface AutoMLTfliteModelOptions extends ModelOptionsBase {
+  tfliteModel: {
+    automlModel: string;
+  };
+}
+export type ModelOptions = ModelOptionsBase | GcsTfliteModelOptions | AutoMLTfliteModelOptions;
+export type ModelUpdateOptions = ModelOptions & { state?: { published?: boolean }};
 
-export interface ModelUpdateOptions extends ModelOptions {
-  state?: { published?: boolean };
+export function isGcsTfliteModelOptions(options: ModelOptions): options is GcsTfliteModelOptions {
+  return (options as GcsTfliteModelOptions).tfliteModel !== undefined &&
+         (options as GcsTfliteModelOptions).tfliteModel.gcsTfliteUri !== undefined;
 }
 
 /** Interface representing listModels options. */
@@ -65,7 +76,9 @@ export interface ModelContent {
     readonly published?: boolean;
   };
   readonly tfliteModel?: {
-    readonly gcsTfliteUri: string;
+    readonly gcsTfliteUri?: string;
+    readonly automlModel?: string;
+
     readonly sizeBytes: number;
   };
 }
--- a/src/machine-learning/machine-learning-api-client.ts
+++ b/src/machine-learning/machine-learning-api-client.ts
@@ -57,8 +57,8 @@ export type ModelOptions = ModelOptionsBase | GcsTfliteModelOptions | AutoMLTfli
 export type ModelUpdateOptions = ModelOptions & { state?: { published?: boolean }};
 
 export function isGcsTfliteModelOptions(options: ModelOptions): options is GcsTfliteModelOptions {
-  return (options as GcsTfliteModelOptions).tfliteModel !== undefined &&
-         (options as GcsTfliteModelOptions).tfliteModel.gcsTfliteUri !== undefined;
+  const gcsUri = (options as GcsTfliteModelOptions)?.tfliteModel?.gcsTfliteUri;
+  return typeof gcsUri !== 'undefined'
 }
 
 /** Interface representing listModels options. */
--- a/test/integration/machine-learning.spec.ts
+++ b/test/integration/machine-learning.spec.ts
@@ -111,6 +111,7 @@ describe('admin.machineLearning', () => {
         tags: ['tagAutoml'],
         tfliteModel: { automlModel: 'this will be replaced below' }
       };
+      // AutoML models require verification. This takes between 20 and 60 seconds
       this.timeout(60000); // Allow up to 60 seconds for this test.
       return getAutoMLModelReference()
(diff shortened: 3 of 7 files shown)


src/machine-learning/machine-learning-api-client.ts
```diff
@@ -37,17 +37,28 @@ export interface StatusErrorResponse {
...
+  tfliteModel: {
+    gcsTfliteUri: string;
+  };
+}
+export interface AutoMLTfliteModelOptions extends ModelOptionsBase {
+  tfliteModel: {
+    automlModel: string;
+  };
+}
+export type ModelOptions = ModelOptionsBase | GcsTfliteModelOptions | AutoMLTfliteModelOptions;
```
hiranya911, line 56: Should ModelOptionsBase be included here? Is it ever used by itself except as one of the derived types?

src/machine-learning/machine-learning-api-client.ts
```diff
@@ -37,17 +37,28 @@ export interface StatusErrorResponse {
...
+export type ModelUpdateOptions = ModelOptions & { state?: { published?: boolean }};
```
hiranya911, line 57:
```
export interface ModelUpdateOptions extends ModelOptions {
  state?: { published?: boolean }
}
```

src/machine-learning/machine-learning-api-client.ts
```diff
@@ -37,17 +37,28 @@ export interface StatusErrorResponse {
...
 
-export interface ModelUpdateOptions extends ModelOptions {
-  state?: { published?: boolean };
+export function isGcsTfliteModelOptions(options: ModelOptions): options is GcsTfliteModelOptions {
+  return (options as GcsTfliteModelOptions).tfliteModel !== undefined &&
```
hiranya911, line 60:
```
const gcsUri = (options as GcsTfliteModelOptions)?.tfliteModel?.gcsTfliteUri;
return typeof gcsUri !== 'undefined';
```

test/integration/machine-learning.spec.ts
```diff
@@ -101,6 +105,58 @@ describe('admin.machineLearning', () => {
         });
     });
 
+    it('creates a new Model with valid AutoML TFLite ModelFormat', function () {
+      const modelOptions: admin.machineLearning.ModelOptions = {
+        displayName: 'node-integ-test-create-automl',
+        tags: ['tagAutoml'],
+        tfliteModel: { automlModel: 'this will be replaced below' }
+      };
+      this.timeout(60000); // Allow up to 60 seconds for this test.
```
hiranya911, line 114: Let's elaborate a bit as to why this takes so long.

test/integration/machine-learning.spec.ts
```diff
@@ -101,6 +105,58 @@ describe('admin.machineLearning', () => {
         });
     });
 
+    it('creates a new Model with valid AutoML TFLite ModelFormat', function () {
+      const modelOptions: admin.machineLearning.ModelOptions = {
+        displayName: 'node-integ-test-create-automl',
+        tags: ['tagAutoml'],
+        tfliteModel: { automlModel: 'this will be replaced below' }
```
hiranya911, line 112: Why not initialize this whole thing below?

test/integration/machine-learning.spec.ts
```diff
@@ -101,6 +105,58 @@ describe('admin.machineLearning', () => {
...
+              return model.waitForUnlocked(55000)
+                .then(() => {
+                  scheduleForDelete(model);
+                  verifyModel(model, modelOptions);
+                });
+            });
+        });
+    });
+
+    // Waiting for BE fix.
```
hiranya911, line 133: Let's implement this test case the way it should be, and just skip the whole thing for now by putting `xit.` on it.

test/integration/machine-learning.spec.ts
```diff
@@ -205,6 +261,33 @@ describe('admin.machineLearning', () => {
         });
     });
 
+    it('updates the automl model', function () {
+      this.timeout(60000); // Allow up to 60 seconds for this test.
+      return createTemporaryModel({
+        displayName: 'node-integ-test-update-automl'
+      }).then((model) => {
+        const modelOptions: admin.machineLearning.ModelOptions = {
+          tfliteModel: { automlModel: 'this will be replaced below' },
```
hiranya911, line 270: Initialize the whole thing below when you have the automl information.

test/integration/machine-learning.spec.ts
```diff
@@ -466,30 +549,34 @@ describe('admin.machineLearning', () => {
...
+  if ((expectedOptions as GcsTfliteModelOptions).tfliteModel?.gcsTfliteUri !== undefined) {
+    verifyGcsTfliteModel(model, (expectedOptions as GcsTfliteModelOptions));
+  } else if ((expectedOptions as AutoMLTfliteModelOptions).tfliteModel?.automlModel !== undefined) {
+    verifyAutomlTfliteModel(model, (expectedOptions as AutoMLTfliteModelOptions));
+  } else {
+    expect(model.validationError).to.equal('No model file has been uploaded.');
+  }
+}
+
+function verifyGcsTfliteModel(model: admin.machineLearning.Model, gcsTfliteOptions: GcsTfliteModelOptions): void {
```
hiranya911, line 578: expectedOptions

test/integration/machine-learning.spec.ts
```diff
@@ -499,3 +586,30 @@ function verifyTfliteModel(model: admin.machineLearning.Model, expectedGcsTflite
     expect(model.validationError).to.be.undefined;
   }
 }
+
+function verifyAutomlTfliteModel(model: admin.machineLearning.Model, automlTfliteOptions: AutoMLTfliteModelOptions): void {
```
hiranya911, line 590: expectedOptions

test/integration/machine-learning.spec.ts
```diff
@@ -499,3 +586,30 @@ function verifyTfliteModel(model: admin.machineLearning.Model, expectedGcsTflite
...
+}
+
+function getAutoMLModelReference(): Promise<string> {
+  try {
+    const { AutoMlClient } = require('@google-cloud/automl').v1;
+    const automl = new AutoMlClient()
+    const parent = automl.locationPath(projectId, 'us-central1');
+    return automl.listModels({ parent, filter:"displayName=admin_sdk_integ_test1" })
+      .then(([models]: [any]) => {
+        let modelRef = "";
```
hiranya911, line 605: Perhaps just: 
```
if (models) {
  return models[0].name;
}

return '';
```

test/integration/machine-learning.spec.ts
```diff
@@ -499,3 +586,30 @@ function verifyTfliteModel(model: admin.machineLearning.Model, expectedGcsTflite
...
+  const expectedAutomlReference = automlTfliteOptions.tfliteModel.automlModel;
+  expect(model.tfliteModel!.automlModel).to.equal(expectedAutomlReference);
+  expect(model.validationError).to.be.undefined;
+  expect(model.tfliteModel!.sizeBytes).to.not.be.undefined;
+  expect(model.modelHash).to.not.be.undefined;
+}
+
+function getAutoMLModelReference(): Promise<string> {
+  try {
+    const { AutoMlClient } = require('@google-cloud/automl').v1;
```
hiranya911, line 600: Use dynamic imports: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-4.html

```
return import('@google-cloud/automl')
  .then((automl) => {

  })
  .catch(() => return '')
```

src/index.d.ts
```diff
@@ -1204,43 +1204,23 @@ declare namespace admin.securityRules {
...
+    tfliteModel: {
+      gcsTfliteUri: string;
+    };
   }
+  interface AutoMLTfliteModelOptions extends ModelOptionsBase {
+    tfliteModel: {
+      automlModel: string;
+    };
+  }
+  type ModelOptions = ModelOptionsBase | GcsTfliteModelOptions | AutoMLTfliteModelOptions;
```
hiranya911, line 1223: Should ModelOptionsBase be included in this union?

src/index.d.ts
```diff
@@ -1249,26 +1229,10 @@ declare namespace admin.machineLearning {
     /** The size of the model. */
     readonly sizeBytes: number;
 
+    // At most one of these two will be specified.
```
hiranya911, line 1232: I think this should be stated in the type description so it gets included in the typedocs.

src/index.d.ts
```diff
@@ -1249,26 +1229,10 @@ declare namespace admin.machineLearning {
...
-    /** Tags for easier model management. */
-    tags?: string[];
-
-    /**
-     * An object containing the URI of the model in Cloud Storage.
-     *
-     * Example: `tfliteModel: { gcsTfliteUri: 'gs://your-bucket/your-model.tflite' }`
-     */
-    tfliteModel?: { gcsTfliteUri: string };
+    readonly automlModel?: string;
```
hiranya911, line 1235: Add description.

test/unit/machine-learning/machine-learning-api-client.spec.ts
```diff
@@ -19,8 +19,8 @@
 import * as _ from 'lodash';
 import * as chai from 'chai';
 import * as sinon from 'sinon';
-import { MachineLearningApiClient, ModelContent,
-  ListModelsOptions } from '../../../src/machine-learning/machine-learning-api-client';
+import { MachineLearningApiClient, ListModelsOptions,
```
hiranya911, line 22: Are we no longer using ModelContent type in the implementation?

test/unit/machine-learning/machine-learning-api-client.spec.ts
```diff
@@ -233,19 +288,38 @@ describe('MachineLearningApiClient', () => {
...
     const NAME_ONLY_MASK = ['displayName'];
+    const GCS_MASK = ['displayName', 'tfliteModel.gcsTfliteUri'];
+    const AUTOML_MASK = ['displayName', 'tfliteModel.automlModel'];
 
-    const invalidContent: any[] = [null, undefined];
-    invalidContent.forEach((content) => {
-      it(`should reject when called with: ${JSON.stringify(content)}`, () => {
-        return apiClient.updateModel(MODEL_ID, content, NAME_ONLY_MASK)
+    const NAME_ONLY_UPDATE_MASK = "updateMask=displayName";
+    const GCS_UPDATE_MASK = "updateMask=displayName,tfliteModel.gcsTfliteUri";
```
hiranya911, line 316: GCS_MASK_STRING to contrast with GCS_MASK

src/machine-learning/machine-learning-api-client.ts
ifielker: It is valid to specify a model with just displayName (and optional tags). So yes, it should be included.

src/machine-learning/machine-learning-api-client.ts
ifielker: Gives me an error: "An interface can only extend an object type or intersection of object types with statically known members."

src/machine-learning/machine-learning-api-client.ts
ifielker: done

test/integration/machine-learning.spec.ts
ifielker: done

test/integration/machine-learning.spec.ts
ifielker: Because it's being used in both the createModel and then one function level down in the verifyModel. I wouldn't have access to it in verifyModel if it was in the createModel scope I don't think? (Unless I'm wrong?)

src/index.d.ts
ifielker: yes

src/index.d.ts
ifielker: done (Also, Kevin will review/fix the docs before release, once we get this submitted)

src/index.d.ts
ifielker: done.

test/unit/machine-learning/machine-learning-api-client.spec.ts
ifielker: It was renamed to ModelOptions

test/unit/machine-learning/machine-learning-api-client.spec.ts
ifielker: That doesn't match the others. I renamed them all for further clarify.

ifielker: *clarity

test/integration/machine-learning.spec.ts
ifielker: It is implemented how it should be. I skipped it with it.skip is xit better?

src/machine-learning/machine-learning-api-client.ts
hiranya911: In that case why not just get rid of the union altogether, and use the type hierarchy in its place?

```
interface ModelOptions {

}

interface GcsTfliteModelOptions extends ModelOptions {

}
```

This avoids the `ModelOptionsBase` type. Any reason to not do this?

test/integration/machine-learning.spec.ts
ifielker: I won't be able to access it from the .then(updatedModel) function if it's a local variable in the then(automlRef) function...

src/machine-learning/machine-learning-api-client.ts
hiranya911: Turns out you cannot extend union types. But we can avoid this if we were to just drop the union type as described in my comment above.

test/integration/machine-learning.spec.ts
ifielker: done

test/integration/machine-learning.spec.ts
ifielker: done

src/index.d.ts
```diff
@@ -1249,26 +1229,11 @@ declare namespace admin.machineLearning {
     /** The size of the model. */
     readonly sizeBytes: number;
 
+    /** At most one of these two will be specified. */
```
hiranya911, line 1232: Doesn't look like correct TypeDoc syntax. This will have to be included in the description of the type.

test/integration/machine-learning.spec.ts
hiranya911:
```
return getAutoMLModelReference()
  .then((automlRef) => {
    if (!automlRef) {
      this.skip();
    }

    const modelOptions = {}; // init the whole thing here
    // create and verify
  })
```

test/integration/machine-learning.spec.ts
hiranya911: Then I don't think the description of the test is correct. I don't see any code that handles a rejection in the body of the test.

test/integration/machine-learning.spec.ts
hiranya911: Why not? It's still be in the scope of the `getAutoMLModelReference().then()`. So everything in that block should see it.

src/machine-learning/machine-learning-api-client.ts
```diff
@@ -37,17 +37,28 @@ export interface StatusErrorResponse {
 }
 
 /**
- * A Firebase ML Model input object
+ * Firebase ML Model input objects
  */
-export interface ModelOptions {
+export interface ModelOptionsBase {
```
hiranya911, line 42: Same here. Just call this ModelOptions and extend it where necessary. Get rid of the union types.

test/unit/machine-learning/machine-learning-api-client.spec.ts
hiranya911: It seems earlier these tests were using `ModelContent` in places where they should've used `ModelOptions`. In any case, this look ok.

src/index.d.ts
hiranya911: In that case lets remove this line for now.

src/machine-learning/machine-learning-api-client.ts
ifielker: The way it's currently implemented is what we agreed on in the design doc/CL that went with the design doc. I'm also pretty sure you are the one that came up with the 'type as a union' idea, and at the time we were discussing it a lot and coming up with how we wanted to implement it we both had a lot more recent context into why we might want to do it this way. (How would ModelUpdateOptions work with that structure?). I think we should leave it.

src/machine-learning/machine-learning-api-client.ts
ifielker: What would you extend? 
If ModelUpdateOptions is just an extension of the current ModelOptionsBase, then what about GcsTFLiteModelOptions? Would we need to do another definition of GcsTFLiteModelUpdateOptions that extends ModelUpdateOptions? (and the same for every new model options object we define (there will likely be many in future)). I prefer the current implementation.

test/integration/machine-learning.spec.ts
ifielker: Apparently I'm wrong. Done.

src/index.d.ts
ifielker: done

src/index.d.ts
ifielker: I removed the line. Kevin will fix the docs.

src/machine-learning/machine-learning-api-client.ts
ifielker: discussed offline.

PR #989, opened 2020-08-12
Firebase ML can gracefully ignore unknown fields in ModelResponse

ifielker:
Firebase ML can gracefully ignore unknown fields in ModelResponse.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -573,7 +573,7 @@
     },
     "@types/firebase-token-generator": {
       "version": "2.0.28",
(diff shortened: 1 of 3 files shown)


lahirumaramba:
@ifielker @hiranya911 We can including this in the upcoming release. Does this change contain release-notes?
Thanks!

ifielker:
There are no release notes for this. Users are not affects, and it doesn't fix any issues that users have seen. It's future proofing so users don't have any issues going forward.

PR #993, opened 2020-08-14
fix(auth): Address several auth typing inconsistencies

MathBunny:
During my work for refactoring auth to auto-generate typings I came across several inconsistencies in source and typing files. I'm not sure how to best address them or which require changes, but I did an attempt here.

#### 1. `auth.ts`
Observation: The `tenant` field is defined at the root of the object in the source while it's in the `firebase` scope in the typings.

Source:
https://github.com/firebase/firebase-admin-node/blob/5e6ebfa57649c425961351fbf772df4fead4f2f8/src/auth/auth.ts#L88-L112

Typing:
https://github.com/firebase/firebase-admin-node/blob/e415188c700a9ba8c2f3eac4fbf09ce598d095b4/src/auth.d.ts#L456-L491

#### 2. `auth-config.ts`
Observation: The `SAMLAuthProviderConfig` contains an `enableRequestSigning` in the source which is absent in typings.

Source:
https://github.com/firebase/firebase-admin-node/blob/5e6ebfa57649c425961351fbf772df4fead4f2f8/src/auth/auth-config.ts#L42-L50

Typing:
https://github.com/firebase/firebase-admin-node/blob/e415188c700a9ba8c2f3eac4fbf09ce598d095b4/src/auth.d.ts#L1139-L1177

As an extension, a similar pattern can be found here for `SAMLAuthProviderConfig`:
https://github.com/firebase/firebase-admin-node/blob/5e6ebfa57649c425961351fbf772df4fead4f2f8/src/auth/auth-config.ts#L118-L128

#### 3. `user-import-builder.ts`
Observation: The `tenantId` for `UserImportRecord` is nullable in typings but not source.

Source:
https://github.com/firebase/firebase-admin-node/blob/5e6ebfa57649c425961351fbf772df4fead4f2f8/src/auth/user-import-builder.ts#L64-L82

Typing:
https://github.com/firebase/firebase-admin-node/blob/e415188c700a9ba8c2f3eac4fbf09ce598d095b4/src/auth.d.ts#L816-L823

#### 4. `auth-config.ts`

Observation: As a consequence from the third point, the `tenantId` field is nullable in one case so it should be propagated.

Source:
https://github.com/firebase/firebase-admin-node/blob/5e6ebfa57649c425961351fbf772df4fead4f2f8/src/auth/user-import-builder.ts#L95-L118

https://github.com/firebase/firebase-admin-node/blob/5e6ebfa57649c425961351fbf772df4fead4f2f8/src/auth/user-import-builder.ts#L216-L229

This would be a compilation error:

> Type 'string | null | undefined' is not assignable to type 'string | undefined'.
  Type 'null' is not assignable to type 'string | undefined'.

#### 5. `user-record.ts`
The typings have `displayName` and `enrollmentTime` as optional while not in the source. 

Source:
https://github.com/firebase/firebase-admin-node/blob/5e6ebfa57649c425961351fbf772df4fead4f2f8/src/auth/user-record.ts#L149-L153

Typing:
https://github.com/firebase/firebase-admin-node/blob/e415188c700a9ba8c2f3eac4fbf09ce598d095b4/src/auth.d.ts#L80-L106

### DIFF
--- a/src/auth.d.ts
+++ b/src/auth.d.ts
@@ -1174,7 +1174,6 @@ export namespace admin.auth {
      * configuration.
      */
     callbackURL?: string;
-
   }
 
   /**
--- a/src/auth.d.ts
+++ b/src/auth.d.ts
@@ -820,7 +820,7 @@ export namespace admin.auth {
      * When not provided in an `admin.auth.TenantAwareAuth` context, the user is uploaded
      * to the tenant corresponding to that `TenantAwareAuth` instance's tenant ID.
      */
-    tenantId?: string | null;
+    tenantId?: string;
 
     /**
      * The user's multi-factor related properties.
@@ -1175,10 +1175,6 @@ export namespace admin.auth {
      */
     callbackURL?: string;
 
-    /**
-     * The ability for SAML provider to perform request signing.
-     */
-    enableRequestSigning?: boolean;
   }
 
   /**
@@ -1263,12 +1259,6 @@ export namespace admin.auth {
      * configuration's value is not modified.
      */
     callbackURL?: string;
-
-    /**
-     * Whether the SAML provider can sign requests. If not provided, the existing
-     * configuration's value is not modified.
-     */
-    enableRequestSigning?: boolean;
   }
 
   /**
--- a/src/auth.d.ts
+++ b/src/auth.d.ts
@@ -1174,6 +1174,11 @@ export namespace admin.auth {
      * configuration.
      */
     callbackURL?: string;
+
+    /**
+     * The ability for SAML provider to perform request signing.
+     */
+    enableRequestSigning?: boolean;
   }
 
   /**
@@ -1258,6 +1263,12 @@ export namespace admin.auth {
      * configuration's value is not modified.
      */
     callbackURL?: string;
+
+    /**
+     * Whether the SAML provider can sign requests. If not provided, the existing
+     * configuration's value is not modified.
+     */
+    enableRequestSigning?: boolean;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


hiranya911:
1. Typings are correct the way this is currently implemented. Perhaps we should just drop that field in the implementation type.
2. I believe `enableRequestSigning` is not publicly exposed yet. So typings are correct. Not sure how to handle this case though.
3., 4. I don't think this field can be nullable. We should probably update the typings.
5. Typings should be correct.

src/auth.d.ts
```diff
@@ -1174,6 +1174,11 @@ export namespace admin.auth {
      * configuration.
      */
     callbackURL?: string;
+
+    /**
+     * The ability for SAML provider to perform request signing.
+     */
+    enableRequestSigning?: boolean;
```
bojeil-google, line 1181: We don't currently support this. I think we should remove it. I think we need the flexibility to have internal implementations not exposed externally.  Many SDKs with auto-generated references already have the ability to set visibility on some APIs and attributes.

src/auth.d.ts
```diff
@@ -1258,6 +1263,12 @@ export namespace admin.auth {
      * configuration's value is not modified.
      */
     callbackURL?: string;
+
+    /**
+     * Whether the SAML provider can sign requests. If not provided, the existing
+     * configuration's value is not modified.
+     */
+    enableRequestSigning?: boolean;
```
bojeil-google, line 1271: Same here

src/auth.d.ts
hiranya911: FWIW, there's no built-in mechanism in TypeScript to hide members of an interface. Any library that's doing so is either using third-party tools or some sort of a hack. So we will have to do the same here (@MathBunny the undocumented `--stripInternal` flag is probably our best bet here).

MathBunny: Thanks, I agree. I verified the `@internal` option and it worked as intended (updated the respective PR).

PR #992, opened 2020-08-14
Allow Auth to auto-generate typings, separate internal vs external APIs

MathBunny:
## Context
The goal is to bridge the gap between today's Admin SDK (manually curated d.ts files, nested namespaces) and a modularized SDK that has auto-generated typings.

This is the eleventh milestone from: `go/firebase-node-auto-typing`.

## Goals for this PR
1. Allow to auto-generated typings for auth
2. Split the internal and external APIs
3. Still use the previous manually curated typings
4. Be a non-breaking change

## Prerequisites
None.

## Test Plan
1. Unit tests: 
> `npm test`
2. Integration tests: 
> `export TYPE_GENERATION_MODE=nonauto && npm run integration && export TYPE_GENERATION_MODE=auto && npm run integration`
3. Sanity tests:  
> `export TYPE_GENERATION_MODE=nonauto && npm pack && ./.github/scripts/verify_package.sh firebase-* && export TYPE_GENERATION_MODE=auto && npm pack && ./.github/scripts/verify_package.sh firebase-*`

PR #994, opened 2020-08-14
auth: Add modularization progress for ActionCodeSettingsBuilder, TokenVerifier (1/N)

MathBunny:

This is the covers: gulpfile, action-code-settings-builder, token-verifier


hiranya911:
@bojeil-google this is part of the typing auto-generation work we are doing. Basically we are moving all the internal-only types to files named `*-internal.ts`. We are also copying the API documentation from the existing d.ts file into the source files. @MathBunny  is going to do this for Auth modules in multiple steps.

PR #998, opened 2020-08-18
auth: Add modularization progress for Identifier, Tenant, TenantManager, TokenGenerator (3/N)

MathBunny:

This covers: Identifier, Tenant, TenantManager, TokenGenerator


src/auth/tenant.ts
```diff
@@ -71,6 +112,7 @@ export class Tenant {
    * @param {TenantOptions} tenantOptions The properties to convert to a server request.
    * @param {boolean} createRequest Whether this is a create request.
    * @return {object} The equivalent server request.
+   * @internal
```
MathBunny, line 115: I wasn't sure about the best approach for these two methods. They're not supposed to be publicly exposed, but if I pulled them out it might eliminate the logical grouping it has with `Tenant`. I also considered doing an interface/implementation split here. Any thoughts?

src/auth/tenant-manager.ts
```diff
@@ -17,18 +17,22 @@
 import { AuthRequestHandler } from './auth-api-request';
 import { FirebaseApp } from '../firebase-app';
 import { TenantAwareAuth } from './auth';
-import {
-  Tenant, TenantServerResponse, ListTenantsResult, TenantOptions,
-} from './tenant';
+import { Tenant, ListTenantsResult } from './tenant';
+import { TenantServerResponse } from './tenant-internal';
+import { TenantOptions } from './tenant';
```
hiranya911, line 22: Merge with previous import from `./tenant`

src/auth/tenant.ts
```diff
@@ -18,51 +18,92 @@ import * as validator from '../utils/validator';
 import { deepCopy } from '../utils/deep-copy';
 import { AuthClientErrorCode, FirebaseAuthError } from '../utils/error';
 import {
-  EmailSignInConfig, EmailSignInConfigServerRequest, EmailSignInProviderConfig,
-  MultiFactorConfig, MultiFactorAuthServerConfig, MultiFactorAuthConfig,
-  validateTestPhoneNumbers,
+  validateTestPhoneNumbers, EmailSignInConfig, MultiFactorAuthConfig,
 } from './auth-config';
+import { TenantServerResponse, TenantOptionsServerRequest } from './tenant-internal';
+import { EmailSignInProviderConfig, MultiFactorConfig  } from './auth-config';
```
hiranya911, line 24: Merge with previous import

src/auth/tenant.ts
hiranya911: Looking at their functionality, I think it's ok to pull them from this class. We can put them in `tenant-internal` as top-level methods or as static methods of a new `TenantUtils` class.

src/index.d.ts
```diff
@@ -857,6 +857,7 @@ declare namespace admin.remoteConfig {
   export import InAppDefaultValue = _remoteConfig.admin.remoteConfig.InAppDefaultValue;
   export import RemoteConfigParameterValue = _remoteConfig.admin.remoteConfig.RemoteConfigParameterValue;
   export import Version = _remoteConfig.admin.remoteConfig.Version;
+  export import ListVersionsOptions = _remoteConfig.admin.remoteConfig.ListVersionsOptions;
```
hiranya911, line 860: Let's skip this change. It's already in master branch.

MathBunny:
Thanks for the feedback, PTAL :)

PR #999, opened 2020-08-18
Add prototype for API extractor in CI

MathBunny:
Allows for verifying API surface changes as part of continuous integration changes. Some notes:
- All JSDoc linting needs to be disabled, otherwise the exit code becomes non-zero
- Ideally, users build and it updates the file automatically and then the tests will discover that it's not updated

The key insight into making this work is that `--local` will automatically updated the report file while omitting `--local` will make `api-extractor` return a non-zero status code if the report file in version control is not equal to what it would generate with the current code. 

This prevents anyone from accidentally forgetting to add the file to their version control so it skips review by accident.


MathBunny:
### Demonstration

`853a862` => I attempted to add a new class to the API surface, but forgot to commit the report file to version control. This fails CI because I'd change the API surface without formal approval.

`3d6c1cf` => I re-run `npm run build` which regenerates the report file. CI passes because the file will be included in code review.

hiranya911:
This was implemented and integrated to the current repo in #1081

PR #1007, opened 2020-08-21
(WIP): Add ability to generate documentation from auto-generated typings

MathBunny:
### Summary
Adds the ability to generate documentation web-pages from the auto-generated typings.

Currently, the repository is an intermediate state where all but some of the services have the ability to auto-generate typings (namely auth, machine-learning and storage cannot). In the current TypeDoc setup, we cannot generate the documentation web-pages when relying only on the auto-generated `d.ts` files.

Note: This is still WIP.

### Background
The first observation is that the generated filenames were incorrect; more specifically the namespace prefix was missing (i.e., `credential.html` vs `admin.credential.credential.html`).

Formerly, `modules` mode for TypeDoc was attempted but it was found that some of the generated `.html` files had different composition, so another approach is taken here. 

In short, the purpose of choosing `modules` is the generated typings from TypeDoc doesn't including the namespace prefix in the filename, but by switching to `modules` the folder name is prefixed allowing us to interpolate, as here: 

https://github.com/firebase/firebase-admin-node/blob/ebf660dd3de311ca0af0421beda8ee52dafa1f92/docgen/generate-docs.js#L293-L373

### Approach

This PR takes the approach of using TypeDoc's `file` mode as it is using today. Now, the `src` folder is traversed and the barrel export files are introspected to determine the namespace prefix instead. 

The technique chosen is to build an index mapping type-names to their parent folder name based on `src`. The underlying assumption for this is that the folder-name corresponds to the namespace (does not apply for `credential` but I addressed it in another PR) and type names are unique (does not apply for `query` so far, but this can be manually patched).

---

### Additional Issues

1. Hyperlinks are broken for links generated by TypeDoc (can be resolved by using previously constructed index).
2. Interfaces are incorrectly labelled as `implements: any` (can be resolved via regex)
      - Incorrect header  as it has a missing namespace prefix
      - `INTERNAL` should not appear (can be resolved via regex)
      - `Accessors` is actually `Properties`
      - Missing top-level documentation example

| Modularized (Local)  | Production [(Devsite)](https://firebase.google.com/docs/reference/admin/node/admin.messaging.Messaging)  |
|---|---|
|   |   |

3. `app: FirebaseApp` vs `app: _admin.app.App` (can be resolved via regex)
4. Even when re-exporting classes as interfaces, constructors still appear (can be resolved via `@hidden`)
5. When using `export import` pattern, the exported types appear as variables (can be resolved via regex)
      - Minimal reproducible example: [MathBunny/typedoc-namespace-bug-demo](https://github.com/MathBunny/typedoc-namespace-bug-demo)
      - GitHub issue: [TypeStrong/typedoc#1353](https://github.com/TypeStrong/typedoc/issues/1353)

| Modularized (Local)  | Production [(Devsite)](https://firebase.google.com/docs/reference/admin/node/admin.messaging)  |
|---|---|
|   |   |

6. The version number needs to be updated for `TypeDoc` in order for `export import` to even appear (went up from `0.15.0` to `0.18.0`)
      - Manually changed one of the flags to hide private members as they appear on doc pages as of `0.16.0`
7. Different behavior for inheriting from non-exposed types. 
For example:

| Modularized (Local)  | Production [(Devsite)](https://firebase.google.com/docs/reference/admin/node/admin.messaging.MulticastMessage)  |
|---|---|
|   |   |

8. RTDB types are not exported (similar to Firestore).

Several other minor fixes:
1. Resolved the broken `.html` link on `index.html` for service pages when run locally


MathBunny:
Several notes:
1. Added `@hidden` to all the `INTERNAL` field variables and removed the associated regex expressions for removing them.
2. Changed the algorithm for building the namespace index, instead now the `toc.yaml` is leveraged rather than observing the barrel exports in `src` folders
3. The links are broken in the service overview pages under variables, but I thought of not touching it yet until the bug is fixed on TypeDoc.

hiranya911:
Superseded by #1072

PR #1010, opened 2020-08-24
auth: Add modularization progress for UserRecord (6/N)

MathBunny:

This covers: `UserRecord`

### DIFF
--- a/src/auth/multi-factor.ts
+++ b/src/auth/multi-factor.ts
@@ -1,0 +1,144 @@
+/*!
+ * Copyright 2020 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as utils from '../utils';
+import { AuthClientErrorCode, FirebaseAuthError } from '../utils/error';
+import { MultiFactorId, MultiFactorInfoResponse } from './multi-factor-internal';
+
+/**
+ * Interface representing the common properties of a user enrolled second factor.
+ */
+export abstract class MultiFactorInfo {
+  /**
+   * The ID of the enrolled second factor. This ID is unique to the user.
+   */
+  public readonly uid: string;
+
+  /**
+   * The optional display name of the enrolled second factor.
+   */
+  public readonly displayName?: string;
+
+  /**
+   * The type identifier of the second factor. For SMS second factors, this is `phone`.
+   */
+  public readonly factorId: MultiFactorId;
+
+  /**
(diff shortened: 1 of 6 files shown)


src/auth/user-record-internal.ts
```diff
@@ -0,0 +1,81 @@
...
+  MultiFactorInfo, PhoneMultiFactorInfo
+} from './user-record';
+  
+export class MultiFactorInfoUtils {
+  /**
+   * Initializes the MultiFactorInfo associated subclass using the server side.
+   * If no MultiFactorInfo is associated with the response, null is returned.
+   *
+   * @param response The server side response.
+   * @constructor
```
hiranya911, line 27: Not a constructor

src/auth/user-record-internal.ts
```diff
@@ -0,0 +1,81 @@
...
+      // Ignore error.
+    }
+    return multiFactorInfo;
+  }
+}
+
+/** Enums for multi-factor identifiers. */
+export enum MultiFactorId {
+  Phone = 'phone',
+}
```
hiranya911, line 44: Add an empty line after this.

src/auth/user-record-internal.ts
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { 
+  MultiFactorInfo, PhoneMultiFactorInfo
```
hiranya911, line 18: How about moving these types into a new module (say `multi-factor`) and break the cyclic dependency?

MathBunny: Thanks, good point, though had to pull in a few more members to a `multi-factor-internal` to get it to be acyclic.

src/auth/user-record-internal.ts
```diff
@@ -0,0 +1,72 @@
...
+    // Only PhoneMultiFactorInfo currently available.
+    try {
+      multiFactorInfo = new PhoneMultiFactorInfo(response);
+    } catch (e) {
+      // Ignore error.
+    }
+    return multiFactorInfo;
+  }
+}
+
```
hiranya911, line 42: Remove extra space

src/auth/multi-factor.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   * The optional display name of the enrolled second factor.
+   */
+  public readonly displayName?: string;
+
+  /**
+   * The type identifier of the second factor. For SMS second factors, this is `phone`.
+   */
+  public readonly factorId: MultiFactorId;
```
hiranya911, line 38: Shouldn't `MultiFactorId` type be exposed?

MathBunny: @bojeil-google In `auth.d.ts` we have `factorId` as `string` while here we expose an enum. Do you have any thoughts on the best way forward?

bojeil-google: Either way is fine. Currently, this can only have the value `phone`. But in the future we will add new types.
I think ideally an enum is preferred.

MathBunny:
Note that afaik it's a breaking change if we use enum instead of string in the typings.

hiranya911:
Let's fix this in a non-breaking way. Changing the type to an enum or a union type can break some developers.

PR #1011, opened 2020-08-24
auth: Add credential service

MathBunny:
This pull request covers the following:

1. Splits the `index.d.ts` file which allows to substitute the typings (i.e., `index.d.ts` import for `credentials` will map to `src/credential.d.ts` or `src/credential/index.d.ts` depending on if auto-generated types are toggled).
   - Note: This might introduce the duplicate "callable" bug in the documentation web-page
2. Refactors `credential.ts` to be a separate service outside of auth
3. Extracted the credentials related component of `auth/*.spec.ts` into new unit test folder `credential/*.spec.ts`
4. Adds documentation for auto-generated typings, as part of #992

### DIFF
--- a/src/credential/credential.ts
+++ b/src/credential/credential.ts
@@ -20,8 +20,10 @@ import os = require('os');
 import path = require('path');
 
 import { AppErrorCodes, FirebaseAppError } from '../utils/error';
+import { FirebaseApp } from '../firebase-app';
 import { HttpClient, HttpRequestConfig, HttpError, HttpResponse } from '../utils/api-request';
 import { Agent } from 'http';
+import { FirebaseServiceInterface, FirebaseServiceInternalsInterface } from '../firebase-service';
 import * as util from '../utils/validator';
 
 const GOOGLE_TOKEN_AUDIENCE = 'https://accounts.google.com/o/oauth2/token';
@@ -53,6 +55,11 @@ const REFRESH_TOKEN_PATH = '/oauth2/v4/token';
 const ONE_HOUR_IN_SECONDS = 60 * 60;
 const JWT_ALGORITHM = 'RS256';
 
+let globalAppDefaultCred: Credential;
+const globalCertCreds: { [key: string]: ServiceAccountCredential } = {};
+const globalRefreshTokenCreds: { [key: string]: RefreshTokenCredential } = {};
+
+
 /**
  * Interface for Google OAuth 2.0 access tokens.
  */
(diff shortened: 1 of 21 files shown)


src/credential/index.ts
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { FirebaseApp } from '../firebase-app';
+import * as credentialApi from './credential';
+import * as firebaseAdmin from '../index';
+
+export function credential(app?: FirebaseApp): credentialApi.CredentialService {
```
hiranya911, line 21: This is not needed.

src/firebase-app.ts
```diff
@@ -300,6 +302,18 @@ export class FirebaseApp {
     });
   }
 
+  /**
```
hiranya911, line 305: Not needed.

src/credential/index.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+/* eslint-disable @typescript-eslint/no-namespace */
+export namespace admin.credential {
+  // See https://github.com/microsoft/TypeScript/issues/4336
+  /* eslint-disable @typescript-eslint/no-unused-vars */
+  // See https://github.com/typescript-eslint/typescript-eslint/issues/363
+  // Allows for exposing classes as interfaces in typings
+  /* eslint-disable @typescript-eslint/no-empty-interface */
+  export import Credential = credentialApi.Credential;
+  export const applicationDefault = credentialApi.CredentialService.applicationDefault;
```
hiranya911, line 43: These should be top-level functions and should not be associated with an object instance. Developers will call these functions without initializing an App.

src/credential/credential.ts
```diff
@@ -64,12 +71,202 @@ export interface GoogleOAuthAccessToken {
...
+   * @return {Promise<()>} An empty Promise that will be fulfilled when the service is deleted.
+   */
+  public delete(): Promise<void> {
+    // There are no resources to clean up
+    return Promise.resolve(undefined);
+  }
+}
+
+
+export class CredentialService implements FirebaseServiceInterface {
```
hiranya911, line 112: Should not be needed. Expose the functions as top-level functions.

src/firebase-namespace.ts
```diff
@@ -335,6 +294,14 @@ export class FirebaseNamespace {
     return Object.assign(fn, { Auth: auth });
   }
 
+  get credential(): FirebaseServiceNamespace<CredentialService> {
```
hiranya911, line 297: Let's remove this and instead map `credential` to an object/namespace that has top-level functions.

src/credential/credential.ts
```diff
@@ -64,12 +71,202 @@ export interface GoogleOAuthAccessToken {
 }
 
 /**
- * Interface for things that generate access tokens.
+ * Interface that provides Google OAuth2 access tokens used to authenticate
```
hiranya911, line 74: Pretty much every type and function in this module that doesn't appear in the existing index.d.ts file is an internal API. We need to move them to an `-internal` file at some point too.

MathBunny: Yes, I'll do it in a subsequent PR

MathBunny:
Thanks for the feedback. I re-attempted this using a simpler approach instead of introducing an entirely new service for credentials. Right now I'm just mapping the functions from the original property `credential` in `FirebaseNamespace` to another file in `credential`.

PR #1012, opened 2020-08-25
Allow Credential to auto-generate typings, separate internal vs external APIs

MathBunny:
Allows `Credential` to auto-generate typings and separate internal vs external APIs. 

Follow-up to #1011.


src/credential/credential-internal.ts
```diff
@@ -0,0 +1,472 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Use untyped import syntax for Node built-ins
+import fs = require('fs');
+import os = require('os');
+import path = require('path');
+
+import { GoogleOAuthAccessToken, Credential } from './credential';
```
hiranya911, line 22: These 2 types should probably in a separate file to avoid the cyclic dependency between `credential` and `credential-internal`. Or move the credential factory functions to a separate file.

src/credential/credential-internal.ts
```diff
@@ -19,7 +19,7 @@ import fs = require('fs');
 import os = require('os');
 import path = require('path');
 
-import { GoogleOAuthAccessToken, Credential } from './credential';
+import { GoogleOAuthAccessToken, Credential } from './google-oauth-access-token';
```
hiranya911, line 22: Lets call it `credential-interfaces.ts` since it has more than just the access token type.

src/credential/index.ts
```diff
@@ -15,6 +15,7 @@
  */
 
 import * as credentialApi from './credential';
+import * as googleOauthAccessTokenApi from './google-oauth-access-token';
```
hiranya911, line 18: `GoogleOAuthAccessToken` interface should also be exported.  But looks like it's part of the `admin` namespace today. So perhaps we should do it later once we start refactoring the top-level `admin` namespace.

PR #1014, opened 2020-08-29
chore: Enabling max-len lint rule

hiranya911:
I'm enabling the `max-len` lint rule with a max line length of 120. Ideally we should be using a max line length of 100, but we have way too many violations to attempt that right now.


test/integration/remote-config.spec.ts
```diff
@@ -246,7 +246,7 @@ describe('admin.remoteConfig', () => {
     INVALID_JSON_STRINGS.forEach((invalidJson) => {
       it(`should throw if the json string is ${JSON.stringify(invalidJson)}`, () => {
         expect(() => admin.remoteConfig().createTemplateFromJSON(invalidJson))
-          .to.throw(/^Failed to parse the JSON string: ([\D\w]*)\. SyntaxError: Unexpected token ([\D\w]*) in JSON at position ([0-9]*)$/);
+          .to.throw(/Failed to parse the JSON string/);
```
lahirumaramba, line 249: Did this pass integration test? I thought the regex should match the complete message.

test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -529,11 +533,19 @@ describe('RemoteConfigApiClient', () => {
...
-    ['', 'abc', 'a123b', 'a123', '123a', 1.2, '70.2', null, NaN, true, [], {}].forEach(
+    ['', null, NaN, true, [], {}].forEach(
+      (invalidVersion) => {
+        it(`should throw if the endVersionNumber is: ${invalidVersion}`, () => {
+          expect(() => apiClient.listVersions({ endVersionNumber: invalidVersion } as any))
+            .to.throw(/^endVersionNumber must be a non-empty string in int64 format or a number$/);
+        });
+      });
+
+    ['abc', 'a123b', 'a123', '123a', 1.2, '70.2'].forEach(
```
lahirumaramba, line 544: Thanks! :)

test/unit/remote-config/remote-config.spec.ts
```diff
@@ -254,7 +257,21 @@ describe('RemoteConfig', () => {
...
+          .stub(RemoteConfigApiClient.prototype, 'listVersions')
+          .resolves(response);
+        stubs.push(stub);
+        return remoteConfig.listVersions()
+          .should.eventually.be.rejected
+          .and.to.match(/^Error: Version number must be a non-empty string in int64 format or a number$/);
+      });
+    });
+
+    ['abc', 'a123b', 'a123', '123a', 1.2, '70.2'].forEach((invalidVersion) => {
```
lahirumaramba, line 274: Thank you! :)

test/integration/remote-config.spec.ts
hiranya911: That was because the regex was defined with `^...$` symbols. Without them you don't need to match the whole message.

PR #1019, opened 2020-09-03
fix(storage): Support typing generation for the storage API

hiranya911:
Continuing the work started by @MathBunny, this PR adds support for auto-generating typings for the `Storage` APIs:

* Moved existing `admin.storage` typings into a new `src/storage.d.ts` file, so they can be easily screened out during typing auto-generation.
* Added `src/storage/index.ts` file with the barrel exports.
* Updated the `gulpfile` to auto-generate typings for `Storage`.


lahirumaramba (approved):
Thanks!!
LGTM!

PR #1021, opened 2020-09-03
fix(auth): Adding auth/index.ts barrel export

hiranya911:
* Adding the `auth/index.ts` file with the namespace definition.
* Moved `EmailSignInProviderConfig` to `auth-config` so it can be exposed.
* Moved `MultiFactorAuthServerConfig` to `auth-config-internal` since it doesn't need to be exposed.
* Made several changes to `BaseAuth` so it (and its child classes) can be cleanly exposed:
  - Changed `BaseAuth<AbstractAuthRequestHandler>` to just `BaseAuth` since `AbstractAuthRequestHandler` is an internal API
  - Added `@internal` to all protected members and the constructor of `BaseAuth`.
* Added `@internal` to the constructors of all publicly exposed classes: `Tenant`, `MultifactorInfo`, `UserInfo`, `UserRecord`

With these changes we are able to generate typings for all Auth APIs.


src/auth/index.ts
```diff
@@ -0,0 +1,88 @@
...
+import * as _identifier from './identifier';
+import * as _import from './user-import-builder';
+import * as _mfa from './multi-factor';
+import * as _tenant from './tenant';
+import * as _tenantManager from './tenant-manager';
+import * as _user from './user-record';
+
+/* eslint-disable @typescript-eslint/no-namespace */
+export namespace admin.auth {
+    /* eslint-disable @typescript-eslint/no-empty-interface */
```
bojeil-google, line 29: The indentation seems to be off in this block

src/auth/tenant.ts
```diff
@@ -89,26 +89,21 @@ export class Tenant {
...
-   * The multi-factor auth configuration on the current tenant.
-   */
-  public readonly multiFactorConfig?: MultiFactorAuthConfig;
-
   /**
    * The map containing the test phone number / code pairs for the tenant.
    */
   public readonly testPhoneNumbers?: {[phoneNumber: string]: string};
 
+  private readonly _emailSignInConfig?: EmailSignInConfig;
```
bojeil-google, line 97: What's the issue with `public readonly`? Are these not compatible with the type definition generator?

bojeil-google (changes requested):

src/auth/tenant.ts
hiranya911: Issue is not with `public readonly`. It's the types of these fields. Both `EmailSignInConfig` and `MultiFactorAuthConfig` are internal types, and therefore cannot be exposed in the public API. Their corresponding public interfaces are `EmailSignInProviderConfig` and `MultiFactorConfig`. I've declared the getters with the public types.

src/auth/index.ts
hiranya911: Good catch. Fixed.

src/auth/tenant.ts
bojeil-google: Ah ok, makes sense. Thanks for explaining.

bojeil-google (approved):

lahirumaramba (approved):

hiranya911:
This approach was superseded by #1072

PR #1022, opened 2020-09-09
chore: Experimental refactor for type auto generation

hiranya911:
**For feedback only**

I'm trying out a few improvements over the process developed by @MathBunny. The current approach can be summarized as:

* Move all internally-public types into `*-internal.ts` files.
* Barrel export all public types from `index.ts` files.

There are a few drawbacks to this approach:

1. A lot of work is required to carefully separate the public APIs from internally-public ones. We have already done most of the initial work towards this goal, but it is bound to come up again as we maintain the codebase in the future.
2. The barrel export syntax for namespaces (`export import Foo = api.Foo`) is rather awkward, repetitive and has little to no support from the tools like TypeDoc and API Extractor.
3. In places where we want to export a class as an interface, even stranger hacks are needed (like extending a class into an interface).

To address these issues I'm proposing the following "inverted" approach:

* Define the public APIs in exactly one place per service (`index.ts` file). This is easily done by copying the content from the existing d.ts files. For example, we can copy the `remote-config.d.ts` file contents into `remote-config/index.ts` to define the public API of the service.
* Then we use the public API types everywhere in the implementation, as opposed to barrel exporting the implementation as the API. For example in `remote-config/remote-config.ts` file we can do:

```ts
// Import the public API defined in index.ts
import { remoteConfig } from './index';

import RemoteConfigInterface = remoteConfig.RemoteConfig;
import RemoteConfigTemplate = remoteConfig.RemoteConfigTemplate;

// Implement the public API type in the impl class
export class RemoteConfig implements RemoteConfigInterface {
  public getTemplate(): Promise<RemoteConfigTemplate> {

  }
}
```

This seems to address all the issues listed above. Type generation works as expected, and the output  should be compatible with tools like TypeDoc as it is a copy of the d.ts file we already use. In this PR I've prototyped this approach for the remote config module. Feedback welcome.


MathBunny:
Chiming in, this makes sense to me! The only thing I might be worried about is the `index.ts` files potentially getting large for services like auth, in the previous design it was separated across multiple files. Indeed, from what I can tell this would work a lot better with TypeDoc but it might still suffer from the `admin.` prefix missing. 

Also, does the sanity test pass with this while types generation mode is auto? I wonder how it deals with the missing `admin.` in the namespace re-export, as the `index.d.ts` wasn't changed.

hiranya911:
> The only thing I might be worried about is the index.ts files potentially getting large for services like auth

We already have large manually-curated `d.ts` files for these services. We are just absorbing them into the implementation, and get the compiler to verify that the typings actually match the implementation.

> it might still suffer from the admin. prefix missing.

We cannot have `admin.` prefix in the service-level `index.ts` files (if you can remember from that small poc example I put together a while ago). We need the namespace name to match the function name, so declaration merging works correctly. The `admin.` prefix gets automatically added at the `firebase-namespace.ts` level, which I have tried to emulate with the `new-firebase-namespace.ts` file. 

 > does the sanity test pass with this while types generation mode is auto?

It probably won't. Not until we fully refactor the `firebase-namespace.ts`.

hiranya911:
@MathBunny to further demonstrate my points, I've also refactored Auth APIs and pushed the changes to this branch.

hiranya911:
Full implementation at #1072

PR #1032, opened 2020-09-16
chore: Removing ambient typing files

hiranya911:
I'm setting up a new `typegen` branch to complete the typing auto-generation work. 

* Removed all manually-curated d.ts files.
* Enabled typing generation in the `tsconfig`. Although the generated d.ts files are not copied to the build output yet.
* Disabled `build:tests` task in CI. Integration tests won't compile until we start copying the typings to the build output.
* Added empty `firebase-namespace.d.ts` and `firebase-namespace-api.ts` files. These files will be populated in the future PRs.
* Removed the obsolete `copyDatabase` task from the gulpfile. This task was essentially a no-op.
* Removed `@typescript-eslint/no-unused-vars-experimental` lint rule. As per https://github.com/typescript-eslint/typescript-eslint/issues/1856#issuecomment-670191639 this rule is now deprecated, and the regular `no-unsed-var` rule just works!

go/firebase-node-one-source


lahirumaramba (approved):
Thanks for this!
LGTM!

PR #1035, opened 2020-09-21
chore: Added the credentials API signatures and exposed admin.credential

hiranya911:
* Moved credential API signatures to `credential/index.ts`.
* Implemented the API signatures in `credential/credential.ts` and `credential/credential-internal.ts`.
* Added the generated `credential.d.ts` file to the build output.
* Exported credential types and the namespace from `firebase-namespace.d.ts`.


test/resources/mocks.ts
```diff
@@ -27,7 +27,7 @@ import * as jwt from 'jsonwebtoken';
 import { FirebaseNamespace } from '../../src/firebase-namespace';
 import { FirebaseServiceInterface } from '../../src/firebase-service';
 import { FirebaseApp, FirebaseAppOptions } from '../../src/firebase-app';
-import { Credential, GoogleOAuthAccessToken } from '../../src/credential/credential-interfaces';
+import { credential as _credential, GoogleOAuthAccessToken } from '../../src/credential/index';
```
lahirumaramba, line 30: Just curious, why we use the `_credential` alias here?

lahirumaramba (approved):
LGTM!

test/resources/mocks.ts
hiranya911: There's already a `credential` global variable in this file.

PR #1038, opened 2020-09-23
fix(rc): Exposed admin.remoteConfig namespace

hiranya911:
* Added RC API definitions to `remoteConfig/index.ts`.
* Exposed `admin.remoteConfig` namespace.
* Exposed `App.remoteConfig()` method.


lahirumaramba (approved):
Thanks! Looks good to me!

PR #1040, opened 2020-09-23
Update default.hbs

egilmorez:
Adding meta tag that will be used later by recommendations feature of devsite.

This shouldn't have any other effect at all.

Thanks!


hiranya911 (approved):
LGTM

PR #1043, opened 2020-09-25
chore(rtdb): Exposed admin.database namespace

hiranya911:
* Exposed `admin.database` namespace.
* Re-exported types from the `@firebase/database-types` package from the new namespace.
* Used module augmentation to add rules management APIs to the `Database` type.


hiranya911:
@Feiyang1 I assigned you as a reviewer too. Plan is to auto-generate typings from the `src/database/index.ts` file for RTDB (as opposed to having RTDB typings duplicated in the Admin SDK repo). Appreciate your input.

lahirumaramba (approved):
LGTM!

src/database/index.ts
```diff
@@ -14,33 +14,83 @@
...
+ * var otherDatabase = admin.database(app);
+ * ```
+ *
+ * @param App whose `Database` service to
+ *   return. If not provided, the default `Database` service will be returned.
+ *
+ * @return The default `Database` service if no app
+ *   is provided or the `Database` service associated with the provided app.
  */
+export declare function database(app?: app.App): database.Database;
```
Feiyang1, line 52: What is it for?

src/database/index.ts
```diff
@@ -14,33 +14,83 @@
...
-  // See https://github.com/typescript-eslint/typescript-eslint/issues/363
-  export import DataSnapshot = firebaseDbTypesApi.DataSnapshot;
-  export import Database = adminDb.Database;
-  export import EventType = firebaseDbTypesApi.EventType;
-  export import OnDisconnect = firebaseDbTypesApi.OnDisconnect;
-  export import Query = firebaseDbTypesApi.Query;
-  export import Reference = firebaseDbTypesApi.Reference;
-  export import ThenableReference = firebaseDbTypesApi.ThenableReference;
-  export import enableLogging = firebaseDbTypesApi.enableLogging;
+  export import Database = rtdb.FirebaseDatabase;
```
Feiyang1, line 86: Can you directly export these types instead of nesting them in the namespace, so you don't have to do `import Database = database.Database;` everywhere?

hiranya911: Our current API puts all RTDB types in an `admin.database` namespace. So this nesting is required.

src/database/index.ts
hiranya911: This is the signature of `admin.database()` API

Feiyang1 (approved):
LGTM

PR #1044, opened 2020-09-28
Add support for Auth Emulator

samtstern:
### Discussion

For Googlers: [this design doc](https://docs.google.com/document/d/1EDQh305tKokFU8A_A9G4SPQBhappt-TkUMLRI9tyBFE/edit?usp=sharing) explains the behavior changes and security implications.

- [x] Basic URL replacement
- [x] Short-circuit custom auth token signing
- [x]  Short-circuit `verifyIdToken`

### Testing

- [x] New unit tests

### API Changes

N/A

### DIFF
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -31,7 +31,9 @@ import {
 
 import * as utils from '../utils/index';
 import * as validator from '../utils/validator';
-import { FirebaseTokenVerifier, createSessionCookieVerifier, createIdTokenVerifier, ALGORITHM_RS256 } from './token-verifier';
+import { 
+  FirebaseTokenVerifier, createSessionCookieVerifier, createIdTokenVerifier, ALGORITHM_RS256 
+} from './token-verifier';
 import { ActionCodeSettings } from './action-code-settings-builder';
 import {
   AuthProviderConfig, AuthProviderConfigFilter, ListProviderConfigResults, UpdateAuthProviderRequest,
@@ -752,8 +754,8 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> {
    * Enable or disable ID token verification. This is used to safely short-circuit token verification with the
    * Auth emulator. When disabled ONLY unsigned tokens will pass verification, production tokens will not pass.
    */
-  // @ts-ignore: This private method will be used in the Functions emulator
-  private setIdTokenVerificationEnabled(enabled: boolean) {
+  // @ts-expect-error: This private method will be used in the Functions emulator
+  private setIdTokenVerificationEnabled(enabled: boolean): void {
     const algorithm = enabled ? ALGORITHM_RS256 : "none";
     this.idTokenVerifier.setAlgorithm(algorithm)
     this.sessionCookieVerifier.setAlgorithm(algorithm);
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -142,7 +142,8 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> {
       this.tokenGenerator = tokenGenerator;
     } else {
       const cryptoSigner = cryptoSignerFromApp(app);
-      this.tokenGenerator = new FirebaseTokenGenerator(cryptoSigner);
+      const useEmulator = !!process.env.FIREBASE_AUTH_EMULATOR_HOST;
+      this.tokenGenerator = new FirebaseTokenGenerator(cryptoSigner, undefined, useEmulator);
     }
 
     this.sessionCookieVerifier = createSessionCookieVerifier(app);
@@ -753,7 +754,8 @@ export class TenantAwareAuth extends BaseAuth<TenantAwareAuthRequestHandler> {
    */
   constructor(app: FirebaseApp, tenantId: string) {
     const cryptoSigner = cryptoSignerFromApp(app);
-    const tokenGenerator = new FirebaseTokenGenerator(cryptoSigner, tenantId);
+    const useEmulator = !!process.env.FIREBASE_AUTH_EMULATOR_HOST;
+    const tokenGenerator = new FirebaseTokenGenerator(cryptoSigner, tenantId, useEmulator);
     super(app, new TenantAwareAuthRequestHandler(app, tenantId), tokenGenerator);
     utils.addReadonlyGetter(this, 'tenantId', tenantId);
   }
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -130,18 +130,6 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> {
   protected readonly sessionCookieVerifier: FirebaseTokenVerifier;
 
   /**
-   * When true the SDK should communicate with the Auth Emulator for all API
-   * calls and also produce unsigned tokens.
-   * 
-   * This alone does <b>NOT<b> short-circuit ID Token verification. 
-   * For security reasons that must be explicitly disabled through
-   * setIdTokenVerificationEnabled(false);
-   */
-  protected static useEmulator(): boolean {
-    return !!process.env.FIREBASE_AUTH_EMULATOR_HOST;
-  }
-
-  /**
    * The BaseAuth class constructor.
    *
    * @param app The FirebaseApp to associate with this Auth instance.
@@ -155,7 +143,7 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> {
     if (tokenGenerator) {
       this.tokenGenerator = tokenGenerator;
     } else {
-      const cryptoSigner = BaseAuth.useEmulator() ? new EmulatedSigner() : cryptoSignerFromApp(app);
+      const cryptoSigner = useEmulator() ? new EmulatedSigner() : cryptoSignerFromApp(app);
       this.tokenGenerator = new FirebaseTokenGenerator(cryptoSigner);
     }
 
@@ -753,10 +741,11 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> {
   /**
    * Enable or disable ID token verification. This is used to safely short-circuit token verification with the
    * Auth emulator. When disabled ONLY unsigned tokens will pass verification, production tokens will not pass.
+   * 
+   * @internal
    */
-  // @ts-expect-error: This private method will be used in the Functions emulator
-  private setIdTokenVerificationEnabled(enabled: boolean): void {
-    const algorithm = enabled ? ALGORITHM_RS256 : "none";
+  public setJwtVerificationEnabled(enabled: boolean): void {
+    const algorithm = enabled ? ALGORITHM_RS256 : 'none';
     this.idTokenVerifier.setAlgorithm(algorithm)
     this.sessionCookieVerifier.setAlgorithm(algorithm);
   }
@@ -777,7 +766,7 @@ export class TenantAwareAuth extends BaseAuth<TenantAwareAuthRequestHandler> {
    * @constructor
    */
   constructor(app: FirebaseApp, tenantId: string) {
-    const cryptoSigner = BaseAuth.useEmulator() ? new EmulatedSigner() : cryptoSignerFromApp(app);
+    const cryptoSigner = useEmulator() ? new EmulatedSigner() : cryptoSignerFromApp(app);
     const tokenGenerator = new FirebaseTokenGenerator(cryptoSigner, tenantId);
     super(app, new TenantAwareAuthRequestHandler(app, tenantId), tokenGenerator);
     utils.addReadonlyGetter(this, 'tenantId', tenantId);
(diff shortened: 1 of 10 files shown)


src/auth/token-verifier.ts
```diff
@@ -237,13 +247,15 @@ export class FirebaseTokenVerifier {
...
    */
-  private verifyJwtSignatureWithKey(jwtToken: string, publicKey: string): Promise<DecodedIdToken> {
+  private verifyJwtSignatureWithKey(jwtToken: string, publicKey: string | null): Promise<DecodedIdToken> {
     const verifyJwtTokenDocsMessage = ` See ${this.tokenInfo.url} ` +
       `for details on how to retrieve ${this.shortNameArticle} ${this.tokenInfo.shortName}.`;
     return new Promise((resolve, reject) => {
-      jwt.verify(jwtToken, publicKey, {
+      // The types are wrong here so it won't accept 'null' for publicKey, but testing (including our unit tests)
+      // confirms that this works and is the required value when validating a token with alg: 'none'
+      jwt.verify(jwtToken, publicKey as any, {
```
yuchenshi, line 256: Drive by comment: consider using `jwt.decode` for emulator only, which has the right typings and explicitly skips signature verification. We do this in Auth Emulator.

test/resources/mocks.ts
```diff
@@ -30,7 +30,7 @@ import { FirebaseApp, FirebaseAppOptions } from '../../src/firebase-app';
 import { Credential, GoogleOAuthAccessToken } from '../../src/credential/credential-interfaces';
 import { ServiceAccountCredential } from '../../src/credential/credential-internal';
 
-const ALGORITHM = 'RS256';
+const ALGORITHM = 'RS256' as 'RS256';
```
yuchenshi, line 33: `as const`

yuchenshi:
BTW awesome work, will take a detailed pass later once we settle the bigger picture questions

src/auth/token-verifier.ts
samtstern: Yeah I considered that my reasoning for this was that we use `verify` in the actual code and I want the tests to prove that won't throw.

samtstern: I submitted a PR to fix the types:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/48348

If it gets merged I can remove the hack.

yuchenshi: nit: Have you tried `""` (empty string) instead of `null`?

test/resources/mocks.ts
yuchenshi:
```suggestion
const ALGORITHM = 'RS256' as const;
```

test/unit/auth/token-verifier.spec.ts
```diff
@@ -104,6 +104,17 @@ function mockFailedFetchPublicKeys(): nock.Scope {
     .replyWithError('message');
 }
 
+function createTokenVerifier(app: FirebaseApp, useEmulator: boolean): verifier.FirebaseTokenVerifier {
```
yuchenshi, line 107: nit: Please convert `useEmulator` to an options hash (e.g. `{useEmulator: true}`). The boolean literal arguments below are pretty hard to read.

yuchenshi (approved):
LGTM (pending `verifyIdToken` change)

src/auth/token-generator.ts
```diff
@@ -297,8 +299,13 @@ export class FirebaseTokenGenerator {
       }
     }
     return this.signer.getAccountId().then((account) => {
+      // When communicating with the Auth Emulator we don't sign custom tokens which means:
```
hiranya911, line 302: What about the call to `getAccountId()` which may also make rpc calls?

src/auth/token-verifier.ts
```diff
@@ -144,6 +150,10 @@ export class FirebaseTokenVerifier {
       );
     }
 
+    if (this.useEmulator) {
```
hiranya911, line 153: This bypasses all the token verification checks. Wouldn't you want to check for timestamps and other standard claims like issuer? A developer might wish to write tests that check for such scenarios.

src/auth/token-verifier.ts
```diff
@@ -75,9 +75,15 @@ export class FirebaseTokenVerifier {
   private publicKeysExpireAt: number;
   private readonly shortNameArticle: string;
 
-  constructor(private clientCertUrl: string, private algorithm: string,
+  constructor(private clientCertUrl: string, private algorithm: jwt.Algorithm,
```
hiranya911, line 78: This really ought to accept an options object. But it's ok to do that later.

src/auth/token-generator.ts
```diff
@@ -242,8 +242,10 @@ export class FirebaseTokenGenerator {
    * @param tenantId The tenant ID to use for the generated Firebase Auth
    *     Custom token. If absent, then no tenant ID claim will be set in the
    *     resulting JWT.
+   * @param useEmulator When true we should generate tokens on behalf of the
+   *     Auth Emulator, which means no signature.
    */
-  constructor(signer: CryptoSigner, public readonly tenantId?: string) {
+  constructor(signer: CryptoSigner, public readonly tenantId?: string, public readonly useEmulator?: boolean) {
```
hiranya911, line 248: Can we handle all the emulator-related logic in a separate implementation of `CryptoSigner`?

```
interface CryptoSigner {
  algorithm: jwt.Algorithm;
  getAccountId(): Promise<string>;
  sign(buffer: Buffer): Promise<Buffer>;
}

class EmulatedSigner implements CryptoSigner {
  public readonly algorithm = 'none';

  // and so on
}
```

test/unit/auth/token-generator.spec.ts
```diff
@@ -308,6 +308,28 @@ describe('FirebaseTokenGenerator', () => {
     tokenGenerator: new FirebaseTokenGenerator(new ServiceAccountSigner(cert), tenantId),
   }];
 
+  describe('Emulator', () => {
```
hiranya911, line 311: We also need to test this via the Auth interface. Programmatically set the env variable and trigger the emulator mode.

test/unit/auth/token-verifier.spec.ts
```diff
@@ -535,6 +540,45 @@ describe('FirebaseTokenVerifier', () => {
         });
     });
 
+    it('should decode an unsigned token when using the emulator', async () => {
```
hiranya911, line 543: Same here. Need some tests via the Auth interface.

Also need to test for other rpc calling APIs like `getUser()` and verify that they behave as expected in the emulator mode.

hiranya911 (changes requested):
Looks mostly good. I've made some suggestions on cleaning up the implementation and improving the test coverage.

src/auth/token-verifier.ts
samtstern: Wow I think that worked!

src/auth/auth.ts
```diff
@@ -138,15 +138,17 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> {
...
    */
   constructor(app: FirebaseApp, protected readonly authRequestHandler: T, tokenGenerator?: FirebaseTokenGenerator) {
+    const useEmulator = !!process.env.FIREBASE_AUTH_EMULATOR_HOST;
+
     if (tokenGenerator) {
       this.tokenGenerator = tokenGenerator;
     } else {
       const cryptoSigner = cryptoSignerFromApp(app);
-      this.tokenGenerator = new FirebaseTokenGenerator(cryptoSigner);
+      this.tokenGenerator = new FirebaseTokenGenerator(cryptoSigner, undefined, useEmulator);
```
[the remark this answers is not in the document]
samtstern: Ok happy to do that, so I assume `FirebaseTokenGenerator` is not a public API and therefore we're not worried about a breaking change?

src/auth/token-generator.ts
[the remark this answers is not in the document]
samtstern: Good idea.  I was worried about "breaking" this API but now see that it's internal only.

src/auth/auth.ts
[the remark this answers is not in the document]
samtstern: Actually you were right, delegating this all to the `CryptoSigner` is much cleaner.

src/auth/token-generator.ts
samtstern: Great idea, thanks.

src/auth/token-verifier.ts
samtstern: Good idea, re-instated all checks.

src/auth/token-generator.ts
samtstern: Yeah moved this into an implementation.

src/auth/token-generator.ts
```diff
@@ -213,6 +226,36 @@ export class IAMSigner implements CryptoSigner {
...
+   */
+  public sign(_: Buffer): Promise<Buffer> {
+    return Promise.resolve(Buffer.from(''));
+  }
+
+  /**
+   * @inheritDoc
+   */
+  public getAccountId(): Promise<string> {
+    return Promise.resolve('firebase-auth-emulator@example.com');
```
samtstern, line 257: I'm not so sure about this ... could we do something better here?

samtstern: cc @yuchenshi

samtstern:
@yuchenshi and @hiranya911 thank you both for your comments, I got the code into a **much** cleaner state where `useEmulator` is no longer passed down beyond `Auth`.

src/auth/token-generator.ts
yuchenshi: This is probably the best thing we can do

yuchenshi (approved):

src/auth/auth.ts
```diff
@@ -127,6 +129,18 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> {
...
 
+  /**
+   * When true the SDK should communicate with the Auth Emulator for all API
+   * calls and also produce unsigned tokens.
+   * 
+   * This alone does <b>NOT<b> short-circuit ID Token verification. 
+   * For security reasons that must be explicitly disabled through
+   * setIdTokenVerificationEnabled(false);
+   */
+  protected static useEmulator(): boolean {
```
hiranya911, line 140: I suggest we implement this as a module-level function (unexported). No reason why it should be in this class.  Since the `Auth` object is exposed to the developer, let's only add the bare minimum to it.

hiranya911 (changes requested):
Thanks. This looks pretty good. Just a bunch of nits and cleanup suggestions. Lets use single quotes around string literals everywhere.

src/auth/token-generator.ts
```diff
@@ -21,9 +21,12 @@ import { AuthorizedHttpClient, HttpError, HttpRequestConfig, HttpClient } from '
 
 import * as validator from '../utils/validator';
 import { toWebSafeBase64 } from '../utils';
+import { Algorithm } from "jsonwebtoken";
```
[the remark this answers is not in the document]
samtstern: Thanks, can lint be set up to find these?  I do it by accident and they're hard to search for.

src/auth/auth.ts
```diff
@@ -737,6 +749,17 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> {
         return decodedIdToken;
       });
   }
+
+  /**
+   * Enable or disable ID token verification. This is used to safely short-circuit token verification with the
+   * Auth emulator. When disabled ONLY unsigned tokens will pass verification, production tokens will not pass.
+   */
+  // @ts-expect-error: This private method will be used in the Functions emulator
+  private setIdTokenVerificationEnabled(enabled: boolean): void {
```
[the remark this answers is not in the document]
samtstern: Didn't know about @internal thanks!

src/auth/token-verifier.ts
```diff
@@ -339,14 +342,16 @@ export class FirebaseTokenVerifier {
  * @param {FirebaseApp} app Firebase app instance.
  * @return {FirebaseTokenVerifier}
  */
-export function createIdTokenVerifier(app: FirebaseApp, useEmulator = false): FirebaseTokenVerifier {
+export function createIdTokenVerifier(
```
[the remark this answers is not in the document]
samtstern: It made me do it to make the line-length lint happy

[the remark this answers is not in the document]
samtstern: Oh nevermind I see, good catch!

hiranya911 (approved):
Thanks. LGTM 👍

src/auth/auth.ts
```diff
@@ -735,6 +737,28 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> {
...
+   * 
+   * WARNING: This is a dangerous method that will compromise your app's security and break your app in
+   * production. Developers should never call this method, it is for internal testing use only.
+   * 
+   * @internal
+   */
+  // @ts-expect-error: this method appears unused but is used privately.
+  private setJwtVerificationEnabled(enabled: boolean): void {
+    if (!enabled && !useEmulator()) {
+      // We only allow verification to be disabled in conjunction with
```
hiranya911, line 753: Consider throwing instead of silently failing.

hiranya911 (approved):
LGTM with a suggestion.

src/auth/auth.ts
samtstern: I figured silently failing adds a little security-by-obscurity since we don't actually want to help people enable this.  So if they randomly call it in an inappropriate environment, we just ignore them rather than throwing an error that may help them call it.

I realize that's not real security, so I can go either way.  @yuchenshi ?

yuchenshi: I'd agree with @hiranya911 here and I think it should just throw, so doing it in production will consistently fail. With that being said, the error message doesn't need to be helpful since it's an uncommon case. Something like `this method is only available in the emulators` can point them at the right direction.

yuchenshi (approved):

joelpoloney:
❤️❤️❤️❤️❤️❤️

nickreynke:
Thank you for implementing the emulator! Good work!

PR #1046, opened 2020-09-28
chore(iid): Exposed admin.instanceId namespace

hiranya911:
* Exposed `admin.instanceId` namespace.
* Exposed `App.instanceId()` method.

### DIFF
--- a/src/instance-id/index.ts
+++ b/src/instance-id/index.ts
@@ -37,15 +37,15 @@ import { app } from '../firebase-namespace-api';
  * // Get the Instance ID service for a given app
  * var otherInstanceId = admin.instanceId(otherApp);
  *```
-  *
-  * @param app Optional app whose `InstanceId` service to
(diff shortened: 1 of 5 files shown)


src/instance-id/index.ts
```diff
@@ -14,24 +14,72 @@
...
+ * // Get the Instance ID service for the default app
+ * var defaultInstanceId = admin.instanceId();
+ * ```
+ *
+ * @example
+ * ```javascript
+ * // Get the Instance ID service for a given app
+ * var otherInstanceId = admin.instanceId(otherApp);
+ *```
+  *
```
lahirumaramba, line 40: Nit: indentation off

lahirumaramba (approved):
LGTM!
Left one comment

src/instance-id/index.ts
hiranya911: Thanks. Fixed.

PR #1054, opened 2020-10-01
chore: Exposed admin.projectManagement namespace

hiranya911:
* Exposed `admin.projectManagement` namespace and the `App.projectManagement()` method.


lahirumaramba (approved):
LGTM!

PR #1057, opened 2020-10-06
chore: Using public API types in FirebaseNamespace impl

hiranya911:
* Using the public `App` type in `AppHook`, `FirebaseNamespace` and `FirebaseServiceInterface` types.
* Configured the gulpfile to copy all `**/index.d.ts` files to the output except `utils/index.d.ts` which is internal.

### DIFF
--- a/src/firebase-namespace.ts
+++ b/src/firebase-namespace.ts
@@ -15,31 +15,40 @@
  */
 
 import fs = require('fs');
+
 import { deepExtend } from './utils/deep-copy';
 import { AppErrorCodes, FirebaseAppError } from './utils/error';
-import { AppOptions } from './firebase-namespace-api';
+import { AppOptions, app } from './firebase-namespace-api';
 import { AppHook, FirebaseApp } from './firebase-app';
 import { FirebaseServiceFactory, FirebaseServiceInterface } from './firebase-service';
-import {
-  cert, refreshToken, applicationDefault
-} from './credential/credential';
+import { cert, refreshToken, applicationDefault } from './credential/credential';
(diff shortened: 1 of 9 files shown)


src/firebase-namespace.ts
```diff
@@ -169,14 +173,14 @@ export class FirebaseNamespaceInternals {
...
    *
-   * @param {string} serviceName The name of the Firebase service to register.
-   * @param {FirebaseServiceFactory} createService A factory method to generate an instance of the Firebase service.
-   * @param {object} [serviceProperties] Optional properties to extend this Firebase namespace with.
-   * @param {AppHook} [appHook] Optional callback that handles app-related events like app creation and deletion.
-   * @return {FirebaseServiceNamespace<FirebaseServiceInterface>} The Firebase service's namespace.
+   * @param serviceName The name of the Firebase service to register.
+   * @param createService A factory method to generate an instance of the Firebase service.
+   * @param serviceProperties Optional properties to extend this Firebase namespace with.
+   * @param appHook Optional callback that handles app-related events like app creation and deletion.
```
lahirumaramba, line 182: Do these need back-ticks? `createService`.
The original docs didn't have `` either so it might be okay :)

lahirumaramba (approved):
LGTM!
Left one comment

src/firebase-namespace.ts
hiranya911: The parameter names don't need backticks. We do need to do some additional formatting changes to meet some API Extractor requirements. I'll do them in a separate PR.

PR #1060, opened 2020-10-09
chore: Updated TypeDoc workflow to generate API docs

hiranya911:
Single API Extractor/API Documenter integration can take a while, I'm fixing our existing TypeDoc pipeline to work with the auto-generated typings.

* Updated the API docs generator to run against auto-generated d.ts files in the `lib/` directory.
* Updated the affected TOC entries.
* Adding RTDB type aliases to the generated HTML (in a manner similar to how Firestore type aliases are handled)
* Fixed a bunch of broken cross links.
* Defined `Database` as a type that extends the `@firebase/database` package, as opposed to using module augmentation. TypeDoc generates a much better HTML output for RTDB types with this change.

Preview at cl/336354281

### DIFF
--- a/docgen/content-sources/node/toc.yaml
+++ b/docgen/content-sources/node/toc.yaml
@@ -36,6 +36,8 @@ toc:
     path: /docs/reference/admin/node/admin.auth.CreatePhoneMultiFactorInfoRequest
   - title: "CreateRequest"
     path: /docs/reference/admin/node/admin.auth.CreateRequest
+  - title: "EmailSignInProviderConfig"
+    path: /docs/reference/admin/node/admin.auth.EmailSignInProviderConfig
   - title: "ListProviderConfigResults"
     path: /docs/reference/admin/node/admin.auth.ListProviderConfigResults
   - title: "ListTenantsResult"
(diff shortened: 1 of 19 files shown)


docgen/content-sources/node/toc.yaml
```diff
@@ -115,25 +117,13 @@ toc:
   path: /docs/reference/admin/node/admin.credential
   section:
   - title: "Credential"
-    path: /docs/reference/admin/node/admin.credential.Credential
+    path: /docs/reference/admin/node/admin.credential.Credential-1
```
lahirumaramba, line 120: Sorry if I missed something obvious! Why do we suffix these with `-1`?

hiranya911: That's the name of the auto-generated file now.

lahirumaramba (approved):
LGTM!

docgen/content-sources/node/toc.yaml
egilmorez: Is that something we want, or is it kind of forced on us?

hiranya911: Not by choice. It was the name auto picked by typedoc

hiranya911: I believe this is because we have merged declarations. `credential` is both a module and a namespace. TypeDoc is probably not handling that well.

egilmorez (approved):
Thanks Hiranya!

PR #1072, opened 2020-10-22
chore: Auto-generating typings from the source

hiranya911:
Merging the `typegen` branch into `master`. Changes include:

* Removed the manually-curated typings files.
* Added a single `firebase-namespace.d.ts` file that defines the typings for the top-level `admin` namespace. This was necessary to retain service lazy loading.
* Declared service namespaces and public APIs at a series of `{serviceName}/index.ts` files.
  - Other implementation files now import public API types from these index files.
* Updated the gulp file to bundle the generated typings files into the release artifact.
* Updated the `typedoc` workflow to generate documentation from auto-generated typings files.

Original proposal for historic context: go/firebase-node-auto-typing
Updated and revamped proposal as implemented here: go/firebase-node-one-source


lahirumaramba (approved):
Thanks! LGTM

PR #1079, opened 2020-11-02
fix: Adding es2018 libraries to the build

hiranya911:
Latest versions of GCP dependencies like Firestore require `es2018` libraries included in the tsc build. This helps avoid build errors like https://github.com/firebase/firebase-admin-node/actions/runs/342357542


lahirumaramba (approved):
Thanks! LGTM!

Issue #1078, opened 2020-11-01
Error checking if user exists - "TypeError: Cannot read property 'localId' of undefined"


aaronosher:
### [READ] Step 1: Are you in the right place?

  * For issues related to __the code in this repository__ file a Github issue.
  * If the issue pertains to Cloud Firestore, read the instructions in the "Firestore issue"
    template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Operating System version: macOS 10.15.7
  * Firebase SDK version: 9.3.0
  * Firebase Product: auth
  * Node.js version: 12.19.0
  * NPM version: 6.14.8

### [REQUIRED] Step 3: Describe the problem

With the functions and authentication emulators, when running `auth().getUser()` with a UID that doesn't exist, I get `TypeError: Cannot read property 'localId' of undefined`. And not an error with code `auth/user-not-found` as per the documentation.

I have confirmed this is specific to the emulator, as I don't get the TypeError when I pushed my function to a production environment.

#### Steps to reproduce:

1. Initialise a firebase project with the authentication and functions emulator
2. Write a function that checks if a user exists by using `auth().getUser()`
3. Execute your function with the emulator

Edit: I've realised it's probably worthwhile adding the stack trace
```
TypeError: Cannot read property 'localId' of undefined
    at new UserRecord (functions/node_modules/firebase-admin/lib/auth/user-record.js:272:23)
    at functions/node_modules/firebase-admin/lib/auth/auth.js:133:20
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Object.exports.userExists (functions/lib/services/users.js:8:22)
    at async functions/lib/routes/discord.js:36:24
```

#### Relevant Code:

Here is the code I've used to check if a user exists

```TypeScript
import { auth } from 'firebase-admin';

export const userExists = async (uid: string): Promise<boolean> => {
  try {
    const user = await auth().getUser(uid);
    if (user) {
      return true;
    }

    return false;
  } catch (e) {
    if (e.code === 'auth/user-not-found') {
      return false;
    }
    throw e;
  }
};
[...]

PR #1082, opened 2020-11-05
Handle lookup returning empty array of users.

yuchenshi:
### Discussion

This is a follow up to #1078 and improves the detection to be more robust for detecting no users returned.

Previously, it depends on `.users` not being present (instead of an empty array) when there are no users. This is not explicitly part of the [API documentation](https://cloud.google.com/identity-platform/docs/reference/rest/v1/projects.accounts/lookup) or [contract](https://cloud.google.com/identity-platform/docs/reference/rest/v1/GetAccountInfoResponse) and we probably shouldn't depend on it.

### Testing

* New tests added and existing tests updated for the responseValidator.

### API Changes

No public API changes included.


test/unit/auth/auth-api-request.spec.ts
```diff
@@ -369,17 +369,23 @@ describe('FIREBASE_AUTH_GET_ACCOUNT_INFO', () => {
...
     const responseValidator = FIREBASE_AUTH_GET_ACCOUNT_INFO.getResponseValidator();
     it('should succeed with users returned', () => {
-      const validResponse: object = { users: [] };
+      const validResponse: object = { users: [{ localId: 'foo' }] };
       expect(() => {
         return responseValidator(validResponse);
       }).not.to.throw();
     });
-    it('should fail when users is not returned', () => {
+    it('should fail when users is empty', () => {
```
bojeil-google, line 377: nit: this test description and the one below sounds similar. Let's rephrase to be clearer:
`should fail when the response object is empty`

test/unit/auth/auth-api-request.spec.ts
```diff
@@ -369,17 +369,23 @@ describe('FIREBASE_AUTH_GET_ACCOUNT_INFO', () => {
...
       const invalidResponse = {};
       expect(() => {
         responseValidator(invalidResponse);
       }).to.throw();
     });
+    it('should fail when users is not returned', () => {
```
bojeil-google, line 383: `should fail when the response object has an empty list of users`

bojeil-google (changes requested):
Just 2 nits

test/unit/auth/auth-api-request.spec.ts
yuchenshi: Done! Thanks for the suggestion!

test/unit/auth/auth-api-request.spec.ts
yuchenshi: Done! Thanks for the suggestion!

bojeil-google (approved):

hiranya911 (approved):
LGTM

Issue #1088, opened 2020-11-13
firebase remote config get template api fails throws "Version update time must be a valid date string"


arshedg:
### Issue is reproducible in Firebase cloud function and in the local cloud function simulator(started occurring from 13-nov-2020)

  * Firebase SDK version: 9.4.0
  * Firebase Product: remote config
  * Node.js version: 10.10.0

### Firebase node admin api "admin.remoteConfig().getTemplate() or config.getTemplateAtVersion" throws error
```
{"code":"remote-config/invalid-argument","message":"Version update time must be a valid date string"}
```
The issue is reproducible for all remote config versions updated today(using firebase ui or rest api).
Older remote config versions can be fetched.
```
config.getTemplateAtVersion(today's version) -> fails
config.getTemplateAtVersion(yesterday or previous version) -> works
```
#### Steps to reproduce:
update a remote config api using firebase ui.
get the template using the admin api
#### additional information
The version details of today's config(which doesn't work) look like
```
{
  "versionNumber": "7771",
  "updateTime": "2020-11-13T08:02:12.468168Z",
.....
}
```
Version details of old remote config (which work)
```
{
  "versionNumber": "7740",
  "updateTime": "2020-11-12T14:53:52.351Z",
...
```
You can see the time format is different for new and old remote config. I guess that's the reason why api is failing

sulayyudiz:
I am also facing the same issue

whereissami:
Experiencing same issue cant get remote config values in my cloud functions.

evilass:
We also facing this one since today.

fkneist:
Same here

WiNloSt:
I'm also facing the same problem, I've debugging like crazy the past few hours until I found this issue.

For me, the parameter was added no longer than the past few hours. And the command
```js
admin.remoteConfig().getTemplate()
```

Will fail with the same error described in the issue.

I've also tried downgrading `firebase-admin` to version `9.2.0`, and `9.0.0` but doesn't really help.

VGJohn:
Seeing this issue today as well

ryanoboril:
Same here, issue appeared overnight with no code changes or even npm update

lukencrocker:
Same issue here. Appeared overnight with no code changes. Tried all I can think of but still fails when hitting this line "admin.remoteConfig().getTemplate();" Only thing I did for it to start failing is add a new Remote Config setting as described.

moshaheen:
[...]

PR #1089, opened 2020-11-13
fix(rc): Fix Version update time parsing failure

lahirumaramba:
- Current check for update time fails for timestamps with more than 3 milliseconds places. Updated the timestamp validation logic as a quick fix.

Resolves: #1088

### DIFF
--- a/src/remote-config/remote-config.ts
+++ b/src/remote-config/remote-config.ts
@@ -366,16 +366,12 @@ class VersionImpl implements Version {
     // as UTC date strings. If a developer uses a previously obtained template with UTC timestamps
     // we could still validate it below.
     if (typeof version.updateTime !== 'undefined') {
-      if (!validator.isISODateString(version.updateTime) &&
-        !validator.isUTCDateString(version.updateTime)) {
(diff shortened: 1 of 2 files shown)


lahirumaramba:
- Tested with both 3 and 6 milliseconds places in timestamps (updated unit tests)
- Added `release:stage` to trigger integration tests

src/remote-config/remote-config.ts
```diff
@@ -394,4 +390,8 @@ class VersionImpl implements Version {
       updateTime: this.updateTime,
     }
   }
+
+  private isValidTimestamp(timestamp: string): boolean {
+    return validator.isNonEmptyString(timestamp) && (new Date(timestamp)).getTime() > 0;
```
lahirumaramba, line 395: This validation fails for timestamps before `January 1, 1970` and considers strings such as `"1.2"` as valid timestamps. I think that would be okay for this particular scenario. Open to suggestions!

hiranya911 (approved):
Looks good. I'd recommend adding another test case for `getTemplate()` so that both old and new date formats get tested.

PR #1097, opened 2020-11-24
fix(fcm): Support arbitrary custom values in the ApnsPayload

### LINKED ISSUE #1023: Incorrect typings for ApnsPayload
`ApnsPayload` is typed as:

```ts
export interface ApnsPayload {

  /**
   * The `aps` dictionary to be included in the message.
   */
  aps: Aps;
  [customData: string]: object;
}
```

However, that prevents from passing string values as custom data.

Note that according to Apple

> Your custom keys must have values with primitive types, such as dictionary, array, string, number, or Boolean. Custom keys are available in the userInfo dictionary of the UNNotificationContent object delivered to your app.

So I think the correct types should be something like either:

```ts
[customData: string]: any;
```
or
```ts
[customData: string]: JSON;

interface JSON {
    [x: string]: string|number|boolean|Date|JSON|JSONArray;
}
interface JsonArray extends Array<string|number|boolean|Date|JSON|JSONArray> { }
```

hiranya911:
`ApnsPayload` can contain any valid JSON values.

Resolves #1023

### DIFF
--- a/etc/firebase-admin.api.md
+++ b/etc/firebase-admin.api.md
@@ -625,7 +625,7 @@ export namespace messaging {
     }
     export interface ApnsPayload {
         // (undocumented)
-        [customData: string]: object;
+        [customData: string]: any;
         aps: Aps;
     }
     export interface Aps {
--- a/src/messaging/index.ts
+++ b/src/messaging/index.ts
@@ -299,7 +299,7 @@ export namespace messaging {
      * The `aps` dictionary to be included in the message.
      */
     aps: Aps;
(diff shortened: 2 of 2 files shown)


lahirumaramba (approved):
Do we need any tests for this? Otherwise, LGTM!

hiranya911:
Already covered in https://github.com/firebase/firebase-admin-node/blob/0d7238040a115099c969021fd9d876788b61d88b/test/unit/messaging/messaging.spec.ts#L3267-L3304

Issue #1101, opened 2020-12-01
[FR] docs: Please update comments in source code.


FrancescoJo:
**Is your feature request related to a problem? Please describe.**
https://github.com/firebase/firebase-admin-node/blob/master/src/auth/index.ts#L503

Now in december 2020, Firebase auth supports various third pary login methods more than `anonymous`, `password`, `facebook.com`, `github.com`, `google.com`, `twitter.com`, or `custom`. However, only all these methods that I listed are written in source code comments, which causes confusion.

Of course enumerating all possible tokens like this way is way much better than other sdks(e.g.: [Firebase Java sdk](https://github.com/firebase/firebase-admin-java/blob/master/src/main/java/com/google/firebase/auth/UserInfo.java#L66)), but it needs to be updated. At least in my case, I can confirm there are more providerIds are returned by auth API, such as `apple.com` or `microsoft.com`. 

So this is my little proposal: could you guys update the comments to elaborate all possible providerIds please?

**Describe the solution you'd like**
Comment includes all possible providerId candidates: not in Java sdk way

**Describe alternatives you've considered**
Or sync to Java sdk way which I think the worst solution but it would decrease confusion

**Additional context**
n/a

Thanks for reading.

PR #1106, opened 2020-12-02
chore: adds missing provider IDs for Auth

bojeil-google:
Adds missing token claims `sign_in_provider` values.

Fixes https://github.com/firebase/firebase-admin-node/issues/1101

### DIFF
--- a/src/auth/index.ts
+++ b/src/auth/index.ts
@@ -503,7 +503,13 @@ export namespace auth {
       /**
        * The ID of the provider used to sign in the user.
        * One of `"anonymous"`, `"password"`, `"facebook.com"`, `"github.com"`,
-       * `"google.com"`, `"twitter.com"`, or `"custom"`.
+       * `"google.com"`, `"twitter.com"`, `"apple.com"`, `"microsoft.com"`,
+       * "yahoo.com"`, `"phone"`, `"playgames.google.com"`, `"gc.apple.com"`,
+       * or `"custom"`.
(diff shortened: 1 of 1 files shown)


src/auth/index.ts
```diff
@@ -503,7 +503,13 @@ export namespace auth {
       /**
        * The ID of the provider used to sign in the user.
        * One of `"anonymous"`, `"password"`, `"facebook.com"`, `"github.com"`,
-       * `"google.com"`, `"twitter.com"`, or `"custom"`.
+       * `"google.com"`, `"twitter.com"`, `"apple.com"`, `"microsoft.com"`,
+       * "yahoo.com"`, `"phone"`, `"playgames.google.com"`, `"gc.apple.com"`,
+       * or `"custom"`.
+       *
+       * Additional Identity Platform provider IDs include `"linkedin.com"`,
```
hiranya911, line 510: Should linkedin be listed in the previous paragraph? Why was it mentioned separately?

hiranya911 (approved):

src/auth/index.ts
bojeil-google: LinkedIn is only available via the cloud console UI at the moment. Though I don't believe it is restricted to GCIP. In theory it could be enabled via REST API for Firebase. I can move it to above but worry Firebase developers may get confused when they don't see it there.

kevinthecheung (approved):

Issue #903, opened 2020-06-13
Service Account json does not match type ServiceAccount required for admin.credential.cert()


paulrenshaw:
## Environment

  * Operating System version: macOS 10.15.5 (Catalina)
  * Firebase SDK version: 8.12.1
  * Firebase Product: firebase-admin
  * Node.js version: 10.20.1
  * NPM version: 6.14.4

## Problem

When using TypeScript in VS Code and trying to pass a service account object imported from a .json file downloaded from the firebase console, serviceAccount is underlined in the following code 

```
import admin from 'firebase-admin'
import serviceAccount from './serviceAccount.json'

if (!admin.apps.length) {
  admin.initializeApp({
    credential: admin.credential.cert(serviceAccount)
  })
}
```

with problem 

```
Type '{ type: string; project_id: string; private_key_id: string; private_key: string; client_email: string; client_id: string; auth_uri: string; token_uri: string; auth_provider_x509_cert_url: string; client_x509_cert_url: string; }' has no properties in common with type 'ServiceAccount'.
```

the type ServiceAccount is defined in firebase-admin/lib/index.d.ts {} admin as 

```
interface ServiceAccount {
    projectId?: string;
    clientEmail?: string;
    privateKey?: string;
  }
```

## Steps to reproduce:

install firebase-admin
download service account json file from Firebase Console
import service account json file using typescript (with resolveJsonModule flag)
attempt to pass the service account variable to admin.credential.cert() to be used to initialise

hiranya911:
You have to use the `require('path/to/serviceAccount')` syntax. If you import as shown above then you must manually convert it into the interface format expected by the SDK.

Issue #624, opened 2019-08-16
[TypeScript] Initialize types too strict


TheDuc:
### Describe your environment

  * Firebase SDK version: 8.3.0
  * Firebase Product: firebase admin
  * Node.js version: 10.16.0
  * NPM version: 6.10.3

### Describe the problem

When initialising firebase admin with standard downloaded service account json, typescript error incompatible types when `"resolveJsonModule": true,` in compiler

#### Relevant Code:

```
import * as admin from 'firebase-admin';

import * as serviceAccount from "path/to/serviceAccountKey.json";

admin.initializeApp({
  credential: admin.credential.cert(serviceAccount),     // <---- type error here
  databaseURL: "https://<DATABASE>.firebaseio.com"
});
```
#### **`tsconfig.json`**
```
"compilerOptions": {
    (...),
    "resolveJsonModule": true,
  },
```

hiranya911:
Duplicate of #522. See the explanations and workarounds discussed there.

Issue #522, opened 2019-05-07
ServiceAccount interface definition errors


dgobaud:
the ServiceAccount interface specifies the member variables using camel case but they come in from Google Cloud JSON download as snake case and if changed in the JSON file to match the interface, when exporting the JSON file path in GOOGLE_APPLICATION_CREDENTIALS to work with other Google Cloud services locally in the emulator https://firebase.google.com/docs/functions/local-emulator#set_up_admin_credentials_optional it gives error `{"result":"The incoming JSON object does not contain a client_email field"}`

https://github.com/firebase/firebase-admin-node/blob/2e48327390dd25971944a066b4246c5eef8302cd/src/index.d.ts#L35

```
  interface ServiceAccount {
    projectId?: string;
    clientEmail?: string;
    privateKey?: string;
  }
```

hiranya911:
`ServiceAccount` interface is used by the `admin.credentials.cert()` function. And that function is programmed to accept both camel case and snake case attributes. Therefore you don't have to modify downloaded service account JSON files to satisfy the interface. If you do it will most certainly break the OAuth flow. 

One of the following should work for pretty much any use case:

```
// loading from file system
admin.credentials.cert('path/to/file.json');
```

```
// loading from the env (json contents in variable)
admin.credential.cert(process.env['GOOGLE_APPLICATION_CREDENTIALS']);
```

```
// loading from the env (file path in variable)
const obj = require(process.env['GOOGLE_APPLICATION_CREDENTIALS']);
admin.credential.cert(obj);
```

MateusZitelli:
@hiranya911 - I am facing this issue when trying something similar to the last code snipped you posted.

```
import * as serviceAccount from "./serviceAccountKey.json";
admin.credential.cert(serviceAccount)
```

Here is the linter message:
```
Argument of type '{ "type": string; "project_id": string; "private_key_id": string; "private_key": string; "client_email": string; "client_id": string; "auth_uri": string; "token_uri": string; "auth_provider_x509_cert_url": string; "client_x509_cert_url": string; }' is not assignable to parameter of type 'string | ServiceAccount'.
  Type '{ "type": string; "project_id": string; "private_key_id": string; "private_key": string; "client_email": string; "client_id": string; "auth_uri": string; "token_uri": string; "auth_provider_x509_cert_url": string; "client_x509_cert_url": string; }' has no properties in common with type 'ServiceAccount'.
```

[...]

PR #1110, opened 2020-12-07
Fix type for cert(...) function, to support standard downloadable serviceAccount.json

AjahnCharles:
### Motivation

The JavaScript implementation of cert(...) **already supports snake_case** account objects. I suspect this is the most popular format for non-Cloud-Function usage since **GCP & Firebase console provide downloadable account configs as snake_case**.

Every single TypeScript client of the Node Admin SDK must navigate this type error, which occurs at the very start of their SDK journey. It has already been raised as an issue 3 separate times. It's a trivial fix that increases the accuracy & readability of the codebase.

### Issues

  * https://github.com/firebase/firebase-admin-node/issues/522
  * https://github.com/firebase/firebase-admin-node/issues/624
  * https://github.com/firebase/firebase-admin-node/issues/903

### Testing

  * Make sure all existing tests pass => *crosses fingers*
  * If you fixed a bug ... add a new test => n/a (compile-time)

### API Changes

  * This isn't a change; rather the API is already supporting snake_case and the TypeScript description is simply deficient

### DIFF
--- a/etc/firebase-admin.api.md
+++ b/etc/firebase-admin.api.md
@@ -395,7 +395,7 @@ export namespace auth {
 // @public (undocumented)
 export namespace credential {
     export function applicationDefault(httpAgent?: Agent): Credential;
-    export function cert(serviceAccountPathOrObject: string | ServiceAccount, httpAgent?: Agent): Credential;
+    export function cert(serviceAccountPathOrObject: string | ServiceAccount | ServiceAccountJson, httpAgent?: Agent): Credential;
     export interface Credential {
         getAccessToken(): Promise<GoogleOAuthAccessToken>;
     }
@@ -1042,12 +1042,22 @@ export namespace securityRules {
 
 // @public (undocumented)
 export interface ServiceAccount {
-    // (undocumented)
-    clientEmail?: string;
-    // (undocumented)
-    privateKey?: string;
-    // (undocumented)
-    projectId?: string;
+  // (undocumented)
+  clientEmail?: string;
+  // (undocumented)
+  privateKey?: string;
+  // (undocumented)
+  projectId?: string;
+}
+
+// @public (undocumented)
+export interface ServiceAccountJson {
+  // (undocumented)
+  client_email?: string;
+  // (undocumented)
+  private_key?: string;
+  // (undocumented)
+  project_id?: string;
 }
 
 // @public
--- a/etc/firebase-admin.api.md
+++ b/etc/firebase-admin.api.md
@@ -1042,22 +1042,22 @@ export namespace securityRules {
 
 // @public (undocumented)
 export interface ServiceAccount {
-  // (undocumented)
-  clientEmail?: string;
-  // (undocumented)
-  privateKey?: string;
-  // (undocumented)
-  projectId?: string;
+    // (undocumented)
+    clientEmail?: string;
+    // (undocumented)
+    privateKey?: string;
+    // (undocumented)
+    projectId?: string;
 }
 
 // @public (undocumented)
 export interface ServiceAccountJson {
-  // (undocumented)
-  client_email?: string;
-  // (undocumented)
-  private_key?: string;
-  // (undocumented)
-  project_id?: string;
+    // (undocumented)
+    client_email?: string;
+    // (undocumented)
+    private_key?: string;
+    // (undocumented)
+    project_id?: string;
 }
 
 // @public
--- a/src/credential/index.ts
+++ b/src/credential/index.ts
@@ -23,6 +23,15 @@ export interface ServiceAccount {
 }
 
 /**
+ * Interface matching downloadable serviceAccount.json
+ */
+export interface ServiceAccountJson {
+  project_id?: string;
+  client_email?: string;
+  private_key?: string;
+}
+
+/**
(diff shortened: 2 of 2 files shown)


AjahnCharles:
@googlebot I signed it!

charles-allen:
@googlebot I fixed it

hiranya911:
Been testing this solution and seems alright to me. I was a bit concerned about tsc potentially flagging an object that has more properties than the declared interface (since service account json contains many other fields), but that doesn't seem to be the case. It still needs to go through our internal API review process. I'll find some time for that in the coming weeks.

charles-allen:
I PR'd this implementation because it's a mirror of the existing camelCase interface (and I thought that would be the smallest possible increment). If you're willing or even prefer to make a larger change, I think the best option is a complete & accurate interface that explicitly describes the expectation of the API!

If you're interested in a more accurate (& hence more powerful) solution, the best case is probably an interface that explicitly matches the expectations of the API (i.e everything that's required from a snake_case JSON would be required, and all the optional properties would be list & explicitly marked optional):
```
export interface ServiceAccountJson {
  type?: string
  project_id: string // required
  private_key_id?: string
  private_key: string // required
  client_email: string // required
  client_id?: string
  auth_uri?: string
  token_uri?: string
  auth_provider_x509_cert_url?: string
  client_x509_cert_url?: string
}
```

## Testing

I wrote previously that I couldn't write a test, but actually I think we could add a small dummy TypeScript program that imports the SDK (perhaps similar to the code above), and add a step to the GitHub actions workflow to compile that program (compilation failure would fail the build).

> I was a bit concerned about tsc potentially flagging an object that has more properties than the declared interface (since service account json contains many other fields)

From the [TypeScript docs](https://www.typescriptlang.org/docs/handbook/interfaces.html#excess-property-checks) - TL;DR: the "additional property check" is only when assigning an object literal:

> Object literals get special treatment and undergo excess property checking when assigning them to other variables, or passing them as arguments. If an object literal has any properties that the “target type” doesn’t have, you’ll get an error

There might be cases where a client constructs their config inline (e.g. case 3):
```
// case 1: static import
// type inferred from JSON itself => checked structurally vs the interface
// not passing a literal => excess property check skipped
import serviceAccount from 'service-account.json'
admin.initializeApp({ credential: admin.credential.cert(serviceAccount) })

// case 2: dynamic import
// type asserted => actual JSON is never checked
// not passing a literal => excess property check skipped
const serviceAccount: ServiceAccountJson = await import(configPath) // or require(...)
admin.initializeApp({ credential: admin.credential.cert(serviceAccount) })

// case 3
// type inferred from the literal => checked structurally vs the interface
// passing a literal => excess property check might fail
admin.initializeApp({ credential: admin.credential.cert({
  project_id = process.env.FIREBASE_PROJECT_ID,
  private_key = process.env.FIREBASE_PRIVATE_KEY,
  client_email = process.env.FIREBASE_PROJECT_ID,
  another_property = process.env.FIREBASE_ANOTHER_PROPERTY // might fail on this line
}) })
```

hiranya911:
> I wrote previously that I couldn't write a test, but actually I think we could add a small dummy TypeScript program that imports the SDK (like the cases below), and add a step to the GitHub actions workflow to compile that program (compilation failure would fail the build).

I think you can (and should) add some unit tests to the existing test suite. Something similar to:

https://github.com/firebase/firebase-admin-node/blob/1862342636e816b61337262254cdfe0cc410c640/test/unit/firebase.spec.ts#L114-L117

You can convert the `mocks.certificateObject` into an instance of `ServiceAccountJson` and pass that into `cert`.

charles-allen:
> I think you can (and should) add some unit tests to the existing test suite. Something similar to:
> 
> https://github.com/firebase/firebase-admin-node/blob/1862342636e816b61337262254cdfe0cc410c640/test/unit/firebase.spec.ts#L114-L117
> 
> You can convert the `mocks.certificateObject` into an instance of `ServiceAccountJson` and pass that into `cert`.

I'll see if I can add something along those lines that verifies a snake_case config can be imported in all the standard ways, passed to cert, and most importantly that tsc is happy about it (the tests will fail the build step if this isn't the case).

charles-allen:
Sorry for the delay. Busy with work. I should have time to revisit this this week :)

paul-uz:
@charles-allen any update on this? I encountered this issue just this week and it was bugging the heck out of me, pardon the pun!

charles-allen:
> @charles-allen any update on this? I encountered this issue just this week and it was bugging the heck out of me, pardon the pun!

I got sidetracked by work. But the next 2 weeks is the perfect time for me to revive it. I think I just need to copy the existing tests and swap camel for snake case.

Can't offer any guarantees on how much more time it would take to get accepted though.

paul-uz:
@charles-allen no worries :)

charles-allen:
> @charles-allen no worries :)

Thanks for bumping it :)

tylim88:
it is almost 2 years, any update?

charles-allen:
> it's been almost 2 years, any update?

I'm not working with Firebase any more, and I think Hiranya has left the Firebase team. This PR is probably horribly out-of-date. You're welcome to replace it with a new PR :)

PR #1128, opened 2021-01-06
fix: Removing FirebaseServiceInterface and FirebaseServiceInternalsInterface

hiranya911:
As a first step towards modularization I'm removing the above 2 internal interfaces. Classes like `Auth` and `RemoteConfig` will no longer have to explicitly implement `FirebaseServiceInterface` or implement a placeholder "Internals" class to support state tear down at app delete. Classes like `DatabaseService` that has internal state that should be cleaned up can simply implement a `delete()` method. The `FirebaseApp` class will look for this method and call it if available during app delete.

At the same time I'm also removing some old, unused code for registering (via monkey patch) services with `FirebaseNamespace` and calling app hooks. Both these functions haven't been used for quite some time, and now that we are looking at modularization, they won't be useful in the future either.


src/firebase-app.ts
```diff
@@ -428,51 +422,24 @@ export class FirebaseApp implements app.App {
...
 
-    return this.services_[serviceName];
-  }
-
-  /**
-   * Callback function used to extend an App instance at the time of service instance creation.
-   */
-  private extendApp_(props: {[prop: string]: any}): void {
-    deepExtend(this, props);
+    return this.services_[serviceName] as T;
```
bojeil-google, line 442: After you removed the tests in `describe('#[service]()', () => {...}` below, do you still have tests to confirm this, eg. the cached service is returned?

bojeil-google (approved):

src/firebase-app.ts
hiranya911: The removed tests were only verifying the caching of monkey-patched services. We don't have any monkey-patched services in the SDK. There are some tests in `firebase-app.spec.ts` to test for caching behavior of actual services like Auth. I've now added some new tests to `firebase-namespace.spec.ts` to verify the same at the FirebaseNamespace level.

test/unit/firebase-namespace.spec.ts
```diff
@@ -349,6 +349,13 @@ describe('FirebaseNamespace', () => {
     it('should return a reference to Auth type', () => {
       expect(firebaseNamespace.auth.Auth).to.be.deep.equal(AuthImpl);
     });
+
+    it('should return a cached version of Auth on subsequent calls', () => {
+      firebaseNamespace.initializeApp(mocks.appOptions);
+      const serviceNamespace1: Auth = firebaseNamespace.auth();
+      const serviceNamespace2: Auth = firebaseNamespace.auth();
+      expect(serviceNamespace1).to.deep.equal(serviceNamespace2);
```
bojeil-google, line 357: all the `to.deep.equal` should be replaced with `to.equal`

bojeil-google (approved):

test/unit/firebase-namespace.spec.ts
hiranya911: Thanks. Done.

lahirumaramba (approved):
Thanks!

PR #1132, opened 2021-01-12
fix: Added lightweight App interface

hiranya911:
As part of the modularization effort we need to introduce a new lightweight `App` interface that doesn't have any dependencies on other services of the SDK. This PR adds this new interface with the following signature:

```
export interface App {
    delete(): Promise<void>;
    name: string;
    options: AppOptions;
}
```

We don't necessarily have to export it from firebase-namespace right now, but doing so results in a more complete API Extractor report.

NOTE: PR is against a new `modular-sdk` branch, where we will continue to develop this feature.


lahirumaramba (approved):
LGTM!

PR #1134, opened 2021-01-13
feat: Added firebase-admin/app submodule

hiranya911:
Introduces the `firebase-admin/app` module which exports the core types and the SDK initialization APIs. In order to maintain backwards compatibility, I've kept the existing `firebase-app` and `firebase-namespace` implementations as is, and just moved them under `app/`. The new `app/lifecycle.ts` file instantiates the default `FirebaseNamespace` instance, which is then exported by top-level `default-namespace.ts`.  This makes sure that both of the following styles work:

```ts
// old style
// Invocation chain: index.js --> default-namespace.js --> app/lifecycle.js
import * as admin from 'firebase-admin';

admin.initializeApp();
```

```ts
// new style
// Invocation chain: app/index.js --> app/lifecycle.js
import { initializeApp } from 'firebase-admin/app';

initializeApp();
```

Other noteworthy changes:
* Moved the `delete()` method from the `App` interface to the old `app.App` interface. The modularized `App` interface will not expose a `delete()` method.
* Added `getApp()`, `getApps()` and `deleteApp()` functions to the new modules. These functions haven't gone through our API review yet, but will be required to support a fully modularized dev experience (JS SDK is adding a similar API).


test/unit/app/index.spec.ts
```diff
@@ -0,0 +1,183 @@
+/*!
+ * @license
+ * Copyright 2017 Google Inc.
```
lahirumaramba, line 3: This is a new file, isn't it?

lahirumaramba (approved):
LGTM

test/unit/app/index.spec.ts
hiranya911: Updated

PR #1136, opened 2021-01-15
feat: Added instanceId() modularized API

hiranya911:
* Added `instanceId()` module-level function.
* Added `firebase-admin/instance-id` module.

### DIFF
--- a/etc/firebase-admin.api.md
+++ b/etc/firebase-admin.api.md
@@ -498,15 +498,18 @@ export interface GoogleOAuthAccessToken {
 export function initializeApp(options?: AppOptions, name?: string): app.App;
 
 // @public
-export function instanceId(app?: app.App): instanceId.InstanceId;
+export class InstanceId {
+    get app(): App;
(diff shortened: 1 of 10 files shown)


etc/firebase-admin.api.md
```diff
@@ -498,15 +498,18 @@ export interface GoogleOAuthAccessToken {
 export function initializeApp(options?: AppOptions, name?: string): app.App;
 
 // @public
-export function instanceId(app?: app.App): instanceId.InstanceId;
+export class InstanceId {
+    get app(): App;
+    deleteInstanceId(instanceId: string): Promise<void>;
+    }
```
lahirumaramba, line 504: This is generated from the api-extractor, isn't it? Is there an indentation issue here?

lahirumaramba (approved):

etc/firebase-admin.api.md
hiranya911: Yes, this is auto-generated. Not sure what we can do about the indentation issue. Everything looks good in source and the generated d.ts files.

PR #1137, opened 2021-01-19
feat: Moved the credential APIs under firebase-admin/app

hiranya911:
* Exposed `ServiceAccount`, `GoogleOAuthAccessToken`, `cert()`, `applicationDefault()` and `refreshToken()` APIs from the `app` submodule.
* Moved credential implementations under the same submodule.
* Aliased the new APIs from the existing `admin.credential` namespace for backward compatibility.

### DIFF
--- a/etc/firebase-admin.api.md
+++ b/etc/firebase-admin.api.md
@@ -45,9 +45,12 @@ export namespace app {
     }
 }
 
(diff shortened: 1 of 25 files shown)


lahirumaramba (approved):
LGTM!

PR #1139, opened 2021-01-20
fix(messaging): Fix error throw when data property exists but is undefined

italoiz:
When use `messaging().sendToDevice(tokens, payload)` where payload contains the `data` property, but it's contains `undefined` value, I given throws the follow error:

```js
TypeError: Cannot use 'in' operator to search for 'from' in undefined
```

The following Pull Request contains the solution of the issue with the covered unit tests.

### DIFF
--- a/src/messaging/messaging.ts
+++ b/src/messaging/messaging.ts
@@ -857,7 +857,7 @@ export class Messaging implements MessagingInterface {
     }
 
     // Validate the data payload object does not contain blacklisted properties
-    if ('data' in payloadCopy) {
+    if ('data' in payloadCopy && payloadCopy.data !== undefined) {
       BLACKLISTED_DATA_PAYLOAD_KEYS.forEach((blacklistedKey) => {
         if (blacklistedKey in payloadCopy.data!) {
           throw new FirebaseMessagingError(
--- a/test/unit/messaging/messaging.spec.ts
+++ b/test/unit/messaging/messaging.spec.ts
@@ -566,6 +566,12 @@ describe('Messaging', () => {
         { token: 'mock-token' },
       ).should.eventually.be.rejected.and.have.property('code', 'messaging/invalid-argument');
     });
+
+    it('should not throws when the payload contains the data property with undefined value', () => {
+      mockedRequests.push(mockSendRequest());
+      const message = { token: 'a', data: undefined };
+      return messaging.send(message).should.eventually.equal('projects/projec_id/messages/message_id');
+    })
   });
 
   describe('sendAll()', () => {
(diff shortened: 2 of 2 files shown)


aprilmintacpineda:
This has recently happened to me.

src/messaging/messaging.ts
```diff
@@ -857,7 +857,7 @@ export class Messaging implements MessagingInterface {
     }
 
     // Validate the data payload object does not contain blacklisted properties
-    if ('data' in payloadCopy) {
+    if ('data' in payloadCopy && payloadCopy.data !== undefined) {
```
hiranya911, line 860: May be just `if (payloadCopy.data)`

test/unit/messaging/messaging.spec.ts
```diff
@@ -566,6 +566,12 @@ describe('Messaging', () => {
         { token: 'mock-token' },
       ).should.eventually.be.rejected.and.have.property('code', 'messaging/invalid-argument');
     });
+
+    it('should not throws when the payload contains the data property with undefined value', () => {
```
hiranya911, line 570: should not throw when

hiranya911, line 570: This test case is actually irrelevant to the actual fix. The fix only applies to legacy FCM APIs like `sendToDevice()`. It has no effect on `send()`. We can remove this.

test/unit/messaging/messaging.spec.ts
```diff
@@ -823,6 +829,29 @@ describe('Messaging', () => {
         [validMessage],
       ).should.eventually.be.rejected.and.have.property('code', 'app/invalid-credential');
     });
+
+    it('should not throws when the payload contains the data property with undefined value', () => {
```
hiranya911, line 833: Please remove. Not relevant to the fix.

test/unit/messaging/messaging.spec.ts
```diff
@@ -1118,6 +1147,31 @@ describe('Messaging', () => {
       ).should.eventually.be.rejected.and.have.property('code', 'app/invalid-credential');
     });
 
+    it('should not throws when the payload contains the data property with undefined value', () => {
```
hiranya911, line 1150: Please remove.

test/unit/messaging/messaging.spec.ts
```diff
@@ -1457,6 +1511,16 @@ describe('Messaging', () => {
         expect(mockOptionsClone).to.deep.equal(mocks.messaging.options);
       });
     });
+
+    it('should not throws when the payload contains the data property with undefined value', () => {
```
hiranya911, line 1515: should not throw when...

test/unit/messaging/messaging.spec.ts
```diff
@@ -1457,6 +1511,16 @@ describe('Messaging', () => {
...
+      mockedRequests.push(mockSendToDeviceStringRequest());
+
+      return messaging.sendToDevice(
+        mocks.messaging.registrationToken,
+        { ...mocks.messaging.payload, data: undefined },
+        mocks.messaging.options,
+      );
```
hiranya911, line 1522: We need an assertion here. May be at least `.should.eventually.be.fulfilled`.

hiranya911 (changes requested):
Sorry for the delay. This PR somehow slipped through the cracks. It looks mostly good, but some of the tests are not relevant to the fix. And the others require some assertions. I can take another look, if those issues are addressed.

italoiz:
Hey @hiranya911, Is this PR still need a fix?

PR #1140, opened 2021-01-22
feat(auth): Modularized firebase-admin/auth API entrypoint

hiranya911:
Implemented the modular `firebase-admin/auth` API entrypoint. All public types are now directly exported from this submodule, along with the old `auth` namespace. One key difference compared to the old way of doing things is that we are now directly exporting the classes like `Auth` and `UserRecord`, instead of wrapping them in interfaces. Aliases like `admin.auth.Auth` reference the corresponding class directly.

I also found a few cyclic imports in the process, and fixed them by introducing a new `base-auth.ts` module. The `BaseAuth` class and several other core interface definitions were moved to this module.

One of the issues I faced was modules like `auth-api-request` exporting their internal members. API Extractor flagged them with errors, and I've fixed them for now by adding the `@internal` annotation. There's a longer conversation to be had here about what's the best way to deal with such _internally public_ members, but we can do it towards the end of the modularization effort. I'll work on a doc outlining our options.

**Note:** The API documentation needs more work to catch up to our old standards. I will do so in a future PR.


lahirumaramba (approved):
Looks great!

src/auth/identifier.ts
```diff
@@ -14,13 +14,48 @@
  * limitations under the License.
  */
 
-import { auth } from './index';
+/**
+   * Used for looking up an account by uid.
```
bojeil-google, line 18: alignment is off

src/auth/tenant.ts
```diff
@@ -46,11 +77,12 @@ export interface TenantServerResponse {
...
  * Tenant class that defines a Firebase Auth tenant.
  */
-export class Tenant implements TenantInterface {
+export class Tenant {
+
   public readonly tenantId: string;
   public readonly displayName?: string;
-  public readonly emailSignInConfig?: EmailSignInConfig;
-  public readonly multiFactorConfig?: MultiFactorAuthConfig;
+  private readonly _emailSignInConfig?: EmailSignInConfig;
```
bojeil-google, line 84: You seem to use trailing underscore for private variables above, eg. `app_`. It would be a good idea to be consistent (either trailing or leading).

bojeil-google (approved):

src/auth/tenant.ts
hiranya911: Thanks. Good point and fixed.

PR #1142, opened 2021-01-25
fix: Renaming service functions to getAuth and getInstanceId

hiranya911:
* Renaming the service accessor functions to be consistent with the approved API proposal.
* Added some unit tests for the `getAuth()` function.


lahirumaramba (approved):
LGTM!

PR #1143, opened 2021-01-27
feat(rtdb): Added firebase-admin/database module entrypoint

hiranya911:
* Added `getDatabase()` and `getDatabaseWithUrl()` entry functions.
* Exported `Database` type with the methods for managing security rules.
* Exported other public interface types from `@firebase/database-types` package.
* Exported `enableLogging` and `ServerValue` members from the `@firebase/database` package.

### DIFF
--- a/test/unit/database/index.spec.ts
+++ b/test/unit/database/index.spec.ts
@@ -23,7 +23,8 @@ import * as chaiAsPromised from 'chai-as-promised';
 
 import * as mocks from '../../resources/mocks';
 import { App } from '../../../src/app/index';
-import { getAuth, Auth } from '../../../src/auth/index';
+import { getDatabase, getDatabaseWithUrl, Database } from '../../../src/database/index';
+import { FirebaseApp } from '../../../src/app/firebase-app';
 
 chai.should();
 chai.use(sinonChai);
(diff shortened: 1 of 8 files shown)


test/unit/database/index.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    it('should not throw given a valid app', () => {
+      expect(() => {
+        return getDatabase(mockApp);
+      }).not.to.throw();
+    });
+
+    it('should return the same instance for a given app instance', () => {
+      const db1: Database = getDatabase(mockApp);
+      const db2: Database = getDatabase(mockApp);
+      expect(db1).to.equal(db2);
```
lahirumaramba, line 64: Do we want to test for not equal here similar to `getDatabaseWithUrl ` test below?
```
const db3: Database = getDatabase(mockAppOther);
expect(db1).to.not.equal(db3);
```

lahirumaramba (approved):
LGTM!

test/unit/database/index.spec.ts
hiranya911: For now I'm adding only minimal tests to these entry points, primarily to check the imports work. The functionality of these APIs already get tested extensively via  the old API (`admin.database()` calls `getDatabase()` internally). We can add some more tests to these new modules once the modularization work is nearing end.

PR #1145, opened 2021-01-28
feat: Exporting FCM_MAX_BATCH_SIZE

### LINKED ISSUE #1049: [FR] Expose FCM_MAX_BATCH_SIZE so it can be reused
**Is your feature request related to a problem? Please describe.**
The `FCM_MAX_BATCH_SIZE` is not exported so it can not be reused by consumers of this plugin: https://github.com/firebase/firebase-admin-node/blob/59f2203dc6d0881dce7fc1af28f7b4dfebaa6bc6/src/messaging/messaging.ts#L46

History has shown that the batch size can change. E.g from v8.6.1: https://github.com/firebase/firebase-admin-node/blob/0f91e0280e8e0dda3d4ded975f8f2c71c41d2934/src/messaging/messaging.ts#L43

So after updating this library we had to manually debug to see that we could achieve better performance by increasing our own internal `FCM_MAX_BATCH_SIZE` const (which we copied from this lib at some point) to 500.

**Describe the solution you'd like**
As a consumer of this plugin I'd like to import the batch size from this plugin and potentially split up huge arrays of FCM push batches into multiple parts:

```
import * as fcm from 'firebase-admin';

function sendFcm(messages: Array<fcm.messaging.Message>): void {
  if (messages.size > fcm.messaging.FCM_MAX_BATCH_SIZE) {
    // Do logic to split up messages into multiple batches...
  }
}
```

**Describe alternatives you've considered**
None.

**Additional context**
None.

BorntraegerMarc:
So lib consumers can easily reuse our max batch size

Fixes #1049


hiranya911:
The file you have modified is not part of any public API. How would a developer import it?

BorntraegerMarc:
@hiranya911 I see your point. I made some changes. I believe I understood your point: `messaging.Messaging` is part of the public API. Thus adding the variable in that class will expose it. Correct?

I think I also should add some tests to verify the changes. Same question here as on my other PR: Could you point me to the right place to add the tests? I was thinking under `./test/integration/typescript`

hiranya911:
Ideally it needs to be exposed from the `admin.messaging` namespace. That's the only way a developer can easily get hold of it. But I don't see a straightforward way of doing so due to the way our namespace exports work.

BorntraegerMarc:
Thanks for the explanation @hiranya911 I believe I understood the problem. Indeed: There was no easy way to expose `FCM_MAX_BATCH_SIZE` as a static member. So I decided to make it non-static for now to solve the problem. Which should be good enough for most.

---

On a different note: I have to admit that I've never seen a TS setup like this. Keeping source & typings separate. May I ask why this was done? Is it some sort of technical debt that was never solved?

I believe we wouldn't have had this problem (JS runtime vs TS compiler) if we would just keep TS sources & ship those compiled with auto-generated definition files.

hiranya911:
Thanks for the changes @BorntraegerMarc 

> So I decided to make it non-static for now to solve the problem.

I'm not sure if exposing this on the `Messaging` class/interface is a good idea. The usage of such a value will be extremely awkward and non-idiomatic: `admin.messaging().FCM_MAX_BATCH_SIZE`. The best place for adding this would be the `messaging` namespace, but like I said in an earlier comment it would be tricky to implement it right.

My preference right now is to just go without any new constants for now. Our batch size doesn't change often. In fact since launch (about 2 years ago) it has only changed once. And whenever we do change something like this, we clearly specify that in the release notes, so the developers can update their implementations accordingly.

Having said that I think we can easily introduce a public constant for this once the new modular API surface is available to use (currently a work in progress). That will give us a natural place to define things like `FCM_MAX_BATCH_SIZE`.

> have to admit that I've never seen a TS setup like this. Keeping source & typings separate.

[Ambient typings](https://www.typescriptlang.org/docs/handbook/modules.html#ambient-modules) are fairly common in TS projects -- especially when the codebase is fairly old, and predates TS. In our case we also have used the ambient typings as a way to regulate our public API surface. It gives us a way to control exactly what's visible to the developers and tools. Although, lot of this are changing. As of last year we are shifting towards a one-source-of-truth model where we try to generate typings from the implementation as much as possible. What you see now in the repo is an intermediate state of that evolution, where we can generate typings, but have to keep some public interface definitions separate.

BorntraegerMarc:
> My preference right now is to just go without any new constants for now

Sure. It's understandable. I was also 50/50 whether the awkward usage makes sense.

---

> As of last year we are shifting towards a one-source-of-truth model where we try to generate typings from the implementation as much as possible. What you see now in the repo is an intermediate state of that evolution

Do you know, whether there is an GitHub issue or PR to follow for updates?

hiranya911:
Right now most of the tracking and design work is happening internally. But there is some preliminary work happening in the `modular-sdk` branch of this repo, and you can keep an eye on the various PRs being merged into that branch.

Issue #1146, opened 2021-01-28
[FR] Export Message types so they can be re-used


BorntraegerMarc:
**Is your feature request related to a problem? Please describe.**
Take the following code:
```ts
import { TokenMessage } from 'firebase-admin/lib/messaging';

const allPushTokenSubscriptionMessage2DArray: Array<Array<fcm.messaging.Message>>; // Contains all the TokenMessage's I'd like to send a FCM message to. It's 2D because the max. nr. of tokens that is expected is 500. So if we want to send to 501 devices then the last device is in the next entry of the 2D array.

const batchResponses: Array<fcm.messaging.BatchResponse | null> = await Promise.all(
    allPushTokenSubscriptionMessage2DArray.map(async (pushTokenSubscriptionMessageArray: Array<fcm.messaging.Message>) =>
        this.firebaseMessaging.sendAll(pushTokenSubscriptionMessageArray)
    )
);

batchResponses.forEach((batch: fcm.messaging.BatchResponse, batchIndex: number) => {
    batch.responses.forEach((res: fcm.messaging.SendResponse, responseIndex: number) => {
        // Based on answers here https://github.com/firebase/firebase-admin-node/issues/600 responses should be ordered as requests
        const token: string = (allPushTokenSubscriptionMessage2DArray[batchIndex][responseIndex] as TokenMessage).token;
        console.log(`Failed to send to token ${token}`);
    });
});
```

The problem is now that `TokenMessage` is no longer exported in this lib. So TS transpilation will fail with the error `Cannot find name 'TokenMessage'`.

**Describe the solution you'd like**
Export all types of Messages.

**Describe alternatives you've considered**
None.

**Additional context**
None.

BorntraegerMarc:
> This issue does not seem to follow the issue template. Make sure you provide all the required information.

Sorry, I don't see what part of the template I didn't follow. Maybe I'm missing something. But the description should be pretty explanatory.

PR #1147, opened 2021-01-28
feat: Exporting all types of Messages so they can be used by consumers

BorntraegerMarc:
Fixes https://github.com/firebase/firebase-admin-node/issues/1146


hiranya911 (changes requested):
This looks good. Just needs some tests.

BorntraegerMarc:
Will do @hiranya911. Could you point me to the right place to add the tests? I was thinking under `./test/integration/typescript`

test/unit/typescript.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+import { messaging } from '../../src/messaging/index';
+
+chai.should();
+chai.use(chaiAsPromised);
+
+const expect = chai.expect;
+
+describe('Typescript', () => {
+  // This test was added as part of https://github.com/firebase/firebase-admin-node/issues/1146
+  it('should export all types of Messages so it can be used by lib users', () => {
```
hiranya911, line 32: Let's try to incorporate this test logic to the existing `messaging.spec` file. No need to carve out a new file for this. Easiest would be to list it under the existing test suite for the `send()` method:

```typescript
      it(`should be fulfilled with a message ID given a TokenMessage`, () => {
        const message: TokenMessage = {token: 'test'}
        mockedRequests.push(mockSendRequest());
        return messaging.send(message)
          .should.eventually.equal('projects/projec_id/messages/message_id');
      });

     // Add similar test cases for other message types.
```

Also note that when you add a new test file, you have to also list it in `index.spec.ts`. Otherwise mocha won't run it. But, let's just add this to the existing unit test file.

hiranya911 (changes requested):
Thanks for the test cases @BorntraegerMarc. Just one comment on how to structure the new test logic. We can merge this once that's addressed.

test/unit/typescript.spec.ts
hiranya911: Or alternatively you can have one test case in the suite for `sendAll()`:

```typescript
    it('should be fulfilled when called with different message types', () => {
      const messageIds = [
        'projects/projec_id/messages/1',
        'projects/projec_id/messages/2',
        'projects/projec_id/messages/3',
      ];
      const tokenMessage: TokenMessage = {token: 'test'};
      const topicMessage: TopicMessage = {topic: 'test'};
      const conditionMessage: ConditionMessage = {condition: 'test'};
      const messages: Message[] = [tokenMessage, topicMessage, conditionMessage];
      mockedRequests.push(mockBatchRequest(messageIds));
      return messaging.sendAll(messages)
        .then((response: BatchResponse) => {
          expect(response.successCount).to.equal(3);
          expect(response.failureCount).to.equal(0);
          response.responses.forEach((resp, idx) => {
            expect(resp.success).to.be.true;
            expect(resp.messageId).to.equal(messageIds[idx]);
            expect(resp.error).to.be.undefined;
          });
        });
    });
```

BorntraegerMarc: Makes sense. Thx for the guidance!

---

> Also note that when you add a new test file, you have to also list it in index.spec.ts. Otherwise mocha won't run it

FYI mocha _did_ run the new test. I also verified when I added the new file 🙂 

Screenshot 2021-02-04 at 08 43 04

Source: https://github.com/firebase/firebase-admin-node/runs/1820854315

BorntraegerMarc:
Would it be possible to release a new version after this PR is merged?

I'd appreciate it, so we can update the lib in our project.

hiranya911 (approved):
Thanks. LGTM 👍 

Will be included in the next release (most likely next week).

PR #1148, opened 2021-01-28
Improve token verification logic with Auth Emulator.

yuchenshi:
### Discussion

This PR implements http://go/firebase-auth-emulator-admin-sdk (Google internal design doc). Token verification now works everywhere, not just in functions emulator.

### Testing

Ran `node ~/w/firebase-tools/lib/bin/firebase.js emulators:exec --project fake-project-id --only auth 'npx mocha test/integration/auth.spec.ts --slow 5000 --timeout 20000 --require ts-node/register'` manually and all tests passed locally (once unrelated pending fixes in the Auth Emulator such as https://github.com/firebase/firebase-tools/pull/3091 and https://github.com/firebase/firebase-tools/pull/3088 are applied).

### API Changes

(secret API removed; no public API changes)

### DIFF
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -115,10 +115,11 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
    *     verification.
    */
   public verifyIdToken(idToken: string, checkRevoked = false): Promise<DecodedIdToken> {
-    return this.idTokenVerifier.verifyJWT(idToken)
+    const isEmulator = useEmulator();
+    return this.idTokenVerifier.verifyJWT(idToken, isEmulator)
       .then((decodedIdToken: DecodedIdToken) => {
         // Whether to check if the token was revoked.
-        if (checkRevoked || useEmulator()) {
+        if (checkRevoked || isEmulator) {
           return this.verifyDecodedJWTNotRevoked(
             decodedIdToken,
             AuthClientErrorCode.ID_TOKEN_REVOKED);
@@ -443,10 +444,11 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
    */
   public verifySessionCookie(
     sessionCookie: string, checkRevoked = false): Promise<DecodedIdToken> {
-    return this.sessionCookieVerifier.verifyJWT(sessionCookie)
+    const isEmulator = useEmulator();
+    return this.sessionCookieVerifier.verifyJWT(sessionCookie, isEmulator)
       .then((decodedIdToken: DecodedIdToken) => {
         // Whether to check if the token was revoked.
-        if (checkRevoked || useEmulator()) {
+        if (checkRevoked || isEmulator) {
           return this.verifyDecodedJWTNotRevoked(
             decodedIdToken,
             AuthClientErrorCode.SESSION_COOKIE_REVOKED);
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -118,12 +118,12 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
     return this.idTokenVerifier.verifyJWT(idToken)
       .then((decodedIdToken: DecodedIdToken) => {
         // Whether to check if the token was revoked.
-        if (!checkRevoked && !useEmulator()) {
-          return decodedIdToken;
+        if (checkRevoked || useEmulator()) {
+          return this.verifyDecodedJWTNotRevoked(
+            decodedIdToken,
+            AuthClientErrorCode.ID_TOKEN_REVOKED);
         }
-        return this.verifyDecodedJWTNotRevoked(
-          decodedIdToken,
-          AuthClientErrorCode.ID_TOKEN_REVOKED);
+        return decodedIdToken;
       });
   }
 
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -446,12 +446,12 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
     return this.sessionCookieVerifier.verifyJWT(sessionCookie)
       .then((decodedIdToken: DecodedIdToken) => {
         // Whether to check if the token was revoked.
-        if (!checkRevoked) {
-          return decodedIdToken;
+        if (checkRevoked || useEmulator()) {
+          return this.verifyDecodedJWTNotRevoked(
+            decodedIdToken,
+            AuthClientErrorCode.SESSION_COOKIE_REVOKED);
         }
-        return this.verifyDecodedJWTNotRevoked(
-          decodedIdToken,
-          AuthClientErrorCode.SESSION_COOKIE_REVOKED);
+        return decodedIdToken;
       });
   }
 
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -29,7 +29,7 @@ import * as utils from '../utils/index';
 import * as validator from '../utils/validator';
 import { auth } from './index';
 import {
-  FirebaseTokenVerifier, createSessionCookieVerifier, createIdTokenVerifier, ALGORITHM_RS256
+  FirebaseTokenVerifier, createSessionCookieVerifier, createIdTokenVerifier
 } from './token-verifier';
 import {
   SAMLConfig, OIDCConfig, OIDCConfigServerResponse, SAMLConfigServerResponse,
@@ -118,7 +118,7 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
     return this.idTokenVerifier.verifyJWT(idToken)
       .then((decodedIdToken: DecodedIdToken) => {
         // Whether to check if the token was revoked.
-        if (!checkRevoked) {
+        if (!checkRevoked && !useEmulator()) {
           return decodedIdToken;
         }
         return this.verifyDecodedJWTNotRevoked(
@@ -674,28 +674,6 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
         // All checks above passed. Return the decoded token.
         return decodedIdToken;
       });
-  }
-
-  /**
-   * Enable or disable ID token verification. This is used to safely short-circuit token verification with the
-   * Auth emulator. When disabled ONLY unsigned tokens will pass verification, production tokens will not pass.
-   *
-   * WARNING: This is a dangerous method that will compromise your app's security and break your app in
-   * production. Developers should never call this method, it is for internal testing use only.
-   *
-   * @internal
-   */
-  // @ts-expect-error: this method appears unused but is used privately.
-  private setJwtVerificationEnabled(enabled: boolean): void {
-    if (!enabled && !useEmulator()) {
-      // We only allow verification to be disabled in conjunction with
-      // the emulator environment variable.
-      throw new Error('This method is only available when connected to the Authentication emulator.');
-    }
-
-    const algorithm = enabled ? ALGORITHM_RS256 : 'none';
-    this.idTokenVerifier.setAlgorithm(algorithm);
-    this.sessionCookieVerifier.setAlgorithm(algorithm);
   }
 }
 
--- a/src/auth/token-verifier.ts
+++ b/src/auth/token-verifier.ts
@@ -216,9 +216,8 @@ export class FirebaseTokenVerifier {
       return Promise.reject(new FirebaseAuthError(AuthClientErrorCode.INVALID_ARGUMENT, errorMessage));
     }
 
-    // When the algorithm is set to 'none' there will be no signature and therefore we don't check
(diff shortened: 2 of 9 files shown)


src/auth/auth.ts
```diff
@@ -118,7 +118,7 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
     return this.idTokenVerifier.verifyJWT(idToken)
       .then((decodedIdToken: DecodedIdToken) => {
         // Whether to check if the token was revoked.
-        if (!checkRevoked) {
+        if (!checkRevoked && !useEmulator()) {
```
hiranya911, line 121: Let's flip the condition for clarity:

```
if (checkRevoked || useEmulator()) {
  return this.verifyDecodedJWTNottRevoked();
}

return decodedIdToken;
```

src/auth/token-verifier.ts
```diff
@@ -185,7 +178,7 @@ export class FirebaseTokenVerifier {
     if (!fullDecodedToken) {
       errorMessage = `Decoding ${this.tokenInfo.jwtName} failed. Make sure you passed the entire string JWT ` +
         `which represents ${this.shortNameArticle} ${this.tokenInfo.shortName}.` + verifyJwtTokenDocsMessage;
-    } else if (typeof header.kid === 'undefined' && this.algorithm !== 'none') {
+    } else if (!useEmulator() && typeof header.kid === 'undefined') {
```
hiranya911, line 181: May be call this function once in the constructor and assign the result to a private member? No need to keep calling it repeatedly afaict.

test/integration/auth.spec.ts
```diff
@@ -37,6 +37,8 @@ chai.use(chaiAsPromised);
 
 const expect = chai.expect;
 
+const authEmulatorHost = process.env.FIREBASE_AUTH_EMULATOR_HOST;
```
hiranya911, line 40: We primarily run integration tests at release time to verify e2e functionality. Without the build/CI infrastructure to run these tests in the emulator mode, they will rarely get invoked and therefore are not all that useful. We need to also update the CI infrastructure to run these tests regularly (like how you set up the Python build pipeline for RTDB+emulator).

test/integration/setup.ts
```diff
@@ -36,61 +36,94 @@ export let noServiceAccountApp: admin.app.App;
...
+  let serviceAccountId: string;
 
-  try {
-    apiKey = fs.readFileSync(path.join(__dirname, '../resources/apikey.txt')).toString().trim();
-  } catch (error) {
-    console.log(chalk.red(
-      'The integration test suite requires an API key for a ' +
-      'Firebase project to be saved to `test/resources/apikey.txt`.',
-      error,
+  /* tslint:disable:no-console */
```
hiranya911, line 52: Same here. We probably need a separate emulator-only test suite and the corresponding changes in CI to run it regularly.

test/unit/auth/auth.spec.ts
```diff
@@ -3214,41 +3214,6 @@ AUTH_CONFIGS.forEach((testConfig) => {
         // Make sure this doesn't throw
         jwt.verify(token, '', { algorithms: ['none'] });
       });
```
hiranya911, line 3216: Can we run a unit test where we force the emulator mode by setting the env variable to a fake value, and then see if we get an error when the SDK unsuccessfully attempts to connect to it?

test/integration/auth.spec.ts
```diff
@@ -433,9 +441,14 @@ describe('admin.auth', () => {
         ), 1000));
       })
       .then(() => {
-        // verifyIdToken without checking revocation should still succeed.
-        return admin.auth().verifyIdToken(currentIdToken)
-          .should.eventually.be.fulfilled;
+        const verifyingIdToken = admin.auth().verifyIdToken(currentIdToken)
+        if (!authEmulatorHost) {
```
hiranya911, line 445: Nit: Check for true condition first.

test/integration/auth.spec.ts
```diff
@@ -2132,12 +2232,11 @@ function safeDelete(uid: string): Promise<void> {
...
  * @return {Promise} A promise that resolves when delete operation resolves.
  */
-function deleteUsersWithDelay(uids: string[]): Promise<admin.auth.DeleteUsersResult> {
-  return new Promise((resolve) => {
-    setTimeout(resolve, 1000);
-  }).then(() => {
-    return admin.auth().deleteUsers(uids);
-  });
+async function deleteUsersWithDelay(uids: string[]): Promise<admin.auth.DeleteUsersResult> {
+  if (!authEmulatorHost) {
```
hiranya911, line 2236: Nit: true condition first

test/integration/auth.spec.ts
```diff
@@ -137,7 +142,10 @@ describe('admin.auth', () => {
       });
   });
 
-  it('createUser() creates a new user with enrolled second factors', () => {
+  it('createUser() creates a new user with enrolled second factors', function () {
+    if (authEmulatorHost) {
```
hiranya911, line 146: Too many conditional skips creeping into this test suite. What are the chances the emulator will support these features in the future so we can get rid of these?

hiranya911 (changes requested):
Thanks @yuchenshi. The implementation looks pretty good. However, we need more work to round off testing. My suggestion is to update the CI setup to invoke the auth integ tests against the emulator as part of our regular builds.

test/integration/auth.spec.ts
yuchenshi: (discussed internally about unreleased features)

test/integration/auth.spec.ts
yuchenshi: As discussed, I'll make a follow up PR to enable testing against Emulator Suite since there are some unrelated tests that will only pass after the next Firebase CLI release. As an upside, we will have coverage for tests not just for Auth, but also all of Firestore ones and most RTDB ones.

In the meantime, I ran these locally with the unreleased CLI patches and they pass.

test/integration/setup.ts
yuchenshi: See comments above.

src/auth/auth.ts
yuchenshi: Done.

src/auth/token-verifier.ts
yuchenshi: If we keep this as part of state, it won't correctly respond if `process.env.FIREBASE_AUTH_EMULATOR_HOST` is changed between two calls. This is something we do for unit tests (including the new one added according to your suggestion) and I think developers will probably do this too and get confused. I'd say just keep it since this function just delegates to reading `process.env` and isn't expensive at all.

test/unit/auth/auth.spec.ts
yuchenshi: Added.

test/integration/auth.spec.ts
yuchenshi: Done.

test/integration/auth.spec.ts
yuchenshi: I don't see how to flip this since the delay is only needed if NOT running against emulator.

test/unit/auth/auth.spec.ts
yuchenshi: Okay, in spirit of a unit test, I did it with a mock since I cannot find an address that fails reliably on all platforms (even the unrouteable `0.0.0.0` is treated as loopback on Linux).

src/auth/token-verifier.ts
```diff
@@ -209,7 +202,7 @@ export class FirebaseTokenVerifier {
         verifyJwtTokenDocsMessage;
     } else if (payload.iss !== this.issuer + projectId) {
       errorMessage = `${this.tokenInfo.jwtName} has incorrect "iss" (issuer) claim. Expected ` +
-        `"${this.issuer}"` + projectId + '" but got "' +
+        `"${this.issuer}` + projectId + '" but got "' +
```
samtstern, line 205: nit: took me a while staring at this diff because of the mixed use of string templates and string concatenation via `+`.   No need to do anything though, probably not in scope.

samtstern (approved):

test/integration/auth.spec.ts
hiranya911: Ah, ignore my comment. I misread the code.

src/auth/token-verifier.ts
hiranya911: Let's at least make sure that a single invocation of the method uses the same value. If the env variable changes while the method is being executed, different parts of the method will see different values (unlikely scenario, but nevertheless we should implement strong semantics). Call `useEmulator()` once and assign the result to a local `const`, and use it everywhere in the method body. Pass the value to any other private helper methods as needed.

hiranya911 (approved):
LGTM with a suggestion. Also give @bojeil-google a chance to review before merging.

src/auth/token-verifier.ts
yuchenshi: Done.

test/integration/setup.ts
```diff
@@ -36,61 +36,94 @@ export let noServiceAccountApp: admin.app.App;
 
 export let cmdArgs: any;
 
+export const isEmulator = !!process.env.FIREBASE_EMULATOR_HUB;
+export function skipForEmulator(): void {
```
bojeil-google, line 40: Where is this function used?

test/unit/auth/auth.spec.ts
```diff
@@ -3215,39 +3217,19 @@ AUTH_CONFIGS.forEach((testConfig) => {
...
 
-      it('verifyIdToken() rejects an unsigned token when only the env var is set', async () => {
+      it('verifyIdToken() rejects an unsigned token if auth emulator is unreachable', async () => {
```
bojeil-google, line 3220: can you add a test for `verifyIdToken()` and `verifySessionCookie()` with the emulator? I don't see any unit tests for that to confirm that `checkRevoked` is set to true.

bojeil-google (changes requested):

test/integration/setup.ts
yuchenshi: Good catch, removed.

test/unit/auth/auth.spec.ts
yuchenshi: Added those tests!

test/unit/auth/auth.spec.ts
```diff
@@ -3217,6 +3223,65 @@ AUTH_CONFIGS.forEach((testConfig) => {
...
+        const unsignedToken = mocks.generateIdToken({
+          algorithm: 'none',
+          subject: uid,
+          issuer: 'https://session.firebase.google.com/' + mocks.projectId,
+        }, {
+          iat: oneSecBeforeValidSince,
+          auth_time: oneSecBeforeValidSince, // eslint-disable-line @typescript-eslint/camelcase
+        });
+
+        // verifyIdToken should force checking revocation in emulator mode,
```
bojeil-google, line 3272: verifySessionCookie

bojeil-google (approved):

test/unit/auth/auth.spec.ts
yuchenshi: Done.

test/integration/setup.ts
yuchenshi: https://github.com/firebase/firebase-admin-node/pull/1155

src/auth/auth.ts
```diff
@@ -115,15 +115,16 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
...
    */
   public verifyIdToken(idToken: string, checkRevoked = false): Promise<DecodedIdToken> {
-    return this.idTokenVerifier.verifyJWT(idToken)
+    const isEmulator = useEmulator();
+    return this.idTokenVerifier.verifyJWT(idToken, isEmulator)
       .then((decodedIdToken: DecodedIdToken) => {
         // Whether to check if the token was revoked.
-        if (!checkRevoked) {
-          return decodedIdToken;
+        if (checkRevoked || isEmulator) {
```
lprhodes, line 122: @yuchenshi can you give any context around why the `verifyIdToken` method always calls `verifyDecodedJWTNotRevokedOrDisabled` if using emulator please?

I'm in the position where I'm working with uid's that may/may not exist in Firebase.

I want to verify a token (which succeeds in prod) but the emulator checks whether there's a user and whether its been disabled regardless of the value of  `checkRevoked` .

If this isn't intended I can create a PR.

TIA

yuchenshi: @lprhodes Tokens from the Auth Emulator aren't really "verifiable" since they are not signed (compared to production), so the closest approximation we can do is to ask the Auth Emulator if they exist. It sounds like you have a specific workflow in mind that you want to emulate and please open a new issue with more context (especially why you'd like to verify tokens from non-existent users) and we'll see what we can do.

PR #1151, opened 2021-02-01
feat(firestore): Exposed Firestore APIs from firebase-admin/firestore…

hiranya911:
* Defined the `firebase-admin/firestore` module entry point.
* Exposed the public APIs of `@google-cloud/firestore` from the new entry point.


lahirumaramba (approved):
LGTM!

PR #1155, opened 2021-02-04
Add emulator-based integration tests.

yuchenshi:

### Discussion

This is a follow up to #1148 and enables emulator-based integration tests.

### Testing

This covers Firestore and most of Auth and database.

### API Changes

N/A


.github/workflows/ci.yml
```diff
@@ -17,10 +17,16 @@ jobs:
...
         npm run build
         npm run build:tests
+    - name: Lint and run unit tests
+      run: |
         npm test
         npm run api-extractor
+    - name: Run emulator-based integration tests
+      run: |
+        npm install -g firebase-tools
+        npm run integration:emulator
```
hiranya911, line 32: Should probably call `npm run test:integration:emulator` here. Otherwise code will be built again, which is not required (it was already built earlier).

package.json
```diff
@@ -15,8 +15,10 @@
     "lint": "run-p lint:src lint:test",
     "test": "run-s lint test:unit",
     "integration": "run-s build test:integration",
+    "integration:emulator": "run-s build test:integration:emulator",
     "test:unit": "mocha test/unit/*.spec.ts --require ts-node/register",
     "test:integration": "mocha test/integration/*.ts --slow 5000 --timeout 20000 --require ts-node/register",
+    "test:integration:emulator": "firebase emulators:exec --project fake-project-id --only auth,database,firestore 'mocha \"test/integration/{auth,database,firestore}.spec.ts\" --slow 5000 --timeout 20000 --require ts-node/register'",
```
hiranya911, line 21: Let's do this in the workflow configuration. Ideally we shouldn't add any emulator specific commands to this file.

.github/workflows/ci.yml
```diff
@@ -17,10 +17,16 @@ jobs:
...
+    - name: Install and build
       run: |
         npm ci
         npm run build
         npm run build:tests
+    - name: Lint and run unit tests
+      run: |
         npm test
         npm run api-extractor
+    - name: Run emulator-based integration tests
```
hiranya911, line 29: Might as well make API Extractor a separate step too.

hiranya911 (changes requested):
Thanks @yuchenshi. Really like how this is turning out. My main point of feedback is about managing the emulator-based integration tests entirely within the Actions workflow, rather than in `package.json`. That's consistent with how we did this in [Python](https://github.com/firebase/firebase-admin-python/blob/b35abb9bb74ab7629ae1ff5c6f73772c6dd450c5/.github/workflows/ci.yml#L29-L32), and keeps the package.json file clean of changes.

package.json
yuchenshi: Done.

.github/workflows/ci.yml
yuchenshi: Done.

yuchenshi:
Sorry I was OOO and I've now reverted `package.json` changes as requested.

.github/workflows/ci.yml
```diff
@@ -17,10 +17,19 @@ jobs:
...
       with:
         node-version: ${{ matrix.node-version }}
-    - name: Install, build and test
+    - name: Install and build
       run: |
         npm ci
         npm run build
         npm run build:tests
+    - name: Lint and run unit tests
+      run: |
```
hiranya911, line 26: Only one command here. Can be inlined:

```
run: npm test
```

.github/workflows/ci.yml
```diff
@@ -17,10 +17,19 @@ jobs:
...
         npm test
+    - name: Run api-extractor
+      run: |
```
hiranya911, line 29: Here too.

hiranya911 (approved):
Thanks @yuchenshi. LGTM with a couple of nits. Feel free to merge once the comments are addressed.

yuchenshi:
I see one flaky tests here and I'll investigate later. Don't merge for now please! https://github.com/firebase/firebase-admin-node/pull/1155/checks?check_run_id=1881208101#step:7:197

It's unclear if it's flaky due to some issue in Auth Emulator or maybe the test itself has race conditions that got worse since Auth Emulator is much faster than prod.

yuchenshi:
Merging now since the tests now pass. Feel free to skip that one test for emulators if this comes up again. Let's at least have coverage for the other tests.

lahirumaramba:
Hey @yuchenshi !
I am getting that flaky test again. I believe it is a race condition as it passes some times with a re-run.
Any thoughts on how we could handle this? Thanks!
https://github.com/firebase/firebase-admin-node/pull/1204/checks?check_run_id=2189478758

hiranya911:
If there's no obvious forward fix, we might want to revert this PR for the moment.

yuchenshi:
I've opened https://github.com/firebase/firebase-admin-node/pull/1205 to disable that flaky test only for emulators.

yuchenshi:
Found the root cause in Auth Emulator, tracked on https://github.com/firebase/firebase-tools/issues/3231. Will add the test back after fixing and another release. We're unblocked for now.

PR #1156, opened 2021-02-04
feat: Exposed Rules APIs from firebase-admin/security-rules entry point

hiranya911:
* Exported `firebase-admin/security-rules` module entry point.
* Added API report.


lahirumaramba (approved):
LGTM!

PR #1159, opened 2021-02-08
feat(storage): Exposed firebase-admin/storage entry point

hiranya911:
* Added the `firebase-admin/storage` module entry point.
* Added API report.


lahirumaramba (approved):
LGTM!

PR #1160, opened 2021-02-09
feat(ml): Added firebase-admin/machine-learning entry point

hiranya911:
* Added `firebase-admin/machine-learning` module entry point.
* Added API report for the new module.
* Updated the gulpfile to include all auto-generated d.ts files in the build (We can decide later if we want to filter some of them out in the future. But for now there's no reason to not include them).

This PR largely completes the refactoring work in individual service modules. In the next PRs I'll work on cleaning up the API documentation, and clearly marking any internal APIs that need to be screened out from the build.


lahirumaramba (approved):
LGTM!

PR #1164, opened 2021-02-09
fix: Moved FirebaseError interface to app module

hiranya911:
* Moved the definitions of `FirebaseError` and FirebaseArrayIndexError` to the `firebase-admin/app` module where they are reusable across all other modules.
* The old firebase namespace API auto exports everything in `firebase-admin/app`, so no changes there.
* Updated some tests to import the public API types from the new module entry points.


lahirumaramba (approved):
LGTM!

PR #1168, opened 2021-02-11
fix: Updated documentation for app and instance-id entry points

hiranya911:
Added documentation for `firebase-admin/app` and `firebase-admin/instance-id` module entry points. For the most part I'm copying the existing documentation from our main branch. Note that the API report indicates the docs status of each API.

### DIFF
--- a/src/app/core.ts
+++ b/src/app/core.ts
@@ -84,6 +84,16 @@ export interface AppOptions {
   httpAgent?: Agent;
 }
 
+/**
+ * A Firebase app holds the initialization information for a collection of
+ * services.
+ *
+ * Do not call this constructor directly. Instead, use
+ * {@link
+ *   https://firebase.google.com/docs/reference/admin/node/admin#.initializeApp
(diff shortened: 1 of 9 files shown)


src/app/core.ts
```diff
@@ -84,6 +84,16 @@ export interface AppOptions {
...
 }
 
+/**
+ * A Firebase app holds the initialization information for a collection of
+ * services.
+ *
+ * Do not call this constructor directly. Instead, use
+ * {@link
+ *   https://firebase.google.com/docs/reference/admin/node/admin#.initializeApp
+ *   `admin.initializeApp()`}
```
lahirumaramba, line 94: Should this be just `initializeApp()`, and below?

lahirumaramba (approved):
LGTM!
Left one comment.

src/app/core.ts
hiranya911: This API is shared between both old and modularized APIs. So we can go either way here. For now I'm going to keep the old documentation here.

PR #1170, opened 2021-02-16
fix(auth): Updated API docs for firebase-admin/auth

hiranya911:
* Updated documentation for public APIs. Documentation text copied from the current main branch.
* Changed `BaseAuth` into an abstract class.

### DIFF
--- a/src/auth/user-record.ts
+++ b/src/auth/user-record.ts
@@ -86,12 +86,28 @@ enum MultiFactorId {
 }
 
 /**
- * Abstract class representing a multi-factor info interface.
+ * Interface representing the common properties of a user enrolled second factor.
  */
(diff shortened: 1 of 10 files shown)


src/auth/user-record.ts
```diff
@@ -339,28 +409,101 @@ export class UserInfo {
...
+  /**
+   * The user's `uid`.
+   */
   public readonly uid: string;
-  public readonly email: string;
+
+  /**
+   * The user's primary email, if set.
+   */
+  public readonly email?: string;
```
lahirumaramba, line 424: Were these not optional in internal types by mistake?

lahirumaramba (approved):
LGTM!

PR #1173, opened 2021-02-19
fix: Updated integration tests to use the modular API surface

hiranya911:
I've updated the existing integration test suite to use the new modular API surface whenever possible. Most of the changes required were pretty straightforward. For example:

1. Changing invocations like `admin.auth()` into `getAuth()`.
2. Changing type annotations like `admin.auth.UserRecord` into `UserRecord`.

However, this change broke some of the existing lazy loading tests. The updated `database.spec.ts` and `firestore.spec.ts` imports `firebase-admin/database` and `firebase-admin/firestore`, which in turn loads the corresponding 3rd party dependencies. The old lazy loading tests relied on this not happening. I've implemented a somewhat hacky solution to get around this issue. But we should probably find a better way to test lazy loading in the future.

### DIFF
--- a/test/integration/app.spec.ts
+++ b/test/integration/app.spec.ts
@@ -30,7 +30,7 @@ describe('admin', () => {
   it('does not load RTDB by default', () => {
     const firebaseRtdb = require.cache[require.resolve('@firebase/database')];
     expect(firebaseRtdb).to.be.undefined;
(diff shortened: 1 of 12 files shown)


lahirumaramba (approved):
LGTM!
I don't have a better suggestion for the lazy loading tests atm... I think we can revisit it in a future PR.

PR #1175, opened 2021-02-22
Add support for FIREBASE_STORAGE_EMULATOR_HOST env var

abeisgoat:
Adds support for mapping the firebase-style `FIREBASE_STORAGE_EMULATOR_HOST` env var to the Cloud style `STORAGE_EMULATOR_HOST`


hiranya911 (changes requested):
Can we add a test case?

abeisgoat:
What would you imagine this test would check for?

hiranya911:
Easiest would be to programmatically set the Firebase env variable, call `admin.storage()` (or equivalent API) and check if that sets the Cloud env variable under the hood.

test/unit/storage/storage.spec.ts
```diff
@@ -29,88 +29,110 @@ describe('Storage', () => {
...
       });
     });
   });
 
-  describe('bucket()', () => {
-    it('should return a bucket object', () => {
-      expect(storage.bucket().name).to.equal('bucketName.appspot.com');
-    });
-  });
+  describe('Emulator Support', () => {
```
hiranya911, line 119: This is the only bit I expected. Why did we change other tests in this file?

test/unit/storage/storage.spec.ts
```diff
@@ -29,88 +29,110 @@ describe('Storage', () => {
...
-    it('should return a bucket object', () => {
-      expect(storage.bucket('foo').name).to.equal('foo');
+      expect(process.env.STORAGE_EMULATOR_HOST).to.equal(EMULATOR_HOST);
     });
-  });
+  
+    afterEach(() => {
+      delete process.env.STORAGE_EMULATOR_HOST;
+      delete process.env.FIREBASE_STORAGE_EMULATOR_HOST;
+      return mockApp.delete();
```
hiranya911, line 135: There's already a hook that deletes mockApp. This is not needed.

test/unit/storage/storage.spec.ts
```diff
@@ -29,88 +29,110 @@ describe('Storage', () => {
...
-    });
-  });
+  describe('Emulator Support', () => {
+    it('sets STORAGE_EMULATOR_HOST if FIREBASE_STORAGE_EMULATOR_HOST is set', () => {
+      const EMULATOR_HOST = 'http://localhost:9199';
+      delete process.env.STORAGE_EMULATOR_HOST;
+      process.env.FIREBASE_STORAGE_EMULATOR_HOST = EMULATOR_HOST;
+
+      mockApp = mocks.app();
+      mockCredentialApp = mocks.mockCredentialApp();
```
hiranya911, line 126: Not used

test/unit/storage/storage.spec.ts
```diff
@@ -29,88 +29,110 @@ describe('Storage', () => {
...
-      expect(storage.bucket().name).to.equal('bucketName.appspot.com');
-    });
-  });
+  describe('Emulator Support', () => {
+    it('sets STORAGE_EMULATOR_HOST if FIREBASE_STORAGE_EMULATOR_HOST is set', () => {
+      const EMULATOR_HOST = 'http://localhost:9199';
+      delete process.env.STORAGE_EMULATOR_HOST;
+      process.env.FIREBASE_STORAGE_EMULATOR_HOST = EMULATOR_HOST;
+
+      mockApp = mocks.app();
```
hiranya911, line 125: Already initialized in a `beforeEach` hook above.

hiranya911 (changes requested):
Test case looks good. Just needs a bit of clean up. Lets also keep the diff small by not changing anything that's no related to this feature.

test/unit/storage/storage.spec.ts
ssbushi: The `beforeEach` hook is now only for the non-emulator tests. I had to split it into 2 because I need to set the env variable before the `beforeEach` hook starts.

test/unit/storage/storage.spec.ts
ssbushi: Deleted.

test/unit/storage/storage.spec.ts
ssbushi: Done.

test/unit/storage/storage.spec.ts
ssbushi: I split the emulator vs non-emulator tests into 2 different `describe` blocks. Git diff is showing the indentations as additions/deletions. These changes are no-op.

hiranya911 (changes requested):

test/unit/storage/storage.spec.ts
ssbushi: Done.

test/unit/storage/storage.spec.ts
ssbushi: Done.

test/unit/storage/storage.spec.ts
```diff
@@ -113,4 +113,22 @@ describe('Storage', () => {
...
+  describe('Emulator mode', () => {
+    const EMULATOR_HOST = 'http://localhost:9199';
+
+    before(() => {
+      delete process.env.STORAGE_EMULATOR_HOST;
+      process.env.FIREBASE_STORAGE_EMULATOR_HOST = EMULATOR_HOST;
+    });
+
+    it('sets STORAGE_EMULATOR_HOST if FIREBASE_STORAGE_EMULATOR_HOST is set', () => {
+      expect(process.env.STORAGE_EMULATOR_HOST).to.equal(EMULATOR_HOST);
```
hiranya911, line 126: I assume this relies on the `beforeEach()` block above calling `new Storage()`. I'd recommend making that call explicit here, so the test case is more self-contained.

```
new Storage(mockApp);
expect(process.env.STORAGE_EMULATOR_HOST).to.equal(EMULATOR_HOST);
```

hiranya911 (approved):
LGTM with a suggestion. Feel free to merge when ready.

test/unit/storage/storage.spec.ts
ssbushi: Done.

PR #1176, opened 2021-02-22
chore: Updated package.json file for entry points

hiranya911:
I'm adding an `exports` section to our top-level package.json file to declare all the different module entry points we expose in the SDK. However, `exports` is only supported in Node.js 12+. Moreover TypeScript doesn't support it in any version (see https://github.com/microsoft/TypeScript/issues/33079). 

~~In order to work around these limitations I'm also adding a set of separate `package.json` files for each of our entry points. This leads to some config duplication, but hopefully TypeScript will support `exports` in the future, and then we can get rid of these additional files.~~

In order to work around these limitations, I'm bumping our minimum platform requirement to Node.js 12. And, I'm adding a `typeVersions` configuration to our `package.json` so that typing resolution works correctly for the new entry points.

I'm also updating the integration test suite to explicitly verify each of the module entry points of the SDK.


hiranya911:
@lahirumaramba @Feiyang1 any thoughts on how we can reduce the duplication?

Or is there a way to not have those additional directories at the root of the repo, but still be able to package them into the release tarball?

lahirumaramba (approved):
Thanks, Hiranya!

I am not against the current solution as it seems we have to pick one workaround over another until typescript officially supports `exports`.

The workaround/solution mentioned in,
https://github.com/teppeis/typescript-subpath-exports-workaround
looks more attractive as it removes the need for having the top level directories and additional `package.json` files.

Since we have to add the top-level directories to the top-level `package.json` anyway, it won't be too bad to add a `typesVersions` field pointing to typings in `lib` for each module... all this is assuming the latter works with our module structure. :)

hiranya911:
> all this is assuming the latter works with our module structure. 

It does work, and really starting to look like a viable option for us. To summarize we are looking to work around 2 problems:

1. Node.js 10 support (`exports` field in `package.json` is only supported Node 12 and up).
2. TypeScript support (https://github.com/microsoft/TypeScript/issues/33079)

The solution of top-level directories with additional `package.json` files solve both these problems. But it results in a quite messy directory structure, and lot of duplicate configuration.

However, we can solve problem 1 by just bumping our minimum platform requirements to Node 12 (Node 10 is going EoL in a couple of months anyway). Then hopefully we can use the `typeVersions` solution to work around problem 2, and not have to introduce a bunch of new directories and `package.json` files.

@Feiyang1 your thoughts?

Feiyang1:
Does customer need to npm install `typescript-subpath-exports-workaround`, which the readme seems to suggest? I'm slightly confused because `typeVersions` is natively supported by Typescript, so it should just work out of the box.

I think it will be a problem if `typescript-subpath-exports-workaround` needs to be installed by customers, otherwise it looks like a great solution.

hiranya911:
`typescript-subpath-exports-workaround` is just an example that demonstrates how to use the `typeVersions` workaround. It doesn't need to be installed. We will just be making changes in our own package.json file.

hiranya911:
I've bumped our platform requirement to Node 12, and implemented the `typeVersions` workaround. Note that the typescript integration tests pass with the release artifact created this way.

@lahirumaramba @Feiyang1 over to you for another look.

lahirumaramba (approved):
Thanks, Hiranya! LGTM!

package.json
```diff
@@ -6,7 +6,7 @@
   "license": "Apache-2.0",
   "homepage": "https://firebase.google.com/",
   "engines": {
-    "node": ">=10.10.0"
+    "node": ">=12.7.0"
```
Feiyang1, line 9: Does Cloud Function still support Node 10? If so, will this prevents people from using `admin-node` with Node 10 in Cloud Function?

Feiyang1: LGTM, otherwise

hiranya911: Their early reaction was that this change is ok. Functions are already defaulting to Node 12 for all new deployments. We will of course bump our major version to indicate this change when we eventually release this.

PR #1189, opened 2021-03-08
chore: Move token verification logic to util

lahirumaramba:
- Move `FirebaseTokenVerifier` to util
- Include error type and codes to `FirebaseTokenInfo` to throw public error types from `FirebaseTokenVerifier`

### DIFF
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -29,9 +29,8 @@ import { AuthClientErrorCode, FirebaseAuthError, ErrorInfo } from '../utils/erro
 import * as utils from '../utils/index';
 import * as validator from '../utils/validator';
 import { auth } from './index';
-import {
-  FirebaseTokenVerifier, createSessionCookieVerifier, createIdTokenVerifier
-} from '../utils/token-verifier';
+import { FirebaseTokenVerifier } from '../utils/token-verifier';
+import { createSessionCookieVerifier, createIdTokenVerifier } from './token-verifier-util';
 import {
   SAMLConfig, OIDCConfig, OIDCConfigServerResponse, SAMLConfigServerResponse,
 } from './auth-config';
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -116,8 +116,9 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
    */
   public verifyIdToken(idToken: string, checkRevoked = false): Promise<DecodedIdToken> {
     const isEmulator = useEmulator();
-    return this.idTokenVerifier.verifyJWT(idToken, isEmulator)
+    return this.idTokenVerifier.verifyJWT<DecodedIdToken>(idToken, isEmulator)
       .then((decodedIdToken: DecodedIdToken) => {
+        decodedIdToken.uid = decodedIdToken.sub;
         // Whether to check if the token was revoked.
         if (checkRevoked || isEmulator) {
           return this.verifyDecodedJWTNotRevoked(
@@ -519,8 +520,9 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
   public verifySessionCookie(
     sessionCookie: string, checkRevoked = false): Promise<DecodedIdToken> {
     const isEmulator = useEmulator();
-    return this.sessionCookieVerifier.verifyJWT(sessionCookie, isEmulator)
+    return this.sessionCookieVerifier.verifyJWT<DecodedIdToken>(sessionCookie, isEmulator)
       .then((decodedIdToken: DecodedIdToken) => {
+        decodedIdToken.uid = decodedIdToken.sub;
         // Whether to check if the token was revoked.
         if (checkRevoked || isEmulator) {
           return this.verifyDecodedJWTNotRevoked(
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -30,7 +30,7 @@ import * as utils from '../utils/index';
 import * as validator from '../utils/validator';
 import { auth } from './index';
 import { FirebaseTokenVerifier } from '../utils/token-verifier';
-import { createSessionCookieVerifier, createIdTokenVerifier } from './token-verifier-util';
+import { createSessionCookieVerifier, createIdTokenVerifier } from './token-verifier';
 import {
   SAMLConfig, OIDCConfig, OIDCConfigServerResponse, SAMLConfigServerResponse,
 } from './auth-config';
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -31,7 +31,7 @@ import * as validator from '../utils/validator';
 import { auth } from './index';
 import {
   FirebaseTokenVerifier, createSessionCookieVerifier, createIdTokenVerifier
-} from './token-verifier';
+} from '../utils/token-verifier';
 import {
   SAMLConfig, OIDCConfig, OIDCConfigServerResponse, SAMLConfigServerResponse,
 } from './auth-config';
--- a/src/auth/token-verifier-util.ts
+++ b/src/auth/token-verifier-util.ts
@@ -1,0 +1,89 @@
+/*!
+ * Copyright 2021 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { FirebaseApp } from '../firebase-app';
+import { AuthClientErrorCode, ErrorCodeConfig, FirebaseAuthError } from '../utils/error';
+import { FirebaseTokenInfo, FirebaseTokenVerifier } from '../utils/token-verifier';
+
+const ALGORITHM_RS256 = 'RS256';
+
+// URL containing the public keys for the Google certs (whose private keys are used to sign Firebase
+// Auth ID tokens)
+const CLIENT_CERT_URL = 'https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com';
+
+// URL containing the public keys for Firebase session cookies. This will be updated to a different URL soon.
+const SESSION_COOKIE_CERT_URL = 'https://www.googleapis.com/identitytoolkit/v3/relyingparty/publicKeys';
+
+/** Error codes that matches the FirebaseAuthError type */
+const AUTH_ERROR_CODE_CONFIG: ErrorCodeConfig = {
+  invalidArg: AuthClientErrorCode.INVALID_ARGUMENT,
+  invalidCredential: AuthClientErrorCode.INVALID_CREDENTIAL,
+  internalError: AuthClientErrorCode.INTERNAL_ERROR,
+}
+
+/** User facing token information related to the Firebase ID token. */
+export const ID_TOKEN_INFO: FirebaseTokenInfo = {
+  url: 'https://firebase.google.com/docs/auth/admin/verify-id-tokens',
+  verifyApiName: 'verifyIdToken()',
+  jwtName: 'Firebase ID token',
+  shortName: 'ID token',
+  expiredErrorCode: AuthClientErrorCode.ID_TOKEN_EXPIRED,
+  errorCodeConfig: AUTH_ERROR_CODE_CONFIG,
+  errorType: FirebaseAuthError,
+};
+
+/** User facing token information related to the Firebase session cookie. */
+export const SESSION_COOKIE_INFO: FirebaseTokenInfo = {
+  url: 'https://firebase.google.com/docs/auth/admin/manage-cookies',
+  verifyApiName: 'verifySessionCookie()',
+  jwtName: 'Firebase session cookie',
+  shortName: 'session cookie',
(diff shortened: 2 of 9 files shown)


src/utils/token-verifier.ts
```diff
@@ -43,6 +44,8 @@ export const ID_TOKEN_INFO: FirebaseTokenInfo = {
   jwtName: 'Firebase ID token',
   shortName: 'ID token',
   expiredErrorCode: AuthClientErrorCode.ID_TOKEN_EXPIRED,
+  errorCodeType: AuthClientErrorCode,
```
hiranya911, line 47: ID_TOKEN_INFO and SESSION_COOKIE_INFO should be defined in the auth module.

src/utils/error.ts
```diff
@@ -338,10 +338,28 @@ export class AppErrorCodes {
   public static UNABLE_TO_PARSE_RESPONSE = 'unable-to-parse-response';
 }
 
+/**
+ * Base class for client error codes and their default messages.
+ */
+export class BaseClientErrorCode {
```
hiranya911, line 344: This is a fairly weak abstraction. It's a class with no specific shape, but a few static members. I think this ought to be an interface:

```typescript
interface ErrorCodeConfig {
  invalidArg: string;
  invalidCredential: string;
  internalError: string;
}
```

WDYT?

hiranya911 (changes requested):
Haven't looked at everything, but looks reasonable. Let's start off with some feedback on the high-level design.

src/utils/token-verifier.ts
lahirumaramba: Good call. Moved all auth related items to a new `token-verifier-util.ts`

src/utils/error.ts
lahirumaramba: Agreed! Given the number of error codes needed in the token verifier util this makes sense. Thanks!

src/auth/token-verifier-util.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/** User facing token information related to the Firebase ID token. */
+export const ID_TOKEN_INFO: FirebaseTokenInfo = {
+  url: 'https://firebase.google.com/docs/auth/admin/verify-id-tokens',
+  verifyApiName: 'verifyIdToken()',
+  jwtName: 'Firebase ID token',
+  shortName: 'ID token',
+  expiredErrorCode: AuthClientErrorCode.ID_TOKEN_EXPIRED,
+  errorCodeConfig: AUTH_ERROR_CODE_CONFIG,
```
lahirumaramba, line 44: I am debating if we should include `expiredErrorCode` in `errorCodeConfig` or not. If we do so `ErrorCodeConfig` interface will look more of a token related config. WDYT?

src/auth/token-verifier-util.ts
```diff
@@ -0,0 +1,89 @@
+/*!
```
hiranya911, line 1: These can go in the token-verifier module like they used to. No need to introduce a new module.

src/auth/token-verifier-util.ts
hiranya911: I expected it to be a token verification specific thing. I don't think it has any use outside of that context. +1 for merging the expiredErrorCode into the same interface.

test/unit/auth/token-verifier.spec.ts
```diff
@@ -45,6 +46,12 @@ const expect = chai.expect;
 const ONE_HOUR_IN_SECONDS = 60 * 60;
 const idTokenPublicCertPath = '/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com';
 
+const AUTH_ERROR_CODE_CONFIG: ErrorCodeConfig = {
```
hiranya911, line 49: Why not reuse the const defined in the implementation?

test/unit/auth/token-verifier.spec.ts
```diff
@@ -309,6 +326,34 @@ describe('FirebaseTokenVerifier', () => {
       }).to.throw('First argument to verifyIdToken() must be a Firebase ID token');
     });
 
+    const errorTypes = [
```
hiranya911, line 329: Why use an array with just one element?

test/unit/auth/token-verifier.spec.ts
```diff
@@ -309,6 +326,34 @@ describe('FirebaseTokenVerifier', () => {
...
+          expiredErrorCode: errorType.config.invalidArg,
+          errorCodeConfig: errorType.config,
+          errorType: errorType.type,
+        },
+        app,
+      );
+      it('should throw with the correct error type and code set in token info', () => {
+        expect(() => {
+          (tokenVerifier as any).verifyJWT();
+        }).to.throw(errorType.type).with.property('code').match(
```
hiranya911, line 351: ```equal(`auth/${errorType.config.invalidArg.code}`)```

hiranya911 (changes requested):
Looks pretty good. Just a few suggestions for cleaning up the PR.

src/auth/token-verifier-util.ts
lahirumaramba: Some of the components `ID_TOKEN_INFO` and `SESSION_COOKIE_INFO` are `auth` specific. Instead of moving them to the `auth` module, I introduced the token verifier util module for better separation. Would you rather have them all in the `auth` module or back in the new `token-verifier` util module?

test/unit/auth/token-verifier.spec.ts
lahirumaramba: I was hoping that we will have other error types to include here in the future. Will remove the array for now and we can add it later when we have more if needed.

test/unit/auth/token-verifier.spec.ts
lahirumaramba: Updated!

test/unit/auth/token-verifier.spec.ts
lahirumaramba: Updated to reuse the ones from the implementation. Thanks!

src/auth/token-verifier-util.ts
lahirumaramba: Renamed back to `auth/token-verifier.ts`. I agree that it makes more sense now. Thanks!

lahirumaramba:
- Merged the expired error code to error code config
- Added validation for the new error code config and error type
- Added unit tests for new validations

src/utils/token-verifier.ts
```diff
@@ -213,9 +216,9 @@ export class FirebaseTokenVerifier {
           if (error) {
             if (error.name === 'TokenExpiredError') {
               const errorMessage = `${this.tokenInfo.jwtName} has expired. Get a fresh ${this.tokenInfo.shortName}` +
-                ` from your client app and try again (auth/${this.tokenInfo.expiredErrorCode.code}).` +
+                ` from your client app and try again (auth/${this.tokenInfo.errorCodeConfig.expiredError.code}).` +
```
hiranya911, line 219: There's a reference to `auth` here. Should probably get rid of it.

test/unit/auth/token-verifier.spec.ts
```diff
@@ -30,10 +30,10 @@ import LegacyFirebaseTokenGenerator = require('firebase-token-generator');
 import * as mocks from '../../resources/mocks';
 import { FirebaseTokenGenerator, ServiceAccountSigner } from '../../../src/auth/token-generator';
 import * as verifier from '../../../src/utils/token-verifier';
-import * as verifierUtil from '../../../src/auth/token-verifier-util';
+import * as verifierAuth from '../../../src/auth/token-verifier';
```
hiranya911, line 33: Easier and less ambiguous to just import named members here.

```
import { ID_TOKEN_INFO, SESSION_COOKIE_INFO } from '...';
```

test/unit/auth/token-verifier.spec.ts
```diff
@@ -301,13 +291,34 @@ describe('FirebaseTokenVerifier', () => {
               verifyApiName: 'verifyToken()',
               jwtName: 'Important Token',
               shortName: 'token',
-              expiredErrorCode: invalidExpiredErrorCode as any,
-              errorCodeConfig: AUTH_ERROR_CODE_CONFIG,
+              errorCodeConfig: verifierAuth.ID_TOKEN_ERROR_CODE_CONFIG,
+              errorType: invalidErrorCodeTypes as any,
```
hiranya911, line 295: Did you mean to pass `invalidErrorCodeType` here?

test/unit/auth/token-verifier.spec.ts
```diff
@@ -301,13 +291,34 @@ describe('FirebaseTokenVerifier', () => {
...
+            },
+            app,
+          );
+        }).to.throw('The provided error type must be a non-null PrefixedFirebaseError type.');
+      });
+    });
+
+    const invalidErrorCodeConfigs = [null, NaN, 0, 1, true, false, [], {}, { a: 1 }, _.noop, '', 'test'];
```
hiranya911, line 303: These tests now belong in utils.

hiranya911 (changes requested):
Thanks for making the changes. Looks mostly good. Couple of points to revise:

1. There's one remaining reference to `auth` from the utils token verifier.
2. Unit tests could use a bit more attention.

src/utils/token-verifier.ts
lahirumaramba: Oh good catch! I actually ended up finding more references to Auth. Ended up using generics in the verify token functions and moved the code to set the user id alias to one level up to `auth.ts`.
Let me know what you think. Thanks!

test/unit/auth/token-verifier.spec.ts
lahirumaramba: Updated!

test/unit/auth/token-verifier.spec.ts
lahirumaramba: Thanks! Fixed.

test/unit/auth/token-verifier.spec.ts
lahirumaramba: Moved the tests to `test/utils`

src/auth/auth.ts
```diff
@@ -520,8 +520,9 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
   public verifySessionCookie(
     sessionCookie: string, checkRevoked = false): Promise<DecodedIdToken> {
     const isEmulator = useEmulator();
-    return this.sessionCookieVerifier.verifyJWT(sessionCookie, isEmulator)
+    return this.sessionCookieVerifier.verifyJWT<DecodedIdToken>(sessionCookie, isEmulator)
       .then((decodedIdToken: DecodedIdToken) => {
+        decodedIdToken.uid = decodedIdToken.sub;
```
lahirumaramba, line 525: Moved the code to set the `uid` alias here. Another way is to add helper functions, `verifySessionCookie()` and `verifyIdToken()`, to `auth/token-verifier`. Let me know your thoughts.

src/utils/token-verifier.ts
lahirumaramba: Re: the error code prefix. I just removed the `auth` prefix, but the error code feels a bit redundant here as the error code will also be included in the public error type.

src/utils/token-verifier.ts
```diff
@@ -0,0 +1,290 @@
...
+  /**
+   * Verifies the format and signature of a Firebase JWT token.
+   *
+   * @template T The returned type of the decoded token.
+   * @param {string} jwtToken The Firebase JWT token to verify.
+   * @param {boolean=} isEmulator Whether to accept Emulator tokens.
+   * @return {Promise<T>} A promise fulfilled with the decoded claims of the Firebase Auth ID
+   *                           token.
+   */
+  public verifyJWT<T extends object>(jwtToken: string, isEmulator = false): Promise<T> {
```
lahirumaramba, line 94: To tighten the type check we can introduce a base type for returned tokens...
```
interface FirebaseToken {
  // registered JWT claims
  iss: string
  aud: string
}

interface DecodedIdToken extends FirebaseToken {
...
}

public verifyJWT<T extends FirebaseToken>
```

Let me know your thoughts. Thanks!

test/unit/utils/token-verifier.spec.ts
```diff
@@ -30,10 +30,12 @@ import LegacyFirebaseTokenGenerator = require('firebase-token-generator');
 import * as mocks from '../../resources/mocks';
 import { FirebaseTokenGenerator, ServiceAccountSigner } from '../../../src/auth/token-generator';
 import * as verifier from '../../../src/utils/token-verifier';
-import * as verifierUtil from '../../../src/auth/token-verifier-util';
+import {
```
hiranya911, line 33: Sorry if I wasn't clear. But I didn't mean moving this entire file into utils. Some tests in this file are still very auth-specific, and therefore should be in auth. But the test cases that specifically test for the `utils.token-verifier`, should be moved into a new test module (like the `invalidErrorTypes` test cases).

I think a good rule of thumb is if a test requires an import from `auth`, then it probably belongs in `auth`.

hiranya911 (changes requested):
Lets find a more typesafe way to handle the returns. Plus separation of tests between auth and utils.

src/utils/token-verifier.ts
hiranya911: This is something that will get cleaned up once we fix error handling for this SDK. For now this solution is adequate.

test/unit/auth/token-verifier.spec.ts
hiranya911: Let's not move the whole file. It still contains many auth-specific tests.

src/utils/token-verifier.ts
hiranya911: Yeah, this looks good to me. But I don't think we need a generic at all. Just make auth extend the TokenVerifier class, and transform the FirebaseToken into DecodedIdToken there.

lahirumaramba:
Closing this as we decided to change the overall approach and design of the refactor. This work is replaced by #1204

PR #1190, opened 2021-03-08
feat(rtdb): Support emulator mode for rules management operations

### LINKED ISSUE #1149: Database rules apis does not automatically use emulator
###  Describe your environment

  * Operating System version:  Windows 10 Pro 20H2
  * Firebase SDK version: 9.4.2
  * Firebase Product: database
  * Node.js version: 12.15.0
  * NPM version: 6.13.4

### Describe the problem
When using the rules APIs of the database (setRules or getRules) against the emulator, the DatabaseRulesClient does not automatically change the URL to FIREBASE_DATABASE_EMULATOR_HOST, while the other APIs do. Moreover, it does not use the EmulatorAdminTokenProvider. Instead, it tries to get a real token.

### Steps to reproduce:

* setup env variables for FIREBASE_DATABASE_EMULATOR_HOST and GCLOUD_PROJECT 
* run a production-ready code that should utilize the emulator when env variables are set

#### Relevant Code:

```js
const admin = require('firebase-admin');
const rules = require('./database.rules.json');

const app = admin.initializeApp();

app.database('https://database-instance.firebasedatabase.app').setRules(rules)
    .then(res=>console.log('Rules set successfully'))
    .catch(err=>console.error(err));

```

#### Analysis

The initStandalone method from @firebase/database is aware of FIREBASE_DATABASE_EMULATOR_HOST and switches to emulator URLs and EmulatorAdminTokenProvider. However, the DatabaseRulesClient does not use getToken nor the URL from the RTDB instance initialized by initStandalone.  

I'm not too familiar with the repo, but I see that the URL and the authProvider are deep inside private fields of the RTDB instance and Repo classes. I would be happy to help to create a PR if anyone can provide me with a nice way to get the actual URL and provider from the RTDB instance.

hiranya911:
Connect to the emulator in `getRules()` and `setRules()` APIs when the emulator mode is enabled. Emulator mode can be enabled by setting the `FIREBASE_DATABASE_EMULATOR_HOST` env variable, or by setting an emulator URL directly in AppOptions. In the first case we pick up the namespace from the `databaseURL` App option.

Resolves #1149 

RELEASE NOTE: Rules management APIs now support the Firebase emulator suite. Methods like `getRules()` and `setRules()` will automatically connect to the local emulator, when the emulator mode is enabled.

### DIFF
--- a/src/database/database-internal.ts
+++ b/src/database/database-internal.ts
@@ -123,12 +123,15 @@ class DatabaseRulesClient {
   private readonly httpClient: AuthorizedHttpClient;
 
   constructor(app: FirebaseApp, dbUrl: string) {
+    let parsedUrl = new URL(dbUrl);
     const emulatorHost = process.env.FIREBASE_DATABASE_EMULATOR_HOST;
     if (emulatorHost) {
-      dbUrl = `http://${emulatorHost}`;
+      const hostname = parsedUrl.hostname;
+      const dotIndex = hostname.indexOf('.');
+      const namespace = hostname.substring(0, dotIndex).toLowerCase();
+      parsedUrl = new URL(`http://${emulatorHost}?ns=${namespace}`);
     }
 
-    const parsedUrl = new URL(dbUrl);
     parsedUrl.pathname = path.join(parsedUrl.pathname, RULES_URL_PATH);
     this.dbUrl = parsedUrl.toString();
     this.httpClient = new AuthorizedHttpClient(app);
--- a/src/database/database-internal.ts
+++ b/src/database/database-internal.ts
@@ -63,7 +63,7 @@ export class DatabaseService {
   /**
    * Returns the app associated with this DatabaseService instance.
    *
-   * @return  The app associated with this DatabaseService instance.
+   * @return The app associated with this DatabaseService instance.
    */
   get app(): FirebaseApp {
     return this.appInternal;
--- a/src/database/database-internal.ts
+++ b/src/database/database-internal.ts
@@ -63,7 +63,7 @@ export class DatabaseService {
   /**
    * Returns the app associated with this DatabaseService instance.
    *
-   * @return {FirebaseApp} The app associated with this DatabaseService instance.
+   * @return  The app associated with this DatabaseService instance.
    */
   get app(): FirebaseApp {
     return this.appInternal;
@@ -123,6 +123,11 @@ class DatabaseRulesClient {
   private readonly httpClient: AuthorizedHttpClient;
 
   constructor(app: FirebaseApp, dbUrl: string) {
+    const emulatorHost = process.env.FIREBASE_DATABASE_EMULATOR_HOST;
+    if (emulatorHost) {
+      dbUrl = `http://${emulatorHost}`;
+    }
+
     const parsedUrl = new URL(dbUrl);
     parsedUrl.pathname = path.join(parsedUrl.pathname, RULES_URL_PATH);
     this.dbUrl = parsedUrl.toString();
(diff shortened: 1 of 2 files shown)


yuchenshi (approved):

yuchenshi:
Does `integration/database.spec.ts` pass locally with `--updateRules`?  I know I still haven't figured out #1155 but let's see if it at least works locally.

hiranya911:
Looks all good:

```
% firebase emulators:exec --project fake-project-id --only database 'npx mocha test/integration/database.spec.ts --require ts-node/register --updateRules' 
⚠  Could not find config (firebase.json) so using defaults.
i  emulators: Starting emulators: database
⚠  database: Did not find a Realtime Database rules file specified in a firebase.json config file. The emulator will default to allowing all reads and writes. Learn more about this option: https://firebase.google.com/docs/emulator-suite/install_and_configure#security_rules_configuration.
i  database: Database Emulator logging to database-debug.log
i  Running script: npx mocha test/integration/database.spec.ts --require ts-node/register --updateRules


Running integration tests against Emulator Suite. Some tests may be skipped due to lack of emulator support.
  admin.database
    Updating security rules to defaults.
    ✓ admin.database() returns a database client
    ✓ admin.database.ServerValue type is defined
    ✓ default App is not blocked by security rules (311ms)
[2021-03-09T02:25:49.112Z]  @firebase/database: FIREBASE WARNING: set at /blocked failed: permission_denied 
    ✓ App with null auth overrides is blocked by security rules
    ✓ App with non-null auth override is not blocked by security rules
    ✓ admin.database().getRules() returns currently defined rules as a string
    ✓ admin.database().getRulesJSON() returns currently defined rules as an object
    admin.database().ref()
      ✓ ref() can be called with ref
      ✓ set() completes successfully
      ✓ once() returns the current value of the reference (80ms)
      ✓ child().once() returns the current value of the child
      ✓ remove() completes successfully
    app.database(url).ref()
      ✓ app.database(url) returns a Database client for URL
      ✓ set() completes successfully
      ✓ once() returns the current value of the reference
      ✓ child().once() returns the current value of the child
      ✓ remove() completes successfully


  17 passing (1s)

✔  Script exited successfully (code 0)
i  emulators: Shutting down emulators.
i  database: Stopping Database Emulator
i  hub: Stopping emulator hub
```

src/database/database-internal.ts
```diff
@@ -63,7 +63,7 @@ export class DatabaseService {
   /**
    * Returns the app associated with this DatabaseService instance.
    *
-   * @return {FirebaseApp} The app associated with this DatabaseService instance.
+   * @return  The app associated with this DatabaseService instance.
```
schmidt-sebastian, line 66: Nit: Remove extra space

test/unit/database/database.spec.ts
```diff
@@ -409,5 +406,101 @@ describe('Database', () => {
...
+
+    describe('emulator mode', () => {
+      interface EmulatorTestConfig {
+        name: string;
+        setUp: () => FirebaseApp;
+        tearDown?: () => void;
+        url: string;
+      }
+
+      const configs: EmulatorTestConfig[] = [
```
schmidt-sebastian, line 418: Optional: Consider merging these tests with the tests above. You can likely re-use EmulatorTestConfig (once renamed) and also validate the production URLs.

schmidt-sebastian (approved):

src/database/database-internal.ts
hiranya911: Done

test/unit/database/database.spec.ts
hiranya911: This is a pretty good idea. However, upon trying it, required quite a bit of changes to the existing tests. Let's revisit this again in a future PR.

Issue #1193, opened 2021-03-10
Token refresh doesn't work in traffic storm


inlined:
This is a reassignment to close out https://github.com/firebase/firebase-functions/issues/579. The customer's function makes _many many_ Admin SDK API calls every 15 minutes. Previously it would crash with 401 errors because the API server was no longer recognizing the token as valid. The customer was able to work around this error by calling `admin.app().INTERNAL.getToken(/*force_refresh=*/true)` before their code ran, so it's definitely an Admin bug.

Given the frequency of the crashes, I expect that this bug is happening on the border of when a token is about to expire and the Admin SDK isn't properly refreshing it. I have two theories for what could cause this:

1. There may be a bug in the code that actually refreshes the tokens. For example, [here](https://github.com/firebase/firebase-admin-node/blob/master/src/firebase-app.ts#L145) is a bug where a token that is supposed to expire from [1m, 2m) will never get refreshed.
2. This may be an issue where the network pool for the auth token is full of serialized requests and it takes longer to flush than the token is valid for. 

My recommendation would probably be something surgical like changing [this line](https://github.com/firebase/firebase-admin-node/blob/master/src/firebase-app.ts#L120) to subtract a few minutes from the actual expiration time so that we don't serialize requests with a token that's about to expire.

hiranya911:
@inlined how is this function staying hot for 15 minutes (and possibly longer) to experience a token expiration?

inlined:
It's a periodic function that runs every 15m. Thanks to the (newer) architecture in the Cloud Functions backend, the code that that implements `metadata.google.internal` lives in the load balancer and not the container anymore. This means there can be aggressive caching across all functions and even across container death. Assumptions can no longer be made that a "fresh" token will last very long, other than that I _think_ the backend returns a fresh token within a few minutes of the cached token expiring. We can try to dig into the backend source together if you want.

hiranya911:
Thanks @inlined.

@trentbrooks, can I ask you to share the token expiry time logs you've collected from your runs? I see you have shared a couple of log entries at https://github.com/firebase/firebase-functions/issues/579#issuecomment-791724938. Do you have any more?

trentbrooks:
Hi @hiranya911 

Here are a couple of screenshots from today with the log. It's been working consistently since @inlined suggested the fix a few days ago. Function runs scheduled every 15 minutes and sends 250-300 topic messages (promises), and function finishes each run in 1500-5500 milliseconds. 

[...]

PR #1194, opened 2021-03-11
fix: Decoupled proactive token refresh from FirebaseApp

hiranya911:
Our existing code for OAuth2 token management is quite complicated. This is due to `FirebaseApp` doing too much -- namely handling token caching as well proactive token refreshing. The proactive token refresh was added to make our RTDB connections more stable, but having this logic in `FirebaseApp` impacts other APIs in the SDK too. For instance this is why the SDK won't let the Node.js process terminate, unless you call `app.delete()`. 

In order to simplify our token management, and align more with other GCP libraries that deal with tokens, I'm proposing to separate the proactive token refresh from the `FirebaseApp` class. To that end this PR implements the following changes:

1. Remove the proactive token refresh logic from `FirebaseApp`.
2. Simplify the token caching and fetching behavior.
3. Treat any cached token that's at most 5 minutes away from expiration to be already expired. 
4. Implement the proactive token refresh as part of the `database` service, where it will only get initialized if a developer starts an RTDB connection.
5. Remove the existing complicated retry logic in the proactive token refresh.

Change no. 3 is pretty standard across GCP libraries (see [here](https://github.com/googleapis/google-auth-library-nodejs/blob/fbad5d383be691099a7dab29dca27d5263c4dfb1/src/auth/oauth2client.ts#L1320-L1325) and [here](https://github.com/googleapis/google-auth-library-nodejs/blob/fbad5d383be691099a7dab29dca27d5263c4dfb1/src/auth/authclient.ts#L94)). We never implemented this in the Admin Node.js SDK since the token refresher was indirectly handling it. But it appears in certain environments we cannot rely on the token refresher to keep the auth token current. Therefore by eagerly treating cached tokens as expired, we can mitigate issues like #1193 entirely.

As for change no. 5, the existing retry logic in the token refresher is rather error prone and redundant. If something goes wrong with a proactive refresh attempt, the worst that could happen is any active RTDB connections dropping out upon the expiration of the current token. The RTDB SDK automatically reconnects when this happens (provided it can fetch a new token at that point). This is what we do in our Java SDK too, so there's some prior art as well.


src/database/database-internal.ts
```diff
@@ -96,9 +106,45 @@ export class DatabaseService {
...
+    if (!this.tokenRefresh) {
+      this.tokenRefresh = true;
+      this.appInternal.INTERNAL.addAuthTokenListener(this.onTokenChange);
+    }
+
     return db;
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  private onTokenChange(_: string): void {
```
inlined, line 119: IIRC there was no other token refresh listener. If the right action is to immediately ask for a new token from app.INTERNAL should we just remove the parameter?

src/database/database-internal.ts
```diff
@@ -96,9 +106,45 @@ export class DatabaseService {
...
+        const delayMillis = token.expirationTime - TOKEN_REFRESH_THRESHOLD_MILLIS - Date.now();
+        // If the new token is set to expire soon (unlikely), do nothing. Somebody will eventually
+        // notice and refresh the token, at which point this callback will fire again.
+        if (delayMillis > 0) {
+          this.scheduleTokenRefresh(delayMillis);
+        }
+      })
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      .catch((_) => {
+        // Unlikely error, since a new token was just fetched before the callback fired.
```
inlined, line 131: I'm used to the idea that we should at least log a caught error. Dropping it silently sounds like we'll have a really hard time debugging any bugs here.

src/database/database-internal.ts
```diff
@@ -96,9 +106,45 @@ export class DatabaseService {
...
+        // Just here to prevent the promise chain from crashing.
+      });
+  }
+
+  private scheduleTokenRefresh(delayMillis: number): void {
+    clearTimeout(this.tokenRefreshTimeout);
+    this.tokenRefreshTimeout = setTimeout(() => {
+      this.appInternal.INTERNAL.getToken(true)
```
inlined, line 139: nit: I don't know if it conflicts with the style guide here, but style guides internally say you should comment to name any boolean variable names:

```js
this.appInternal.INTERNAL.getToken(/*forceRefresh=*/true);
```

src/firebase-app.ts
```diff
@@ -57,129 +59,80 @@ export interface FirebaseAccessToken {
...
+  public getToken(forceRefresh = false): Promise<FirebaseAccessToken> {
+    if (forceRefresh || this.shouldRefresh()) {
+      return this.refreshToken();
     }
+
+    return Promise.resolve(this.cachedToken_);
+  }
+
+  private refreshToken(): Promise<FirebaseAccessToken> {
+    return Promise.resolve(this.credential_.getAccessToken())
```
inlined, line 78: I'm confused. isn't getAccesssToken() already returning a promise? Why do we need to resolve the promise instead of just chaining?

inlined (approved):
Absolutely awesome turnaround time. Thanks Hiranya!

src/database/database-internal.ts
hiranya911: I think an even better approach would be to change the signature of the listener to accept a `FirebaseAccessToken` which contains both token and expiry time. That way we wouldn't have to call `getToken()` in the listener at all (we currently do it to get hold of the expiry time). 

However, this API currently mirrors an API in the JS SDK:

https://github.com/firebase/firebase-js-sdk/blob/ec6ccc7d319732588ceacae0824d6c7551afb026/packages/app-types/private.d.ts#L76

The RTDB implementation expects our App and App.INTERNAL APIs to be compatible with the JS SDK. Therefore we'll need to check with the JS SDK team before making any changes to the listener signature.

src/firebase-app.ts
hiranya911: I thought so too :)

But when I made the change, it actually caused a test failure here:
https://github.com/firebase/firebase-admin-node/blob/994fd43e36c0824deba176763010dc141e9bb0f4/test/unit/firebase-app.spec.ts#L677-L691

This is because the `credential` in this case returns a value which cannot be chained. I'm not sure how much we care about this sort of edge cases, but for now I've opted to keep it as is.

src/database/database-internal.ts
hiranya911: Done

src/database/database-internal.ts
hiranya911: Done

src/database/database-internal.ts
```diff
@@ -28,9 +28,13 @@ import { getSdkVersion } from '../utils/index';
 
 import Database = database.Database;
 
+const TOKEN_REFRESH_THRESHOLD_MILLIS = 5 * 60 * 1000;
+
 export class DatabaseService {
 
   private readonly appInternal: FirebaseApp;
+  private tokenRefresh: boolean;
```
schmidt-sebastian, line 36: This doesn't sound like a boolean variable to me (but naming is hard and criticizing is easy)

src/database/database-internal.ts
```diff
@@ -96,9 +106,43 @@ export class DatabaseService {
...
+      .then((token) => {
+        const delayMillis = token.expirationTime - TOKEN_REFRESH_THRESHOLD_MILLIS - Date.now();
+        // If the new token is set to expire soon (unlikely), do nothing. Somebody will eventually
+        // notice and refresh the token, at which point this callback will fire again.
+        if (delayMillis > 0) {
+          this.scheduleTokenRefresh(delayMillis);
+        }
+      })
+      .catch((err) => {
+        console.error('Unexpected error while attempting to schedule a token refresh:', err);
```
schmidt-sebastian, line 130: It think the error most likely occurred during `getToken()` and not during the scheduling step.

src/database/database-internal.ts
```diff
@@ -96,9 +106,43 @@ export class DatabaseService {
 
       this.databases[dbUrl] = db;
     }
+
+    if (!this.tokenRefresh) {
+      this.tokenRefresh = true;
```
schmidt-sebastian, line 111: Optional suggestion:

You can make this a little simpler if you combine `tokenRefresh` and `tokenRefreshTimeout`. You could initialize `tokenRefreshTimeout` to `undefined` initially and then set the schedule right here, which seems more intuitive to me. With the current setup, I have to look at `onTokenChange` to figure out the scheduling.

I also wonder if we should use `setInterval`. It seems a little more appropriate, but might make things more complicated in some places.

src/firebase-app.ts
```diff
@@ -57,129 +59,80 @@ export interface FirebaseAccessToken {
...
-            'present at https://console.firebase.google.com/iam-admin/serviceaccounts/project. If ' +
-            'not, generate a new key file at ' +
-            'https://console.firebase.google.com/project/_/settings/serviceaccounts/adminsdk.';
-          }
-
-          throw new FirebaseAppError(AppErrorCodes.INVALID_CREDENTIAL, errorMessage);
-        });
-
-      return this.cachedTokenPromise_;
+  public getToken(forceRefresh = false): Promise<FirebaseAccessToken> {
```
schmidt-sebastian, line 69: Can the RTDB component just set `forceRefresh` to true if the expiration time has been hit? That way, we only need to deal with TOKEN_REFRESH_THRESHOLD_MILLIS in one place.

schmidt-sebastian (approved):

src/database/database-internal.ts
lahirumaramba: `refreshToken` or `shouldRefreshToken` might work... I agree with "naming is hard and criticizing is easy" :)

src/database/database-internal.ts
```diff
@@ -50,6 +54,12 @@ export class DatabaseService {
    * @internal
    */
   public delete(): Promise<void> {
+    if (this.tokenRefresh) {
+      this.appInternal.INTERNAL.removeAuthTokenListener(this.onTokenChange);
+      clearTimeout(this.tokenRefreshTimeout);
```
lahirumaramba, line 59: Does it matter the what order we disarm here? Could this cause a race condition?

lahirumaramba (approved):
Thanks, Hiranya! LGTM!
Left a couple comments.

src/database/database-internal.ts
hiranya911: True. But overall this function is attempting to schedule a new token refresh event. It calls `getToken()` just to figure out when to schedule that event.

src/database/database-internal.ts
hiranya911: I felt like we should unregister the listener before clearing out the timer. If we clear the timer first, theoretically it's possible for the listener to fire again and re-initialize the timer.

src/database/database-internal.ts
hiranya911: This code is just intended to register the token change listener. There's no guarantee that the listener will fire immediately. So theoretically it is possible for `tokenRefresh` to be true, while `tokenRefreshTimeout` is not yet set. What we really need is a way to track whether we have registered the listener or not (so we can avoid duplicate registrations, and also unregister it at delete). So I'm renaming `tokenRefresh` to `tokenListenerRegistered`. That should clear up any confusions, and also address the above comment regarding naming.

As for `setInterval()`, given that we can receive tokens with arbitrary expiration times from auth (we've seen tokens with expiry times like 30min and 45min), an interval scheduler is not very useful. So we use the token listener to watch for token changes, and schedule refresh events according to the expiry time we get.

src/firebase-app.ts
hiranya911: These 2 constants are used for slightly different purposes:

1. In firebase-app it is used to determine how early we should treat a valid token as expired.
2. In database-internal it is used to determine when to schedule the next refresh event.

These don't have to be the same value. For instance, we can have FirebaseApp invalidate a token 5 minutes before it expires, but have RTDB trigger the token refresh 10 minutes before the current token expires.

To make the above distinction clear, I'm renaming the constant in firebase-app to `TOKEN_EXPIRY_THRESHOLD_MILLIS`. This constant doesn't have anything to do with refreshing tokens. It's simply when we treat the token as expired.

src/database/database-internal.ts
hiranya911: Renamed to `tokenListenerRegistered`. See below.

PR #1197, opened 2021-03-15
fix: Separated service namespaces from module entry points

hiranya911:
Currently, our new module entry points expose both the new API members and the old API members. For example, the `firebase-admin/auth` entry point exposes both `getAuth()` (new) and `auth()` (old).  However, the old API members in these entry points only exist in their typings. The implementation of the `auth()` function is obviously somewhere else -- namely the `FirebaseNamespace` class. This means code like following will be allowed by the TS compiler, but will not actually run:

```typescript
import { auth } from 'firebase-admin/auth';

auth();
```

In order to prevent developers from erroneously writing code like the above, I'm removing the old API members entirely from the module entry points. Going forward, the old API members will only be available through the global `admin` import:

```typescript
import * as admin from 'firebase-admin';

admin.auth();
```

To implement this separation, I've taken the following steps:
* Each service now implements a new `<serviceName>-namespace.ts` module file, that defines the old API members. For example Auth gets an `auth-namespace.ts` module which contains `auth()` function+namespace.
* The module entry points (index files) do not make any references to the namespace.ts files.
* All namespace.ts files are combined into a single admin namespace in the `firebase-admin-api.ts` module.


lahirumaramba (approved):
Great! LGTM!

PR #1203, opened 2021-03-19
fix(rtdb): Fixing the RTDB token listener callback

hiranya911:
#1194 has introduced a regression that is causing an [integration test failure](https://github.com/firebase/firebase-admin-node/actions/runs/668209320):

```
Error: Credential implementation provided to initializeApp() via the "credential" property failed to fetch a valid Google OAuth2 access token with the following error: "Cannot read property 'appInternal' of undefined".
      at FirebaseAppError.FirebaseError [as constructor] (lib/utils/error.js:44:28)
      at FirebaseAppError.PrefixedFirebaseError [as constructor] (lib/utils/error.js:90:28)
      at new FirebaseAppError (lib/utils/error.js:125:28)
      at /home/runner/work/firebase-admin-node/firebase-admin-node/lib/firebase-app.js:81:19
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

This is because we are registering an instance method as the token listener callback in `DatabaseService`. Instance methods loose all their context when passed as callback functions. Specifically, references like `this` will not work as intended. Consequently, `this.appInternal` in our callback doesn't resolve.

To mitigate this issue we can explicitly `bind()` the instance method before passing it as a callback. This also enables us to use the bound callback method as an indicator of whether the token listener has been registered or not. 

I also investigated why this issue didn't pop up during unit tests, and realized our unit tests have been silently failing due to some missing promise returns 🤦 . I've fixed those instances too. However, one of the test cases do not work as originally intended. We need a change to the token listener interface to be able to fix it, so in the meantime I've added a comment there, and skipped the test.

### DIFF
--- a/src/database/database-internal.ts
+++ b/src/database/database-internal.ts
@@ -33,7 +33,7 @@ const TOKEN_REFRESH_THRESHOLD_MILLIS = 5 * 60 * 1000;
 export class DatabaseService {
 
   private readonly appInternal: FirebaseApp;
-  private tokenListenerRegistered: boolean;
+  private tokenListener: (token: string) => void;
   private tokenRefreshTimeout: NodeJS.Timeout;
 
   private databases: {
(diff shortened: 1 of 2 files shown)


lahirumaramba (approved):
Thanks for the fix, Hiranya!

PR #1204, opened 2021-03-21
(chore): Add JWT Decoder and Signature Verifier

lahirumaramba:
- Refactoring verify token take 2
- Add JWT Decoder class


hiranya911 (changes requested):
I think this needs a bit more work in terms of getting the abstractions right. I've proposed some ideas. PTAL.

src/utils/jwt-decoder.ts
```diff
@@ -0,0 +1,125 @@
...
+      });
+    }
+
+    const header = fullDecodedToken?.header;
+    const payload = fullDecodedToken?.payload;
+
+    return { header, payload };
+  }
+
+  public isSignatureValid(jwtToken: string, publicKey: string | null): Promise<boolean> {
```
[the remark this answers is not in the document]
hiranya911: For 1 and 3 we can add even more abstractions to further customize the behavior. For example, we can have a `PublicKeyFetcher` interface for 1 that knows how to fetch public keys from different sources, and so on. But we probably don't need that level of customization right now.

src/auth/token-verifier.ts
```diff
@@ -149,27 +158,54 @@ export class FirebaseTokenVerifier {
 
     return util.findProjectId(this.app)
       .then((projectId) => {
-        return this.verifyJWTWithProjectId(jwtToken, projectId, isEmulator);
+        return Promise.all([this.safeDecode(jwtToken), projectId]);
```
hiranya911, line 161: You can untangle this with some helper functions:

```typescript
return this.ensureProjectId()
      .then((projectId) => {
        return this.decodeAndVerify(jwtToken, projectId, isEmulator);
      })
      .then((decoded) => {
        return decoded.payload as DecodedIdToken;
      });
```

```typescript
  private ensureProjectId(): Promise<string> {
    // get project ID or reject
    return Promise.resolve('');
  }

  private decodeAndVerify(token: string, projectId: string, isEmulator: boolean): Promise<Decoded> {
    return this.verifyContent(token, projectId, isEmulator)
      .then((decoded) => {
        return this.verifySignature(token, isEmulator)
          .then(() => decoded);
      });
  }

  private verifyContent(token: string, projectId: string, isEmulator: boolean): Promise<Decoded> {
    // Decode, validate and return
    return Promise.resolve({} as any);
  }

  private verifySignature(token: string, isEmulator: boolean): Promise<void> {
    return Promise.resolve();
  }
```

hiranya911 (changes requested):
Overall approach and design is correct. I found a bunch of nits that can be cleaned up. I haven't looked at all the error handling bits yet. I'll get to that in the next round.

src/auth/token-verifier.ts
```diff
@@ -69,14 +72,13 @@ export interface FirebaseTokenInfo {
...
+ * Class for verifying ID tokens and session cookies.
  */
 export class FirebaseTokenVerifier {
-  private publicKeys: {[key: string]: string};
-  private publicKeysExpireAt: number;
   private readonly shortNameArticle: string;
+  private readonly signatureVerifier: PublicKeySignatureVerifier;
 
-  constructor(private clientCertUrl: string, private algorithm: jwt.Algorithm,
+  constructor(clientCertUrl: string, private algorithm: jwt.Algorithm,
```
lahirumaramba, line 81: I did not remove the `algorithm` from here, but we can use the `const ALGORITHM_RS256` directly here instead of passing it around.
We have a unit test: `should not decode a signed token when the algorithm is set to none (emulator)`, which will become obsolete if we use the const.

src/utils/jwt-signature-verifier.ts
```diff
@@ -0,0 +1,235 @@
...
+    else if (!validator.isNonEmptyString(algorithm)) {
+      throw new Error('The provided JWT algorithm is an empty string.');
+    }
+
+    this.publicKeyProvider = new PublicKeyFetcher(clientCertUrl, app);
+    this.getKeyHandler = this.getKey.bind(this);
+  }
+
+  public verify(token: string): Promise<void> {
+    const error = new SignatureVerifierError({
```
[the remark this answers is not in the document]
lahirumaramba: `isSignatureValid()` returns a boolean, if we throw this in `isSignatureValid` it will never false. Having it return a boolean will let us loop through multiple public keys if we ever needed to in the future. WDYT?

src/auth/token-verifier.ts
hiranya911: The reason I'm keen on dropping the algorithm parameter from this API is that it's `jsonwebtoken` specific (the type is `jwt.Algorithm`). I think we can just hardcode RS256 into this implementation, and drop the parameter. It's extremely unlikely to change. And of course change/remove any affected tests too.

src/utils/jwt.ts
```diff
@@ -0,0 +1,240 @@
...
+export class PublicKeySignatureVerifier implements SignatureVerifier {
+  constructor(private keyFetcher: UrlKeyFetcher) {
+    if (!validator.isNonNullObject(keyFetcher)) {
+      throw new Error('The provided key fetcher is not an object or null.');
+    }
+  }
+
+  public verify(token: string): Promise<void> {
+    const ALGORITHM_RS256: jwt.Algorithm = 'RS256' as const;
+    const error = new JwtError(JwtErrorCode.INVALID_TOKEN,
```
hiranya911, line 125: This bit is duplicated in the other implementation. If you change the `isSignatureValue()` helper to not return a boolean, and instead signal errors by rejecting, you can avoid this. Given a signature like:

```
function verifyJwtSignature(token, getKey, options): Promise<void>
```

this will become:

```
return verifyJwtSignature(token, getKeyCallback(this.keyFetcher), {algorithms: [RS_256]});
```

src/utils/jwt.ts
```diff
@@ -0,0 +1,240 @@
...
+export enum JwtErrorCode {
+  INVALID_ARGUMENT = 'invalid-argument',
+  INVALID_CREDENTIAL = 'invalid-credential',
+  TOKEN_EXPIRED = 'token-expired',
+  INVALID_TOKEN = 'invalid-token',
+  NO_MATCHING_KID = 'no-matching-kid-error',
+  INTERNAL_ERROR = 'internal-error',
+}
+
+const NO_MATCHING_KID_ERROR_MESSAGE = 'no-matching-kid-error';
```
hiranya911, line 240: Move up the file.

src/utils/jwt.ts
```diff
@@ -0,0 +1,240 @@
...
+ */
+export class PublicKeySignatureVerifier implements SignatureVerifier {
+  constructor(private keyFetcher: UrlKeyFetcher) {
+    if (!validator.isNonNullObject(keyFetcher)) {
+      throw new Error('The provided key fetcher is not an object or null.');
+    }
+  }
+
+  public verify(token: string): Promise<void> {
+    const ALGORITHM_RS256: jwt.Algorithm = 'RS256' as const;
```
hiranya911, line 124: Define as a module-level constant.

src/utils/jwt.ts
```diff
@@ -0,0 +1,240 @@
...
+      throw err;
+    });
+  }
+}
+
+/**
+ * Verifies JWT signature with a public key.
+ */
+export class PublicKeySignatureVerifier implements SignatureVerifier {
+  constructor(private keyFetcher: UrlKeyFetcher) {
```
hiranya911, line 117: Should accept any `KeyFetcher`.

src/utils/jwt.ts
```diff
@@ -0,0 +1,240 @@
...
+        const resp = err.response;
+        if (resp.isJson() && resp.data.error) {
+          errorMessage += `${resp.data.error}`;
+          if (resp.data.error_description) {
+            errorMessage += ' (' + resp.data.error_description + ')';
+          }
+        } else {
+          errorMessage += `${resp.text}`;
+        }
+        throw new JwtError(JwtErrorCode.INTERNAL_ERROR, errorMessage);
```
hiranya911, line 106: Does it make sense to throw JwtError from this level? I'd think jsonwebtoken will catch the error and convert it into one of its error types.

src/utils/jwt.ts
```diff
@@ -0,0 +1,240 @@
...
+  verify(token: string): Promise<void>;
+}
+
+interface KeyFetcher {
+  fetchPublicKeys(): Promise<{ [key: string]: string }>;
+}
+
+export class UrlKeyFetcher implements KeyFetcher {
+  private publicKeys: { [key: string]: string };
+  private publicKeysExpireAt: number;
```
hiranya911, line 39: Initialize this to 0. That will help simplify some of the remaining logic.

src/utils/jwt.ts
```diff
@@ -0,0 +1,240 @@
...
+      return this.refresh();
+    }
+    return Promise.resolve(this.publicKeys);
+  }
+
+  private shouldRefresh(): boolean {
+    const publicKeysExist = (typeof this.publicKeys !== 'undefined');
+    const publicKeysExpiredExists = (typeof this.publicKeysExpireAt !== 'undefined');
+    const publicKeysStillValid = (publicKeysExpiredExists && Date.now() < this.publicKeysExpireAt);
+    return !(publicKeysExist && publicKeysStillValid);
```
hiranya911, line 65: This can be just:

```
return !this.publicKeys || this.publicKeysExpireAt < Date.now();
```

src/utils/jwt.ts
```diff
@@ -0,0 +1,240 @@
...
+        throw new HttpError(resp);
+      }
+      if (Object.prototype.hasOwnProperty.call(resp.headers, 'cache-control')) {
+        const cacheControlHeader: string = resp.headers['cache-control'];
+        const parts = cacheControlHeader.split(',');
+        parts.forEach((part) => {
+          const subParts = part.trim().split('=');
+          if (subParts[0] === 'max-age') {
+            const maxAge: number = +subParts[1];
+            this.publicKeysExpireAt = Date.now() + (maxAge * 1000);
```
hiranya911, line 88: It seems this gets set conditionally. I think for other cases, we should 0 it out just to be on the safe side.

src/auth/token-verifier.ts
```diff
@@ -128,6 +130,9 @@ export class FirebaseTokenVerifier {
     }
     this.shortNameArticle = tokenInfo.shortName.charAt(0).match(/[aeiou]/i) ? 'an' : 'a';
 
+    this.signatureVerifier = new PublicKeySignatureVerifier(
```
hiranya911, line 133: If we want we can un-export the `UrlKeyFetcher`, and provide a static API like:

```
PublicKeySignaureVerifier.withCertificateUrl(url, agent);
```

src/auth/token-verifier.ts
```diff
@@ -147,29 +152,70 @@ export class FirebaseTokenVerifier {
...
+    return this.safeDecode(token).then((decodedToken) => {
+      this.validateTokenContent(decodedToken, projectId, isEmulator);
+      return Promise.resolve(decodedToken);
     });
+  }
+
+  private safeDecode(jwtToken: string): Promise<DecodedToken> {
+    return decodeJwt(jwtToken)
+      .catch((err) => {
+        if (!(err instanceof JwtError)) {
```
hiranya911, line 198: Can we somehow guarantee that decodeJwt only ever throws JwtError instances?

src/auth/token-verifier.ts
```diff
@@ -147,29 +152,70 @@ export class FirebaseTokenVerifier {
...
+          .then(() => decoded);
+      });
+  }
 
-    const fullDecodedToken: any = jwt.decode(jwtToken, {
-      complete: true,
+  private verifyContent(token: string, projectId: string, isEmulator: boolean): Promise<DecodedToken> {
+    return this.safeDecode(token).then((decodedToken) => {
+      this.validateTokenContent(decodedToken, projectId, isEmulator);
+      return Promise.resolve(decodedToken);
```
hiranya911, line 191: return decodedToken

src/auth/token-verifier.ts
```diff
@@ -147,29 +152,70 @@ export class FirebaseTokenVerifier {
...
+            AuthClientErrorCode.INVALID_CREDENTIAL,
+            'Must initialize app with a cert credential or set your Firebase project ID as the ' +
+            `GOOGLE_CLOUD_PROJECT environment variable to call ${this.tokenInfo.verifyApiName}.`,
+          );
+        }
+        return Promise.resolve(projectId);
+      })
+  }
+
+  private decodeAndVerify(token: string, projectId: string, isEmulator: boolean): Promise<DecodedToken> {
```
hiranya911, line 180:
```typescript
return this.safeDecode(token)
  .then((decoded) => {
    this.validateTokenContent(decoded, projectId, isEmulator);
    return this.verifySignature(token, isEmulator)
      .then(() => decoded);
  })
```

src/auth/token-verifier.ts
```diff
@@ -217,122 +260,42 @@ export class FirebaseTokenVerifier {
...
-          ),
-        );
-      } else {
-        return this.verifyJwtSignatureWithKey(jwtToken, publicKeys[header.kid]);
-      }
-
-    });
+  private verifySignature(jwtToken: string, isEmulator: boolean):
+    Promise<void> {
+    const verifier = isEmulator ? new EmulatorSignatureVerifier() : this.signatureVerifier;
```
hiranya911, line 269: May be we can define a module-level constant for emulator verifier, and use that instance for all invocations.

src/auth/token-verifier.ts
```diff
@@ -217,122 +260,42 @@ export class FirebaseTokenVerifier {
...
-   *
-   * @return {Promise<object>} A promise fulfilled with public keys for the Google certs.
-   */
-  private fetchPublicKeys(): Promise<{[key: string]: string}> {
-    const publicKeysExist = (typeof this.publicKeys !== 'undefined');
-    const publicKeysExpiredExists = (typeof this.publicKeysExpireAt !== 'undefined');
-    const publicKeysStillValid = (publicKeysExpiredExists && Date.now() < this.publicKeysExpireAt);
-    if (publicKeysExist && publicKeysStillValid) {
-      return Promise.resolve(this.publicKeys);
+    if (!(error instanceof JwtError)) {
```
hiranya911, line 279: Just curious as to what other error conditions can pop up here. Can we guarantee that this API only ever sees JwtError instances?

src/utils/jwt.ts
```diff
@@ -0,0 +1,240 @@
...
+  constructor(readonly code: JwtErrorCode, readonly message: string) {
+    super(message);
+    (this as any).__proto__ = JwtError.prototype;
+  }
+}
+
+/**
+ * JWT error codes.
+ */
+export enum JwtErrorCode {
```
hiranya911, line 231: Let's define a specific `INVALID_SIGNATURE` error code for signature verification errors.

src/utils/jwt.ts
```diff
@@ -0,0 +1,240 @@
...
+        if (error.name === 'TokenExpiredError') {
+          return reject(new JwtError(JwtErrorCode.TOKEN_EXPIRED,
+            'The provided token has expired. Get a fresh token from your ' +
+            'client app and try again.'));
+        } else if (error.name === 'JsonWebTokenError') {
+          const prefix = 'error in secret or public key callback: ';
+          if (error.message && error.message.includes(prefix)) {
+            const message = error.message.split(prefix).pop() || 'Error fetching public keys.';
+            const code = (message === NO_MATCHING_KID_ERROR_MESSAGE) ? JwtErrorCode.NO_MATCHING_KID :
+              JwtErrorCode.INVALID_ARGUMENT;
```
hiranya911, line 181: `JwtErrorCode.KEY_FETCH_ERROR` or similar

src/utils/jwt.ts
```diff
@@ -0,0 +1,240 @@
...
+            return reject(new JwtError(code, message));
+          }
+          return resolve(false);
+        }
+        return reject(new JwtError(JwtErrorCode.INVALID_TOKEN, error.message));
```
hiranya911, line 186: `JwtErrorCode.INVALID_SIGNATURE` or similar

hiranya911 (changes requested):
Thanks for addressing all the feedback. Looks pretty good! Few more things to revise and we are good to go.

src/utils/jwt.ts
```diff
@@ -0,0 +1,240 @@
...
+      (error: jwt.VerifyErrors | null) => {
+        if (!error) {
+          return resolve(true);
+        }
+        if (error.name === 'TokenExpiredError') {
+          return reject(new JwtError(JwtErrorCode.TOKEN_EXPIRED,
+            'The provided token has expired. Get a fresh token from your ' +
+            'client app and try again.'));
+        } else if (error.name === 'JsonWebTokenError') {
+          const prefix = 'error in secret or public key callback: ';
```
hiranya911, line 177: Should probably define a constant for this too: `JWT_CALLBACK_ERROR_PREFIX` or similar. And add a comment indicating how it came to be.

src/utils/jwt-signature-verifier.ts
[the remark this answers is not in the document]
hiranya911: See my latest comments. My recommendation is still to go with a `Promise<void>` return type (it will help reduce some of the duplication present now). To verify against multiple keys, we can  always do something like:

```
return Promise.any(keys.map((key) => verifySignature(token, key)));
```

lahirumaramba (commented):
Thank you for the feedback!!
Updated the PR.
I removed the algorithm parameter from the API and updated the unit tests to reflect that change.
I will follow up with another update to add unit tests for the new components. Thanks!

src/utils/jwt.ts
lahirumaramba: Good call! Added the following at line 88 to reset the parameter.
```
// reset expire at from previous set of keys.
this.publicKeysExpireAt = 0;
```

src/utils/jwt.ts
```diff
@@ -0,0 +1,237 @@
...
+          return resolve();
+        }
+        if (error.name === 'TokenExpiredError') {
+          return reject(new JwtError(JwtErrorCode.TOKEN_EXPIRED,
+            'The provided token has expired. Get a fresh token from your ' +
+            'client app and try again.'));
+        } else if (error.name === 'JsonWebTokenError') {
+          if (error.message && error.message.includes(JWT_CALLBACK_ERROR_PREFIX)) {
+            const message = error.message.split(JWT_CALLBACK_ERROR_PREFIX).pop() || 'Error fetching public keys.';
+            const code = (message === NO_MATCHING_KID_ERROR_MESSAGE) ? JwtErrorCode.KEY_FETCH_ERROR :
```
hiranya911, line 179: I feel like this should be something like:

```
const code = (message === NO_MATCHING_KID_ERROR_MESSAGE) ? JwtErrorCode.NO_MATCHING_KID :
  JwtErrorCode.KEY_FETCH_ERROR;
```

src/utils/jwt.ts
```diff
@@ -0,0 +1,237 @@
...
+              JwtErrorCode.INVALID_ARGUMENT;
+            return reject(new JwtError(code, message));
+          }
+          return reject(new JwtError(JwtErrorCode.INVALID_SIGNATURE,
```
hiranya911, line 183: Can we drop this and let it be rejected below?

src/auth/token-verifier.ts
```diff
@@ -69,27 +73,20 @@ export interface FirebaseTokenInfo {
...
 
 /**
- * Class for verifying general purpose Firebase JWTs. This verifies ID tokens and session cookies.
+ * Class for verifying ID tokens and session cookies.
  */
 export class FirebaseTokenVerifier {
-  private publicKeys: {[key: string]: string};
-  private publicKeysExpireAt: number;
   private readonly shortNameArticle: string;
+  private readonly signatureVerifier: PublicKeySignatureVerifier;
```
hiranya911, line 80: Nit: Type this as `SignatureVerifier` instead?

src/auth/token-verifier.ts
hiranya911: Since we don't call that  API directly, I think that case can never occur right? We will only see `JwtError` here.

src/auth/token-verifier.ts
```diff
@@ -217,135 +245,54 @@ export class FirebaseTokenVerifier {
...
-      }
-      throw err;
-    });
+    if (error.code === JwtErrorCode.TOKEN_EXPIRED) {
+      const errorMessage = `${this.tokenInfo.jwtName} has expired. Get a fresh ${this.tokenInfo.shortName}` +
+        ` from your client app and try again (auth/${this.tokenInfo.expiredErrorCode.code}).` +
+        verifyJwtTokenDocsMessage;
+      return new FirebaseAuthError(this.tokenInfo.expiredErrorCode, errorMessage);
+    }
+    else if (error.code === JwtErrorCode.INVALID_SIGNATURE) {
```
hiranya911, line 273: In-line: `} else if (`

src/auth/token-verifier.ts
```diff
@@ -27,7 +29,7 @@ import DecodedIdToken = auth.DecodedIdToken;
 // Audience to use for Firebase Auth Custom tokens
 const FIREBASE_AUDIENCE = 'https://identitytoolkit.googleapis.com/google.identity.identitytoolkit.v1.IdentityToolkit';
 
-export const ALGORITHM_RS256 = 'RS256';
+const ALGORITHM_RS256 = 'RS256' as const;
```
hiranya911, line 32: Does it make sense to expose this constant in utils/jwt and import it here?

hiranya911 (approved):
I'm at an LGTM with a few open comments/suggestions.

lahirumaramba:
Thank you, @hiranya911 !
I added unit tests to `utils/jwt` and modified some tests in `auth/token-verifier`.
Let me know what you think.

test/unit/auth/token-verifier.spec.ts
```diff
@@ -593,110 +490,5 @@ describe('FirebaseTokenVerifier', () => {
       await tokenVerifier.verifyJWT(idTokenNoHeader)
         .should.eventually.be.rejectedWith('Firebase ID token has no "kid" claim.');
     });
```
hiranya911, line 492: Ideally we need some tests to ensure that:

1. KEY_FETCH_ERROR is handled gracefully, and
2. AppOptions.httpAgent is passed to the verifier.

hiranya911 (approved):
Looks pretty good. A couple of suggestions for improvement.

lahirumaramba:
Thank you @hiranya911 !
Added two new tests to cover the following:

> 1. KEY_FETCH_ERROR is handled gracefully, and
> 2. AppOptions.httpAgent is passed to the verifier.

PR #1218, opened 2021-04-05
chore: Experimental markdown docs

hiranya911:
Experimental markdown docs generated from the API Documenter.

Process followed:

### Admin SDK repo

```
git clone https://github.com/firebase/firebase-admin-node
cd firebase-admin-node
git checkout modular sdk
npm install
npm run api-extractor:local
```

### JS SDK repo

```
git clone https://github.com/firebase/firebase-js-sdk
yarn
npx ts-node-script repo-scripts/api-documenter/src/start.ts markdown --input path/to/admin/sdk/repo/temp --output adminsdk
```

Most of the generated markdown content looks ok. Open issues to be addressed:

A. The API Documenter doesn't correctly handle old namespaces. Specifically, cases like `admin.auth` being both a function and a namespace are not reflected correctly.
B. Exports from external packages like RTDB and Firestore are not listed.
C. Some invalid tsdoc annotations in the Admin SDK source resulting in some warnings.

### DIFF
--- a/docgen/markdown/firebase-admin.app.app.md
+++ b/docgen/markdown/firebase-admin.app.app.md
@@ -1,0 +1,73 @@
+{% extends "_internal/templates/reference.html" %}
+{% block title %}App interface{% endblock title %}
+{% block body %}
+A Firebase app holds the initialization information for a collection of services.
+
+Do not call this constructor directly. Instead, use  to create an app.
+
+<b>Signature:</b>
+
+```typescript
+export interface App 
+```
+
+## Properties
+
+|  Property | Type | Description |
+|  --- | --- | --- |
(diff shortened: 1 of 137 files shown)


hiranya911:
@davideast let me know what you think about the docs content I've pushed to this branch. 

@Feiyang1 please note the open issues I've pointed out above. Do you know if (2) is the expected behavior? If so we  need to find a reasonable workaround for that (possibly as a post-processing step).

hiranya911:
@deast I've removed the devsite header and footer from all files. Scripts used for future reference:

```sh
# replace.sh
INPUT_FILE=$1
sed -n -i '' -e "4,$(($(wc -l < $INPUT_FILE) - 1))p" $INPUT_FILE
```

```
# Mac OS
$ find markdown -iname "*.md" | xargs -n 1 ./replace.sh 
```

davideast (approved):
LGTM

PR #1220, opened 2021-04-06
feat(auth): Added code flow support for OIDC flow.

xil222:
RELEASE NOTE: Added code flow support for OIDC flow(previously only support idToken flow).
RELEASE NOTE: Defined `OAuthResponseType` for specifying responseType either idToken or code.
RELEASE NOTE: Defined two new error codes: `INVALID_OAUTH_RESPONSETYPE` and `MISSING_OAUTH_CLIENT_SECRET`.


src/auth/auth-config.ts
```diff
@@ -806,6 +882,20 @@ export class OIDCConfig implements OIDCAuthProviderConfig {
...
+    }
+    // If we do not set responseType, we have idToken flow
+    // set to true by default.
+    if (typeof response.responseType !== 'undefined') {
+      this.responseType = response.responseType;
+    } else {
+      const responseType = {
+        idToken: true,
+      }
+      this.responseType = responseType;
```
hiranya911, line 897: `this.responseType = {idToken: true};`

src/auth/auth-config.ts
```diff
@@ -650,6 +655,8 @@ export class OIDCConfig implements OIDCAuthProviderConfig {
   public readonly providerId: string;
   public readonly issuer: string;
   public readonly clientId: string;
+  public readonly clientSecret: string;
```
hiranya911, line 658: Can be optional?

test/integration/auth.spec.ts
```diff
@@ -1710,32 +1726,28 @@ describe('admin.auth', () => {
...
+        clientSecret: 'CLIENT_SECRET',
+        responseType: {
+          idToken: false,
+          code: true,
+        },
       };
       // Only above fields should be modified.
       const modifiedConfigOptions = {
-        displayName: 'OIDC_DISPLAY_NAME4',
+        displayName: 'OIDC_DISPLAY_NAME3',
```
hiranya911, line 1743: Also add the providerId here and remove the deepExtend below.

test/unit/auth/auth-config.spec.ts
```diff
@@ -769,6 +783,27 @@ describe('OIDCConfig', () => {
...
+
+    it('should set readonly property expected responseType', () => {
+      const expectedResponseType = {
+        code: true,
+      };
+      expect(config.responseType).to.deep.equal(expectedResponseType);
+    });
+
+    it('should set readonly property default responseType', () => {
+      delete serverResponse.clientSecret;
```
hiranya911, line 798: serverResponse is a variable shared across many test cases. Deep copy it before making changes to it.

src/auth/auth-config.ts
```diff
@@ -816,6 +906,8 @@ export class OIDCConfig implements OIDCAuthProviderConfig {
       providerId: this.providerId,
       issuer: this.issuer,
       clientId: this.clientId,
+      clientSecret: this.clientSecret,
```
hiranya911, line 909: What if clientSecret is undefined?

src/auth/auth-config.ts
```diff
@@ -773,6 +792,63 @@ export class OIDCConfig implements OIDCAuthProviderConfig {
...
+            if (options.responseType && options.responseType.code) {
+              codeType = options.responseType.code;
+            }
+          }
+        }
+      }
+
+      // Exact one of OAuth response type needs to be set to true.
+      if ((idTokenType && codeType) ||
+          (!idTokenType && !codeType)) {
```
hiranya911, line 838: This condition will fire if the developer does something like `options.responseType = {}`. I think that's not technically an error?

test/unit/auth/auth-config.spec.ts
```diff
@@ -892,6 +941,43 @@ describe('OIDCConfig', () => {
       expect(() => OIDCConfig.validate(partialRequest, true)).not.to.throw();
     });
 
+    it('should throw on OAuth responseType contains invalid parameters', () => {
+      const invalidRequest = deepCopy(clientRequest) as any;
+      invalidRequest.responseType.token = true;
```
hiranya911, line 946: Lets make it `invalidRequest.responseType.unknownField = true;` for clarity.

test/unit/auth/auth-config.spec.ts
```diff
@@ -892,6 +941,43 @@ describe('OIDCConfig', () => {
...
+      expect(() => OIDCConfig.validate(invalidRequest, true))
+        .to.throw('"token" is not a valid OAuthResponseType parameter.');
+    });
+
+    it('should not throw when exact one OAuth responseType is true', () => {
```
hiranya911, line 951: exactly one

test/unit/auth/auth-config.spec.ts
```diff
@@ -892,6 +941,43 @@ describe('OIDCConfig', () => {
...
+      const validRequest = deepCopy(clientRequest) as any;
+      validRequest.responseType.code = false;
+      validRequest.responseType.idToken = true;
+      expect(() => OIDCConfig.validate(validRequest, true)).not.to.throw();
+    });
+
+    it('should throw on two OAuth responseTypes set to true', () => {
```
hiranya911, line 958: Can we also test the scenarios where only one responseType is set?

```
responseType = { idToken: true }
responseType = { code: true }
```

I assume these are valid use cases.

hiranya911 (changes requested):
Looks good. I mostly had nits. There might be couple of edge cases that need better handling and some tests.

src/auth/auth-config.ts
xil222: Fixed by setting clientSecret to optional.

src/auth/auth-config.ts
xil222: ok, I added 2 parameters checking whether responseType has been reset.

src/auth/auth-config.ts
```diff
@@ -799,46 +799,51 @@ export class OIDCConfig implements OIDCAuthProviderConfig {
         '"OIDCAuthProviderConfig.clientSecret" must be a valid string.',
       );
     }
-    if (typeof options.responseType !== 'undefined') {
-      let idTokenType = false;
-      let codeType = false;
+    if (validator.isNonNullObject(options.responseType) && typeof options.responseType !== 'undefined') {
+      let idTokenType;
```
hiranya911, line 803: Instead of a giant loop with many conditionals, how about this?

```typescript
Object.keys(options.responseType).forEach((key) => {
  if (!validResponseTypes.includes(key)) {
    // throw invalid parameter
  }
});

const idToken = options.responseType.idToken;
if (typeof idToken !== 'undefined') {
  if (!validator.isBoolean(idToken)) {
    // throw idToken must be a boolean
  }
}

const code = options.responseType.code;
if (typeof code !== 'undefined') {
  if (!validator.isBoolean(code)) {
    // throw code must be a boolean
  }
}

const allKeys = Object.keys(options.responseType).length;
const enabledCount = Object.values(options.responseType).filter(Boolean).length;
if (allKeys > 0 && enabledCount != 1) {
   // throw exactly one must be set
}
```

This allows following:
```
{}, {idToken: true}, {code: true}, {idToken: true, code: false}, {idToken: false, code: true}
```

Doesn't allow:
```
{idToken: false}, {code: false}, {idToken: true, code: true}, {idToken: false, code: false}
```

test/unit/auth/auth-config.spec.ts
```diff
@@ -943,32 +944,52 @@ describe('OIDCConfig', () => {
...
       const validRequest = deepCopy(clientRequest) as any;
       validRequest.responseType.code = false;
       validRequest.responseType.idToken = true;
       expect(() => OIDCConfig.validate(validRequest, true)).not.to.throw();
     });
 
+    it('should not throw when only idToken responseType is set to true', () => {
+      const validRequest = deepCopy(clientRequest) as any;
+      const validResponseType = { idToken: true };
+      validRequest.responseType = validResponseType;
```
hiranya911, line 962: Nit: Can directly assign the literal without the local `const`.

```
validRequest.resposneType = {idToken: true};
```

hiranya911 (changes requested):
Thanks for making the changes. Looks pretty good. Just one last suggestion to cleanup the large loop in the implementation.

src/auth/auth-config.ts
xil222: I think {idToken: false}, {code: false} could be allowed if the request doesn't change the original responseType. In this API, since we're only verifying the correctness of options.responseType itself, we can't see whether we allow these without knowing about original OIDC config responseType.

hiranya911: That makes sense. I'd suggest relaxing the check to allow those 2 cases:

```
if (allKeys > 1 && enabledCount != 1) {
   // throw exactly one must be set
}
```

xil222: sounds good

hiranya911 (approved):
Thanks. LGTM 👍

src/auth/auth-config.ts
```diff
@@ -773,6 +792,61 @@ export class OIDCConfig implements OIDCAuthProviderConfig {
         '"OIDCAuthProviderConfig.displayName" must be a valid string.',
       );
     }
+    if (typeof options.clientSecret !== 'undefined' &&
```
lsirac, line 795: why can't you just do if (!options.clientSecret)?

src/auth/auth-config.ts
```diff
@@ -806,6 +880,17 @@ export class OIDCConfig implements OIDCAuthProviderConfig {
     // When enabled is undefined, it takes its default value of false.
     this.enabled = !!response.enabled;
     this.displayName = response.displayName;
+
+    if (typeof response.clientSecret !== 'undefined') {
+      this.clientSecret = response.clientSecret;
+    }
+    // If we do not set responseType, we have idToken flow
```
lsirac, line 887: we don't need to do this - the backend will default to the ID token flow

xil222: Yes, I think line 892 is the case setting to ID token flow by default, this section I am fetching the secret(for need if this is code flow).

src/auth/auth-config.ts
xil222: I think (!options.clientSecret) is also correct.
I was just trying to follow the code style, I guess the reason here is that (!options.clientSecret) return true for clientSecret as empty string, null, false, {}, etc, while has overlap functionality with !validator.isNonEmptyString(options.clientSecret), but we only accept string type here, so here first check on whether clientSecret is defined could suffice.

src/auth/auth-config.ts
lsirac: The server response should already have a responseType set to ID token so we wouldn't have to set it here IIRC

xil222: Thanks for explaining. I experimented what you said in integration test, by calling createProviderConfig or getProviderConfig triggering RESTAPI with OIDCConfig(if doesnt include responseType, the config I get doesn't include responseType as well).
Therefore, I think it is necessary to include the logic (auto-set idToken flow)

lsirac: The backend definitely returns this - it must be not getting set somewhere else. Here's a sample response when I make the call myself:

{
  "name": "projects/projectId/oauthIdpConfigs/oidc.testing-1234",
  "clientId": "CLIENT_ID",
  "issuer": "https://google.com",
  "displayName": "displayName",
  "responseType": {
    "idToken": true
  }
}

Might be worth looking at the backend response in your integration test, and see why it's not part of the response here

xil222: I thought the original integration test covered the case you mentioned, and it didn't return an error. Thanks for correction.

lsirac (approved):

src/auth/auth-config.ts
```diff
@@ -816,6 +897,8 @@ export class OIDCConfig implements OIDCAuthProviderConfig {
       providerId: this.providerId,
       issuer: this.issuer,
       clientId: this.clientId,
+      clientSecret: this.clientSecret,
+      responseType: this.responseType,
```
bojeil-google, line 901: I would copy the content here so developers cannot modify the content.
`responseType:  deepCopy(this.responseType),`

test/unit/auth/auth-config.spec.ts
```diff
@@ -769,6 +783,17 @@ describe('OIDCConfig', () => {
       expect(config.enabled).to.be.true;
     });
 
+    it('should set readonly property clientSecret', () => {
```
bojeil-google, line 786: Can you add a test for legacy flow (when we didn't expose responseType)?
Basically, what happens if someone was enabling an OIDC provider (without any response type) and upgrades to a new version of the Admin SDK. Will the old code continue to work without any new error?

test/unit/auth/auth-config.spec.ts
```diff
@@ -727,20 +727,34 @@ describe('OIDCConfig', () => {
     issuer: 'https://oidc.com/issuer',
     displayName: 'oidcProviderName',
     enabled: true,
+    clientSecret: 'CLIENT_SECRET',
```
bojeil-google, line 730: It may be a good idea to add one test in `auth-api-request.spec.ts` for code flow:
https://github.com/firebase/firebase-admin-node/blob/011c53043896587e2d41228b3c8fdc463faff56a/test/unit/auth/auth-api-request.spec.ts#L3496

bojeil-google (changes requested):

src/auth/auth-config.ts
```diff
@@ -773,6 +792,61 @@ export class OIDCConfig implements OIDCAuthProviderConfig {
...
+          throw new FirebaseAuthError(
+            AuthClientErrorCode.INVALID_ARGUMENT,
+            '"OIDCAuthProviderConfig.responseType.idToken" must be a boolean.',
+          );
+        }
+      }
+      
+      const code = options.responseType.code;
+      if (typeof code !== 'undefined') {
+        if (!validator.isBoolean(code)) {
```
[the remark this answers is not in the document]
xil222: There're 2 sub cases in the if statement, which I think its better not to group them.

src/auth/auth-config.ts
```diff
@@ -829,7 +827,7 @@ export class OIDCConfig implements OIDCAuthProviderConfig {
         }
         
         // If code flow is enabled, client secret must be provided.
-        if (typeof options.clientSecret === 'undefined') {
+        if (code === true && typeof options.clientSecret === 'undefined') {
```
bojeil-google, line 830: `if (code && typeof options.clientSecret === 'undefined') {`

test/unit/auth/auth-api-request.spec.ts
```diff
@@ -3503,12 +3503,20 @@ AUTH_REQUEST_HANDLER_TESTS.forEach((handler) => {
         enabled: true,
         clientId: 'CLIENT_ID',
         issuer: 'https://oidc.com/issuer',
+        clientSecret: 'CLIENT_SECRET',
```
bojeil-google, line 3506: I am concerned about legacy customers and would recommend keeping a test without `clientSecret` and `responseType`. Can you add new tests instead of modifying the existing ones?

bojeil-google (changes requested):
Change looks good. Just 2 minor issues.

test/unit/auth/auth-api-request.spec.ts
xil222: I add tests for createOAuthIdpConfig and updateOAuthIdpConfig

bojeil-google (approved):

egilmorez (approved):
One optional nit if it's not too big a PITA to regen the reference.  

Thanks!

Thaina:
Is this feature would allow us to use any OpenID provider even it was not supported by firebase console directly?

Can we just create config for LINE or LoginWithAmazon to connect firebase from those account provider as of now?

Issue #1233, opened 2021-04-21
[BUG] Firebase auth triggers timeout


bolds07:
### [REQUIRED] Step 2: Describe your environment

  * Operating System version: GCF
  * Firebase SDK version:  9.4.2 [already tried 9.6.0 and it has exact same proble]
  * Firebase Product: auth (auth, database, storage, etc)
  * Node.js version: 10
  * NPM version: I've no idea what version google uses

### [REQUIRED] Step 3: Describe the problem
I’ve been working on an Android project for over 3 years now. Since beginning my project uses Firebase for many purposes including authentication.
Nowadays has grown a lot and I got a few thousand users per day and many hundreds users per hour. With this expansion I started noticing some errors that weren’t before [or weren’t that noticiable]
I use a firebase function to authenticate my users. The function is simple it validates a request cryptographic signature (to guarantee that it came from my app and not from someone bruteforcing it), then extract the parameters and query the firebase database for that credentials (only 1 result/or no result, will always been returned) if everything is okay a firebase authentication token is created and returned to user.
I’m using GCF with node 10 runtime, 40 seconds timeout and 128MB memory limit

```
const admin = require('firebase-admin');
admin.initializeApp({
});

const auth = admin.auth();
const igDatabase = admin.database();
const ServerValue = admin.database.ServerValue;

const crypto = require("crypto");
 exports.auth = (req, res) => {
    if (validateReqParams(req.body)) {

        if (validateCredentials(req.body)) {
	//here I extract the params from the json body
            authenticate(uid, newUser, regDate, res);
        } else
            invalidRequest('invalid body.d', req.body.d, res);

    } else
        invalidRequest('body.s doesn\'t match', req.body, res);

};


function validateReqParams(body) {

    return body && body.sign && body.data && body.sign == crypto.createHmac('sha256', 'my key').update(body.data).digest('hex');

}

function invalidRequest(error, item, res) {
    console.error(error);
    console.log(JSON.stringify(item));
    res.end();
}

function authenticate(uid, newUser, regDate, res) {
    auth.createCustomToken(uid).then(function(customToken) {

        res.status(200).json({
            t: customToken,
            n: newUser,
            r: regDate
        });
    }).catch(function(error) {
        console.error("Error creating custom token: " + JSON.stringify(error));
        res.status(200).json({
            e: error
        });
    });

}

```

[...]

PR #1234, opened 2021-04-21
fix(rtdb): Fixing a token refresh livelock in Cloud Functions

### LINKED ISSUE #1231: Realtime db triggers timeout after upgrading admin sdk to 9.6.0
### [REQUIRED] Describe your environment

  * Operating System version: GCF
  * Firebase SDK version: 8.4.1
  * Firebase Product: database, functions
  * Node.js version: 10
  * NPM version: 7.7.6

### [REQUIRED] Describe the problem

After upgrading to firebase-admin 9.6.0 realtime database triggers timeout periodically. Downgrade to 9.5.0 solved the issues.
 
#### Steps to reproduce:

Write a realtime db trigger. Use root reference to realtime db provided by the trigger (change.after.ref.root) to access data (read or write). Do not initialize firebase admin (this maybe irrelevant). The function will work correctly for a number of invocations. Eventually, one of the invocations will lead to the function timing out with whatever timeout value is configured for that specific function (60 seconds by default).

#### Relevant Code:

The code is proprietary, but even the simplest code that reads data from a different location than triggered will exhibit this behavior.
```
return change.after.ref.root
        .child(`some/db/path`)
        .once('value').then(() => return true) // this will never resolve or fail
```
I wish I could attach a screenshot of the functions Health tab showing spike in latencies after the update. Reverting to 9.5.0 eliminated timeouts.

hiranya911:
Resolves #1231 and #1233

We originally assumed that a credential would never issue tokens with a TTL shorter than 5 minutes. But turns out this assumption does not hold in the latest GCF runtime, where the metadata server can issue tokens with TTLs as short as 2 minutes. In such situations the SDK ends up in a livelock resulting in high CPU/memory usage, and other normal operations timing out.

Livelock event sequence:

1. Attempt to fetch a new token: the metadata server sends a token with a TTL less than 5 minutes.
2. Fire token change listeners.
3. RTDB token change listener calls `FirebaseApp.INTERNAL.getToken()` to access the last fetched token.
4. But the last fetched token has a TTL less than 5 minutes (i.e. eligible for refresh). Back to step 1.

In order to break the above cycle, I'm introducing a new `getCachedToken()` method to `FirebaseAppInternals`. This method simply returns the last fetched/cached token without ever attempting to refresh it. In step 3, we call this method instead of `getToken()`.

We already had implemented a test case for this scenario but had skipped it in the past. We can enable it now.

### DIFF
--- a/src/database/database-internal.ts
+++ b/src/database/database-internal.ts
@@ -116,18 +116,15 @@ export class DatabaseService {
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   private onTokenChange(_: string): void {
-    this.appInternal.INTERNAL.getToken()
-      .then((token) => {
-        const delayMillis = token.expirationTime - TOKEN_REFRESH_THRESHOLD_MILLIS - Date.now();
-        // If the new token is set to expire soon (unlikely), do nothing. Somebody will eventually
-        // notice and refresh the token, at which point this callback will fire again.
-        if (delayMillis > 0) {
-          this.scheduleTokenRefresh(delayMillis);
-        }
-      })
-      .catch((err) => {
-        console.error('Unexpected error while attempting to schedule a token refresh:', err);
-      });
+    const token = this.appInternal.INTERNAL.getCachedToken();
+    if (token) {
+      const delayMillis = token.expirationTime - TOKEN_REFRESH_THRESHOLD_MILLIS - Date.now();
+      // If the new token is set to expire soon (unlikely), do nothing. Somebody will eventually
+      // notice and refresh the token, at which point this callback will fire again.
+      if (delayMillis > 0) {
+        this.scheduleTokenRefresh(delayMillis);
+      }
+    }
   }
 
   private scheduleTokenRefresh(delayMillis: number): void {
--- a/src/firebase-app.ts
+++ b/src/firebase-app.ts
@@ -74,6 +74,10 @@ export class FirebaseAppInternals {
     return Promise.resolve(this.cachedToken_);
   }
 
+  public getCachedToken(): FirebaseAccessToken | null {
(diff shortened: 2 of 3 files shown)


schmidt-sebastian (commented):
Is it possible to use `forceRefresh` in getToken() and set it to false? You can still use shouldRefresh() when `forceRefresh` is undefined.

hiranya911:
I think the existing semantics are:

* `forceRefresh = true`: Force a token refresh
* `forceRefresh = false` (same as unspecified): Do not _force_ a token refresh. The implementation may still refresh the token if necessary. 

I don't think changing the latter to mean "do not refresh the token under any circumstances" is exactly correct.

Plus I like the new API `getCachedToken()` is synchronous. Makes it very obvious that no RPCs are taking place.

schmidt-sebastian (approved):
The token management is getting a bit hard to understand, but LGTM. If we want to, we can probably reduce the surface quite a bit by just passing the token to `onTokenChange`, but that likely has other complications.

hiranya911:
Yes, if we could pass both token string and expiry time to the `onTokenChange` event callback, this would be so much easier. But it appears the web SDK actually expects a certain signature for these token change events. `@firebase/database` in particular registers a listener using the same API, and listens to token change events. So either we need to update the web SDK code, or implement some kind of adapter code in the Admin SDK. Neither options are exactly simple to pull off.

hiranya911:
Tested this out in GCF using the same function I used to repro the problem. With v9.6.0 the livelock usually occurs in about 25-30 minutes since the first invocation. With the fix I managed to run the function for 3+ hours with no timeouts (over 1600 invocations in that time). Memory usage stable in 60-70MB range the whole time. I think we are in a good position to merge this fix.

PR #1236, opened 2021-04-27
fix: Removed App from the top-level admin namespace

hiranya911:
`App` is exposed via the `admin.app` namespace. Not necessary to expose it on the `admin` namespace too.


lahirumaramba (approved):
LGTM!

Issue #1202, opened 2021-03-19
auth: generatePasswordResetLink returns auth/internal-error on valid non-registered email address


loukotal:
### Describe your environment

  * Operating System version: macOS 10.15.7
  * Firebase SDK version: 9.4.2
  * Firebase Product: auth
  * Node.js version: 14.16.0
  * NPM version: 6.14.11

### Describe the problem
Calling `generatePasswordResetLink` with a valid email that is "not registered" (or, in other words, is not tied to any user), results in `auth/internal-error`. We used to catch `auth/user-not-found` error code for cases where the email was not found, but it seems like the error code changed to `EMAIL_NOT_FOUND`. `EMAIL_NOT_FOUND` is not handled/mapped and therefore the response results in `auth/internal-error`. The error code is not present in the docs [here](https://firebase.google.com/docs/auth/admin/errors), is it not handled on purpose then?
I would have thought an error code like this is something that could have its own error code and not default to `auth/internal-error`, hence this issue.

Or is to something to be reported on the official Firebase bug report site [here](https://firebase.google.com/support/troubleshooter/report/bugs)? If so, then I am sorry for opening the issue here and I will report it on the Firebase website. 

Relevant pieces of code I could (or could not) find:
- query for `EMAIL_NOT_FOUND` response code: https://github.com/firebase/firebase-admin-node/search?q=EMAIL_NOT_FOUND
- default value if response code is not present: https://github.com/firebase/firebase-admin-node/blob/1862342636e816b61337262254cdfe0cc410c640/src/utils/error.ts#L157
- AuthClientErrorCode definition: https://github.com/firebase/firebase-admin-node/blob/1862342636e816b61337262254cdfe0cc410c640/src/utils/error.ts#L344
-  Auth server to client enum error codes: https://github.com/firebase/firebase-admin-node/blob/1862342636e816b61337262254cdfe0cc410c640/src/utils/error.ts#L830

I also went through the python sdk and it seems like `EMAIL_NOT_FOUND` is also not handled. Maybe it is on purpose then?

#### Steps to reproduce:
1. Initialize firebase-admin app with a service account
2. Call `generatePasswordResetLink` with a valid email that is not tied to any user
3. Catch error

Or curl command to REST API for quick testing:
```bash
curl 'https://identitytoolkit.googleapis.com/v1/accounts:sendOobCode?key=API_KEY' \
-H 'Content-Type: application/json' \
[...]

PR #1246, opened 2021-05-03
fix(auth): adds missing EMAIL_NOT_FOUND error code

bojeil-google:
Catch `EMAIL_NOT_FOUND` and translate to `auth/email-not-found` when `/accounts:sendOobCode` is called for password reset on a user that does not exist.
Fixes https://github.com/firebase/firebase-admin-node/issues/1202


hiranya911 (approved):
LGTM 👍

PR #1247, opened 2021-05-03
chore: Added docstrings to type aliases in namespaces

hiranya911:
Added docstrings referencing the source of truth definitions in new module entrypoints. For example, the `admin.auth.UserRecord` type alias gets a docstring with a reference to the `UserRecord` export in the `firebase-admin/auth` module entry point. These references are correctly converted into hyperlinkg by API Documenter.

Screen Shot 2021-05-03 at 2 08 32 PM


lahirumaramba (approved):
LGTM!

PR #1249, opened 2021-05-05
feat: Added ESM entry points to the package

hiranya911:
Up until now we have only exposed CJS entry points from our modules. This means native ESM users weren't able to use our named exports:

```js
// This doesn't work today:
import { initializeApp } from 'firebase-admin/app';

// Instead ESM users must do this:
import cjsSugar from 'firebase-admin/app';
```

With this PR we are adding native ESM entry points to the package. This enables using named imports for ESM users.

## Methodology

We are essentially wrapping our existing CJS entry points as ESM entry points. The new script `generate-esm-wrapper.js` iterates through each of the CJS entry points, and generates ESM wrappers for them, which are stored at the `lib/esm` subdirectory. Here's the auto-generated ESM wrapper for the `firebase-admin/app` package as an example:

```js
import mod from "../../app/index.js";

export const SDK_VERSION = mod.SDK_VERSION;
export const applicationDefault = mod.applicationDefault;
export const cert = mod.cert;
export const deleteApp = mod.deleteApp;
export const getApp = mod.getApp;
export const getApps = mod.getApps;
export const initializeApp = mod.initializeApp;
export const refreshToken = mod.refreshToken;
```

Then we define a conditional exports for all the modules in the `package.json`:

```
"./app": {
      "require": "./lib/app/index.js",
      "import": "./lib/esm/app/index.js"
},
```

This enables Node.js to pick the correct entry point depending on whether it's running in CJS mode or the ESM mode.

## Testing

I have added a new package verification check that tests the modules in the native ESM mode. It's tied to the existing `verify_package` script that runs as part of our staging build.

```
> ./node_modules/.bin/mocha esm/*.js


  ESM entry points
    ✓ Should return an initialized App
    ✓ Should return an Auth client
    ✓ Should return a Messaging client
    ✓ Should return a ProjectManagement client
    ✓ Should return a SecurityRules client
    ✓ Should return a Database client
    ✓ Should return a Database client for URL
    ✓ Should return a Database ServerValue
    ✓ Should return a Cloud Storage client (61ms)
    ✓ Should return a Firestore client
    ✓ Should return a Firestore FieldValue
    ✓ Should return a DocumentReference
    ✓ Should return an InstanceId client
    ✓ Should return a MachineLearning client
    ✓ Should return a RemoteConfig client


  15 passing (125ms)
```


lahirumaramba (approved):
This is great! Thanks, Hiranya!

PR #1257, opened 2021-05-07
Fix storage emulator env formatting

abeisgoat:
The Firebase style env var for storage emulator can't have protocol and the Cloud version *must* have a protocol.

### DIFF
--- a/test/unit/storage/storage.spec.ts
+++ b/test/unit/storage/storage.spec.ts
@@ -127,7 +127,7 @@ describe('Storage', () => {
       expect(process.env.STORAGE_EMULATOR_HOST).to.equal(`http://${EMULATOR_HOST}`);
     });
 
-    it('sets STORAGE_EMULATOR_HOST if FIREBASE_STORAGE_EMULATOR_HOST is set', () => {
+    it('throws if FIREBASE_STORAGE_EMULATOR_HOST has a protocol', () => {
       expect(() => new Storage(mockApp)).to.throw;
     });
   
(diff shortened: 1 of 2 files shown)


test/unit/storage/storage.spec.ts
```diff
@@ -115,17 +115,20 @@ describe('Storage', () => {
...
 
     it('sets STORAGE_EMULATOR_HOST if FIREBASE_STORAGE_EMULATOR_HOST is set', () => {
-      new Storage(mockApp);
-      
-      expect(process.env.STORAGE_EMULATOR_HOST).to.equal(EMULATOR_HOST);
+      new Storage(mockApp)
+      expect(process.env.STORAGE_EMULATOR_HOST).to.equal(`http://${EMULATOR_HOST}`);
+    });
+
+    it('throws if FIREBASE_STORAGE_EMULATOR_HOST has a protocol', () => {
```
hiranya911, line 130: This test is currently passing by accident. There's nowhere in this test that we have set an env variable with the `http://` prefix. If you update the assertion as follows the error surfaces immediately:

```js
expect(() => new Storage(mockApp)).to.throw(
  'FIREBASE_STORAGE_EMULATOR_HOST should not contain a protocol');
```

hiranya911 (changes requested):
Change looks good. Need a minor fix in the test case.

abeisgoat:
All fixed, feel free to approve and merge!

hiranya911 (approved):

lahirumaramba (approved):
LGTM! Thank you, @abeisgoat !

Issue #1121, opened 2020-12-28
Conflicting Types for @types/node when using node 14


DrCheddar:
**Thank you for submitting your issue. We are operating at reduced capacity from Dec 18 2020 to Jan 4 2021. Please expect delayed responses. For more urgent requests please reach us via our  support channels https://firebase.google.com/support**

### [READ] Step 1: Are you in the right place?

  * For issues related to __the code in this repository__ file a Github issue.
  * If the issue pertains to Cloud Firestore, read the instructions in the "Firestore issue"
    template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Operating System version: Windows
  * Firebase SDK version: firebase-admin@9.4.2
  * Firebase Product: firebase-admin
  * Node.js version: 14.2.0
  * NPM version: 6.14.4

### [REQUIRED] Step 3: Describe the problem

When using node 14 with typescript's @types/node@14.14.16, attempting to compile with tsc gives a conflicting definition of 'node' error: 
![image](https://user-images.githubusercontent.com/46302202/103184160-b5540f00-4884-11eb-87c2-dc1a5c9a256e.png)

I need to use node 14 because something in fs-extra is broken in node 12.

#### Steps to reproduce:

1. Upgraded node version to node 14
2. Upgraded node types to @types/node@latest to fix fs-extra error in node 12
3. Ran tsc - got the conflicting node error

#### Relevant Code:

```tsconfig.json
  "dependencies": {
    "@types/node": "^14.14.16",
...
    "firebase-admin": "^9.4.2",

```

DrCheddar:
for some reason if I downgrade firebase-admin to 9.0.0 it seems to work fine

hiranya911:
> for some reason if I downgrade firebase-admin to 9.0.0 it seems to work fine

That's strange. Both v9.0.0 and v9.4.0 depends on same version of `@types/node`. 

https://github.com/firebase/firebase-admin-node/blob/v9.0.0/package.json#L58
https://github.com/firebase/firebase-admin-node/blob/v9.4.0/package.json#L61

[...]

PR #1258, opened 2021-05-08
Fix @types/node conflict with grpc and port type

nikhilag:
Upgraded the `@types/node` dependency to v12.12.47


lahirumaramba:
Removed the issue from the description to prevent #1121 from closing. Please see the comment by @hiranya911.
Thanks!

PR #1263, opened 2021-05-10
chore: Update FAC to beta

lahirumaramba:
- Update to beta
- Add integration test for verify token api

### DIFF
--- a/src/app-check/app-check-api-client-internal.ts
+++ b/src/app-check/app-check-api-client-internal.ts
@@ -28,7 +28,7 @@ import * as validator from '../utils/validator';
 import AppCheckToken = appCheck.AppCheckToken;
 
 // App Check backend constants
-const FIREBASE_APP_CHECK_V1_API_URL_FORMAT = 'https://firebaseappcheck.googleapis.com/v1alpha/projects/{projectId}/apps/{appId}:exchangeCustomToken';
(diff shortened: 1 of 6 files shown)


src/app-check/app-check-api-client-internal.ts
```diff
@@ -147,9 +147,9 @@ export class AppCheckApiClient {
    */
   private toAppCheckToken(resp: HttpResponse): AppCheckToken {
     const token = resp.data.attestationToken;
-    // `timeToLive` is a string with the suffix "s" preceded by the number of seconds,
+    // `ttl` is a string with the suffix "s" preceded by the number of seconds,
```
weixifan, line 150: Interesting -- unrelated to this review, the `ttl` is actually a textual representation of a `Duration` proto as per [AIP-142](https://google.aip.dev/142#durations) and [the `Duration` documentation](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/duration.proto). The only downside is that parsing it via `Duration` might cost us an additional dependency.

weixifan:
LGTM!

src/app-check/app-check-api-client-internal.ts
lahirumaramba: Yeah. I think for Javascript (Node.js) we can parse the JSON by referring to the JSON Mapping section in the proto. Can confirm that the JS SDK follows the same approach.

weixifan (approved):

PR #1268, opened 2021-05-10
Fix types for Create / Update MultifactorRequest

### LINKED ISSUE #1267: Incorrect typing on `createUser` api in admin.auth leads to broken multi-factor auth
* Operating System version: Mac OS 11.2.3
  * Firebase SDK version: 9.7.0
  * Firebase Product: auth
  * Node.js version: 12.21.0
  * NPM version: 7.7.0

Typescript types are incorrect for `CreateMultiFactorInfoRequest` and `CreatePhoneMultiFactorInfoRequest` in `admin.auth.createUser`. I expect to be able to pass in a phone number but am not allowed by the type system. Instead, I have to cast the type as `any`.

#### Steps to reproduce:

```
admin.auth.createUser({
  displayName: 'somebody',
  email: 'some@one.com'
  emailVerified: true,
  multiFactor: { enrolledFactors: [{ factorId: 'phone', phoneNumber: request.phoneNumber }] }
}
```

This yields the error 
  
image

You can work around this by casting the object to an `any`, but the type system is still wrong here.

This was also raised in https://github.com/firebase/firebase-admin-node/issues/1048 but for some reason that ticket was closed by the OP.

As a side-note, it feels like the code was written by someone who thought that extending a typescript `interface` would mean that instances of the derived class can be passed in instead of the base class, but with structural typing, this is not allowed.

jakebiesinger-storyhealth:
Fixes #1267.

Derived classes are not interchangeable with Base classes as types. Rather than making the request take a generic `<T extends CreateMultifactorInfoRequest>`, this assigns a union of the (one currently allowed) type.

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/src/auth/index.ts
+++ b/src/auth/index.ts
@@ -339,7 +339,7 @@ export namespace auth {
    * Interface representing common properties of a user enrolled second factor
    * for an `UpdateRequest`.
    */
-  export interface UpdateMultiFactorInfoRequest {
+  export interface BaseUpdateMultiFactorInfoRequest {
 
     /**
      * The ID of the enrolled second factor. This ID is unique to the user. When not provided,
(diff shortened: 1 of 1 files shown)


jakebiesinger-storyhealth:
@googlebot I signed it!

hiranya911:
Superseded by #1294

PR #1272, opened 2021-05-14
chore: Implementing API Documenter workflow for generating API references

hiranya911:
This PR implements an API Documenter workflow to replace the old Typedoc workflow. The new workflow is triggered using the same command as before:

```
npm run apidocs
```

This runs the following sequence:
1. builds the source
2. generates API Extractor model files in the `temp/` directory
3. generates markdown docs using the API documenter
4. runs a post-processing script to add extra RTDB and Firestore content to the generated pages

This doesn't generate a toc.yaml file at the moment, but we are working with the JS SDK team to get that feature implemented at the API Documenter level.

The post-processing script is very simple. It looks at the `extras/` directory for additional markdown content sources, and adds/merges those content to the top of same named files in the output. For example:

* Content in `extras/firebase-admin.firestore.md` is added to the top of `markdown/firebase-admin.firestore.md` file.
* Content in `extras/firebase-admin.database.md` is added to the top of `markdown/firebase-admin.database.md` file.

This provides a simple way to augment the generated markdown files with some hand-written content. Particularly useful in places where we want to link to external sites (JS SDK docs, Firestore docs etc) for more context.


docgen/extras/firebase-admin.database.md
```diff
@@ -0,0 +1,12 @@
+## External API Re-exports
+
+Following externally defined APIs are re-exported from this module entry point for convenience.
```
egilmorez, line 3: Suggest "The following" here and for firestore.md.

egilmorez (approved):
This looks great; really like the "extras" functionality.  One nit. Thanks!

docgen/extras/firebase-admin.database.md
hiranya911: Done

hiranya911:
Some content generated using this workflow can be found at: https://github.com/firebase/firebase-admin-node/blob/modular-sdk-docs/docgen/markdown/index.md

Legacy API namespace: https://github.com/firebase/firebase-admin-node/blob/modular-sdk-docs/docgen/markdown/firebase-admin.md

Sample modular API (`app` module): https://github.com/firebase/firebase-admin-node/blob/modular-sdk-docs/docgen/markdown/firebase-admin.app.md

Augment docs with extras (`firestore` module):  https://github.com/firebase/firebase-admin-node/blob/modular-sdk-docs/docgen/markdown/firebase-admin.firestore.md

lahirumaramba (approved):

PR #1273, opened 2021-05-17
chore: Added post processing script to fix doc titles and links

hiranya911:
* Updated dotted module names like `firebase-admin.auth` to the correct `firebase-admin/auth` form.
* Replaces the work "package" with "module" for accuracy.

Preview home page: https://github.com/firebase/firebase-admin-node/blob/modular-sdk-docs/docgen/markdown/index.md

### DIFF
--- a/docgen/post-process.js
+++ b/docgen/post-process.js
@@ -20,9 +20,75 @@ const path = require('path');
 const readline = require('readline');
 
 async function main() {
+  await applyExtras();
+  await fixHomePage();
+  await fixTitles();
+}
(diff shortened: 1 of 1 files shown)


egilmorez (approved):
LGTM!

In passing, I notice that we are missing descriptions for the modules and for some lower-level items.  We should make sure all of those are in place (separately) so that our end product does not have lots of empty table cells.

lahirumaramba (approved):
Thank you, Hiranya!

PR #1274, opened 2021-05-18
Add support for a service account credential file at the GCLOUD_CREDENTIAL_PATH

flagbug:
I noticed that the SDK would fail to detect the Application Default Credentials when I set the `GCLOUD_CREDENTIAL_PATH` environment variable the path of a service account key file. Since the same is working perfectly fine in the C# SDK, I assumed that this is a bug.

Please let me know if I should do anything different here or if this is the expected behavior, since I'm not familiar with the internals of the Node SDK at all!

### DIFF
--- a/src/credential/credential-internal.ts
+++ b/src/credential/credential-internal.ts
@@ -354,9 +354,13 @@ export function getApplicationDefault(httpAgent?: Agent): Credential {
 
   // It is OK to not have this file. If it is present, it must be valid.
   if (GCLOUD_CREDENTIAL_PATH) {
-    const refreshToken = readCredentialFile(GCLOUD_CREDENTIAL_PATH, true);
-    if (refreshToken) {
-      return new RefreshTokenCredential(refreshToken, httpAgent, true);
+    const credentialFile = readCredentialFile(GCLOUD_CREDENTIAL_PATH, true);
+    if (credentialFile) {
+      if (credentialFile.type === 'service_account') {
+        return new ServiceAccountCredential(credentialFile, httpAgent, true);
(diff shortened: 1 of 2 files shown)


lahirumaramba:
Hi @flagbug !
To my understanding the service account key file path should be set using `GOOGLE_APPLICATION_CREDENTIALS` environment variable. See https://firebase.devsite.corp.google.com/docs/admin/setup#initialize-sdk

Is that what you were looking for?

flagbug:
@lahirumaramba Looks like this is a link to an internal Google tool which I can access unfortunately. If I'm understanding you correctly, you're saying that `GCLOUD_CREDENTIAL_PATH` is supposed to be used internally by the Google Cloud SDK and as a developer, I should always prefer to set the `GOOGLE_APPLICATION_CREDENTIALS` path, correct?

lahirumaramba:
> @lahirumaramba Looks like this is a link to an internal Google tool which I can access unfortunately. If I'm understanding you correctly, you're saying that `GCLOUD_CREDENTIAL_PATH` is supposed to be used internally by the Google Cloud SDK and as a developer, I should always prefer to set the `GOOGLE_APPLICATION_CREDENTIALS` path, correct?

Apologies for the incorrect link! Just updated the original comment with https://firebase.google.com/docs/admin/setup#initialize-sdk

You’re right. You can use the GOOGLE_APPLICATION_CREDENTIALS environment variable to point to a service account file to use application default credentials at initialization. I might have misunderstood your requirements though so give it a try and let us know if this works for your project. Thanks!

Issue #1276, opened 2021-05-18
The second factor "uid" must be a valid non-empty string


flyingluscas:
### Describe your environment

  * Operating System version: Arch Linux
  * Firebase SDK version: 9.5.0
  * Firebase Product: Auth
  * Node.js version: 14.15
  * NPM version: 6.14.11

### Describe the problem

The `admin.auth().updateUser` function does not allow me to add a second factor to an existing user, always triggering the error `The second factor "uid" must be a valid non-empty string.`, the function expects a valid uid but [according to the docs it should not, if no uid is informed it should auto generate one](https://cloud.google.com/identity-platform/docs/admin/manage-mfa-users#updating_a_user). Can anyone help me understand if this is a bug in the lib or am I doing something wrong?

#### Steps to reproduce:

1. Create a user without any secondary factors.
2. Try to add a new secondary factor using the [`.updateUser`](https://cloud.google.com/identity-platform/docs/admin/manage-mfa-users#updating_a_user) function.

#### Relevant Code:

```js
admin.auth().updateUser('123456789', {
  phoneNumber: '+16505550003',
  emailVerified: true,
  multiFactor: {
    enrolledFactors: [
      {
        phoneNumber: '+16505550003',
        displayName: 'Bender Bending Rodríguez',
        factorId: 'phone',
      },
    ],
  },
})
```

hiranya911:
Current implementation treats UID to be required for update operations:

https://github.com/firebase/firebase-admin-node/blob/e7155eacddd59cb449991476227c3a38681176cf/src/auth/auth-api-request.ts#L278

@bojeil-google is that a bug?

bojeil-google:
I think it is a bug in our code. I tried to send the request directly to confirm the backend API does not require the `uid` and it seems to work without it. It will provision a new `uid` on its own. We should remove that check `uidRequired`. I am also double checking this is the intended behavior with some backend folks.

bojeil-google:
The tests seem to be currently broken (only partially running). If you can help fix that, I am happy to review.

silviokucz:
As of 1/31/2023 I am still experiencing this issue.
My environment:

    Operating System version: Ubuntu Linux
    Firebase SDK version:11.19.0
    Firebase Product: Auth
    Node.js version: v14.21.2
    NPM version: 6.14.17
    
My exact call:
admin.auth().updateUser('yBap141NgrR6o4cvaVOAEQVdaGCl', {
  multiFactor: {
[...]

PR #1278, opened 2021-05-18
fix(auth): make MFA uid optional for updateUser operations

bojeil-google:
MFA `uid` should be optional for `updateUser` operations.
When not specified, the backend will provision a `uid` for the
enrolled second factor.

Fixes https://github.com/firebase/firebase-admin-node/issues/1276

### DIFF
--- a/src/auth/auth-api-request.ts
+++ b/src/auth/auth-api-request.ts
@@ -274,8 +274,8 @@ function validateAuthFactorInfo(request: AuthFactorInfo): void {
   // No enrollment ID is available for signupNewUser. Use another identifier.
   const authFactorInfoIdentifier =
       request.mfaEnrollmentId || request.phoneInfo || JSON.stringify(request);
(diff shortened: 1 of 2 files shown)


hiranya911 (approved):
Thanks. LGTM 👍

silviokucz:
As of 1/31/2023 I am still experiencing this issue.
My environment:

Operating System version: Ubuntu Linux
Firebase SDK version:11.19.0
Firebase Product: Auth
Node.js version: v14.21.2
NPM version: 6.14.17

My exact call:
admin.auth().updateUser('yBap141NgrR6o4cvaVOAEQVdaGCl', {
multiFactor: {
enrolledFactors: [
{
phoneNumber: '+14255555504',
displayName: 'MFA',
factorId: 'phone',
},
],
},
})

Seems to work when the user is already multifactor but fails otherwise.

PR #1285, opened 2021-05-20
chore: Remove gulp-replace dependency

hiranya911:
We no longer seem to use this dependency. Likely because we switched to reading the package version from the `package.json` file, as opposed to writing it to source files during build time.


lahirumaramba (approved):
Thanks, Hiranya! LGTM!

PR #1292, opened 2021-05-24
chore: Temporarily disabling sendToDeviceGroup integration test

hiranya911:
The backend server has implemented a behavior change that is causing this test to fail. Temporarily disabling the test as we work with the FCM team to understand the new behavior, and incorporate it into our implementation correctly.


lahirumaramba (approved):
Thank you!

PR #1293, opened 2021-05-25
feat(fac): Expose FAC APIs from firebase-admin/app-check entry point

lahirumaramba:
- Migrated FAC APIs to modular SDK
- Exposed FAC API types from the firebase-admin/app-check entry point.
- Added API report for new entry point.


src/app-check/app-check-api.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * The time-to-live duration of the token in milliseconds.
+   */
+  ttlMillis: number;
+}
+
+/**
+ * Interface representing a decoded Firebase App Check token, returned from the
+ * {@link appCheck.AppCheck.verifyToken `verifyToken()`} method.
```
hiranya911, line 35: This needs to follow the conventions described at https://tsdoc.org/pages/tags/link/

src/app-check/app-check-api.ts
```diff
@@ -0,0 +1,98 @@
...
+ */
+export interface DecodedAppCheckToken {
+  /**
+   * The issuer identifier for the issuer of the response.
+   *
+   * This value is a URL with the format
+   * `https://firebaseappcheck.googleapis.com/<PROJECT_NUMBER>`, where `<PROJECT_NUMBER>` is the
+   * same project number specified in the [`aud`](#aud) property.
```
hiranya911, line 43: Same here. Use the correct tsdoc syntax to refer to aud property.

You can try running `npm run apidocs` and inspect the markdown output to see what it looks like.

hiranya911 (changes requested):
Thanks. This looks great! Only had a couple of minor comments regarding the references in API ref docs.

src/app-check/app-check-api.ts
lahirumaramba: Thanks! Good catch. Used @link to fix these.

src/app-check/app-check-api.ts
lahirumaramba: Thanks! Fixed the links.

hiranya911 (approved):
LGTM 👍

PR #1294, opened 2021-05-25
fix(auth): Better type hierarchies for Auth API

### LINKED ISSUE #1267: Incorrect typing on `createUser` api in admin.auth leads to broken multi-factor auth

hiranya911:
Defining `AuthProviderConfig`, `CreateMultiFactorInfoRequest`, and `UpdateMultiFactorInfoRequest` as union types. 

I'd also like to the move the `factorId` field to the child interfaces, and further restrict its domain (`string` --> `'phone'`), but this seems to be a breaking change (a couple of compilation issues detected in tests). So for now I'm keeping this field as is. 

Resolves #1267


src/auth/index.ts
```diff
@@ -463,14 +469,20 @@ export namespace auth {
...
 
     /**
      * The phone number associated with a phone second factor.
      */
     phoneNumber: string;
   }
 
+  /**
+   * Type representing the properties of a user enrolled second factor
+   * for an `CreateRequest`.
```
bojeil-google, line 482: a `CreateRequest`.

bojeil-google (approved):

lahirumaramba (approved):
LGTM! Thanks!
Should we get the reference docs reviewed?

src/auth/index.ts
hiranya911: Done

hiranya911:
@egilmorez can you take a look at the comment updates in auth/index.ts?

src/auth/index.ts
```diff
@@ -367,14 +367,20 @@ export namespace auth {
...
+  export interface UpdatePhoneMultiFactorInfoRequest extends BaseUpdateMultiFactorInfoRequest {
 
     /**
      * The phone number associated with a phone second factor.
      */
     phoneNumber: string;
   }
 
+  /**
+   * Type representing the properties of a user enrolled second factor
```
egilmorez, line 379: This would be clearer as "user-enrolled" 

There are 7 other instances in the file; if that's a PITA, I can do it in a follow-on.

egilmorez (approved):
One possible nit, LG, thanks!

src/auth/index.ts
hiranya911: Fixed

PR #1300, opened 2021-05-26
Update supported Node version to 10.13.0v and up

### LINKED ISSUE #1295: P0 issue: accessing .auth() throwing error that util.promisify() was not receiving a function! breaks several of our applications

  * Operating System version: Linux (Heroku Dyno)
  * Firebase SDK version: ^9.0.0
  * Firebase Product: auth
  * Node.js version: 10.6.0
  * NPM version: 6.7.0

### [REQUIRED] Step 3: Describe the problem:
My Heroku Server randomly started crashing recently and wont boot back up (even though it works perfectly fine locally). If I comment out `const firebaseAdmin = firebaseApp.auth()` , then it deploys with no crash, but as soon as I comment it back in, it crashes. This is breaking our applications and so this has become a P0 issue for us. Any help would be greatly appreciated. I will post screenshots of the error and the line of code I commented out that fixed the error. Thank you!

#### Steps to reproduce:
Deploy a heroku server and import the firebase-admin sdk and call the `.auth()` method with the same node and npm env. This issue has arisen within the past week, I'm not sure exactly when it started.

#### Relevant Code:

Error:
Screen Shot 2021-05-25 at 7 20 00 PM

Fixed by commenting out:
Screen Shot 2021-05-25 at 7 20 48 PM

I've gone through and tested it several times over, and consistently this ^ is the line of code causing it to crash. 

Again this is NOT happening locally but for some reason only on a deployed heroku dyno instance. 

Any help on this would be SUPER appreciated because it is currently breaking several of our apps 

our `firebaseApp` firebase configuration has been working perfectly up until around last week and the error does not get thrown until `.auth()` is called so Im inclined to believe we have everything set up properly, unless something has changed recently about how it should all be setup configuration wise

lahirumaramba:
- Update Node.js version to >=10.13.0
- Resolves #1295

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -87,7 +87,7 @@ information on using pull requests.
 
 ### Prerequisites
 
-1. Node.js 10.10.0 or higher.
+1. Node.js 10.13.0 or higher.
 2. NPM 5 or higher (NPM 6 recommended).
 3. Google Cloud SDK ([`gcloud`](https://cloud.google.com/sdk/downloads) utility)
 
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ requests, code review feedback, and also pull requests.
 
 ## Supported Environments
 
-We support Node.js 10.10.0 and higher.
+We support Node.js 10.13.0 and higher.
 
 Please also note that the Admin SDK should only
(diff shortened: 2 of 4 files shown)


hiranya911 (approved):
LGTM 👍

PR #1314, opened 2021-06-01
chore: Remove unused dependency del

hiranya911:
Seems to be another unused dev dependency. Cannot find any direct imports of it anywhere.


hiranya911:
Never mind. Used in the gulpfile

PR #1317, opened 2021-06-02
fix: Applying extra markdown content before first header

hiranya911:
We currently apply the extra markdown content just after the `{% block body %}` tag. However, usually a package description text appears after this tag. This results in the following ordering:

```
{% block body %}

<extra content>

Package description text

# First heading
```

Ideally, the extra content should appear after the package description text. To obtain this behavior I'm changing the markdown post processing job to look for the `# First heading` in the file, and apply the extra content before it:

```
{% block body %}

Package description text

<extra content>

# First heading
```

Also making it possible to run the API Documenter portion of the docs pipeline by just invoking `npm run api-documenter`, and correcting some doc comments to meet the typedoc standard.


lahirumaramba (approved):
LGTM!

PR #1342, opened 2021-06-23
fix: Updated TOC for new Auth type aliases

hiranya911:
Adding the new `Base*` types introduced in #1294 to the TOC. The type aliases are documented in the `admin.auth` page, and do not require separate TOC entries.


lahirumaramba (approved):
Thanks Hiranya! LGTM

PR #1344, opened 2021-06-24
Add generated client files to fcm

maceonthompson:
[Draft]
The result of moving autogenerated fcm client code using [GAPIC](https://github.com/googleapis/gapic-generator) into the messaging directory. Subsequent PRs will be made into this branch as development continues (namely, migrating to the generated client over [messaging-api-request-internal.ts](https://github.com/firebase/firebase-admin-node/blob/6f73c8c803ed88e32c17c9c62a644074ed6ec342/src/messaging/messaging-api-request-internal.ts)

This is part of an ongoing internship project/experiment in the feasibility of using autogenerated clients, which will hopefully make supporting multiple apis across multiple languages easier in the long term.


src/messaging/src/index.ts
```diff
@@ -0,0 +1,25 @@
...
+//
+// ** This file is automatically generated by gapic-generator-typescript. **
+// ** https://github.com/googleapis/gapic-generator-typescript **
+// ** All changes to this file may be overwritten. **
+
+import * as v1 from './v1';
+const FcmServiceClient = v1.FcmServiceClient;
+type FcmServiceClient = v1.FcmServiceClient;
+export {v1, FcmServiceClient};
+export default {v1, FcmServiceClient};
```
hiranya911, line 23: Please remove.

src/messaging/src/index.ts
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+//
+// ** This file is automatically generated by gapic-generator-typescript. **
+// ** https://github.com/googleapis/gapic-generator-typescript **
+// ** All changes to this file may be overwritten. **
+
+import * as v1 from './v1';
+const FcmServiceClient = v1.FcmServiceClient;
+type FcmServiceClient = v1.FcmServiceClient;
+export {v1, FcmServiceClient};
```
hiranya911, line 22: We need a way to suppress these lint errors. We can either ask the REGAPIC folks to fix it so the generated code has no lint errors, or update our eslint configuration to avoid lint checking these files.

tsconfig.json
```diff
@@ -6,14 +6,17 @@
...
     "noImplicitThis": true,
     "alwaysStrict": true,
     "strictBindCallApply": true,
     "strictNullChecks": true,
     "strictFunctionTypes": true,
     //"strictPropertyInitialization": true,
-    "lib": ["es2018"],
+    "lib": [
+      "es2018",
+      "dom"],
```
hiranya911, line 19: closing bracket should be on a separate line.

src/messaging/src/v1/fcm_service_client.ts
```diff
@@ -0,0 +1,362 @@
+/* eslint-disable */
```
hiranya911, line 1: Let's put these in a directory named `generated` instead of `src`. The full path should be something like `src/messaging/generated/v1/fcm_service_client.ts`.

Also, how do you feel about keeping the generated sources and protos in completely separate directory hierarchy. Something like:

```
REPO_ROOT
  src/
    messaging/
  generated/
    messaging/
      src/
        v1/
      protos/
```

This will likely require some work at the tsconfig and our build config (gulpfile) to compile and bundle correctly.

hiranya911 (commented):
Looks pretty good. We need to figure out a way to resolve the lint errors. Plus some suggestions on the directory structure.

src/messaging/src/v1/fcm_service_client.ts
maceonthompson: I agree that it will probably be a bit of extra work but think it'll be clearer/easier to parse, so definitely worth it.

src/messaging/src/index.ts
hiranya911: Ignore this. Didn't realize this is part of the generated code.

tsconfig.json
maceonthompson: Fixed!

src/messaging/src/index.ts
maceonthompson: Updated eslint config for now.

src/messaging/src/v1/fcm_service_client.ts
hiranya911: Do you want to do that as part of this PR, or do that later? I think for our immediate purposes what you have here now is good enough.

maceonthompson: It's a quick/easy enough change, and I think it's worth not having to do a larger refactor down the line.

hiranya911:
I see that you've nested `generated` within the existing `src` directory, which is not quite what I had in mind. I was suggesting that we put `generated` at the root of the repo (same level as `src`). However, this is likely to complicate our build quite a bit (TypeScript will have to pull sources from 2 directories, and combine the compiled output into one).

So lets use the structure you currently have for now. Later we can discuss other approaches for organizing these files in our repo.

PR #1353, opened 2021-06-30
Update Gulpfile to copy files required for generated client to properly run

maceonthompson:
`npm run build` etc. will now copy over the the `.d.ts` and `.json` files found in `generated/messaging`. Before, this didn't happen, and the generated client would error on runtime.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -118,7 +118,6 @@ gulp.task('copyTypings', function() {
 
 gulp.task('copyJSON', function() {
   return gulp.src([
-    'src/**/protos/*.json',
     // This isn't ideal, but doing something like
     // 'src/generated/**/*.json' results in incorrect paths in the /lib dir
     'src/**/*.json'
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -101,7 +101,7 @@ gulp.task('compile', function() {
  */
 gulp.task('compile_test', function() {
   return gulp.src(paths.test)
-    // Compile Typescript ianto .js and .d.ts files
+    // Compile Typescript into .js and .d.ts files
     .pipe(buildTest())
 });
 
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -106,21 +106,36 @@ gulp.task('compile_test', function() {
...
     .pipe(gulp.dest(paths.build))
 });
 
-gulp.task('compile_all', gulp.series('compile', 'copyTypings', 'compile_test'));
+gulp.task('copyJSON', function() {
+  return gulp.src([
+    'src/**/protos/*.json',
+    // This isn't ideal, but doing something like
+    // 'src/generated/**/*.json' results in incorrect paths in the /lib dir
+    'src/**/*.json'
```
hiranya911, line 124: Can we say `src/generated/**/*.json` instead?

gulpfile.js
```diff
@@ -106,21 +106,36 @@ gulp.task('compile_test', function() {
...
+  ])
     // Add header
     .pipe(header(banner))
     .pipe(gulp.dest(paths.build))
 });
 
-gulp.task('compile_all', gulp.series('compile', 'copyTypings', 'compile_test'));
+gulp.task('copyJSON', function() {
+  return gulp.src([
+    'src/**/protos/*.json',
```
hiranya911, line 121: I think this is included in the next pattern, and therefore is probably not needed.

gulpfile.js
hiranya911: Ah I see. It ends up copying them into `lib/messaging` instead of `lib/generated/messaging`. I guess it only matches path segments after `src/generated` when the pattern is specified this way. Then I think the way you've implemented this is correct.

gulpfile.js
maceonthompson: Good point, updated.

gulpfile.js
maceonthompson: Yeah, doing the more general wildcard is less "comfortable" than doing the specific `src/generated/**/*.json`, but it's (as far as I could tell) the only way for the generated path to be structured properly.

hiranya911 (approved):
LGTM 👍

PR #1354, opened 2021-07-02
Fcm GAPIC type conversion

maceonthompson:
Add private method that converts from public `messaging.Message` type to the similar but generated `IMessage` type.


src/messaging/messaging.ts
```diff
@@ -41,6 +42,12 @@ import MessagingConditionResponse = messaging.MessagingConditionResponse;
 import DataMessagePayload = messaging.DataMessagePayload;
 import NotificationMessagePayload = messaging.NotificationMessagePayload;
 
+// GAPIC Generated client types
+import * as protos from '../generated/messaging/protos/protos';
+import { FcmServiceClient } from '../generated/messaging/src/v1/fcm_service_client'
+import ClientMessage = protos.google.firebase.fcm.v1.IMessage;
```
hiranya911, line 48: Use the original names `IMessage` and `ISendMessageRequest`. I don't see a reason to change them.

src/messaging/messaging.ts
```diff
@@ -228,6 +235,55 @@ export class Messaging implements MessagingInterface {
...
+   * type.
+   *
+   * @param message The {@link messaging.Message `Message`} to be converted
+   * @returns A {@link protos.google.firebase.fcm.v1.Message `clientMessage`},
+   * where like fields are the same as the inptuted message.
+   */
+  private convertToClientMessage(message: Message): ClientMessage {
+    //TODO: Add conversion for android, webpush, apns
+
+    const clientMessage: ClientMessage = {
```
hiranya911, line 250: Can we just do this:

```ts
return {
  data: message.data,
  notification: message.notification,
  fcmOptions: message.fcmOptions,
  token: message.token,
  topic: message.topic,
  condition: message.condition,
}
```

src/messaging/messaging.ts
```diff
@@ -228,6 +235,55 @@ export class Messaging implements MessagingInterface {
...
+    }
+
+    return clientMessage;
+  }
+
+  /**
+   * Method that gets the current App's credentials
+   * @returns The credential key of the current Firebase App
+   */
+  private getAppKey(): object{
```
hiranya911, line 271: This should be able to handle application default credentials too. I'd recommend implementing this as follows:

```ts
private getClientOptions(): object {
  const options = {fallback: 'rest'};
  // If service account credentials, add "credentials" to options.
  return options;
}
```

src/messaging/messaging.ts
```diff
@@ -245,15 +301,26 @@ export class Messaging implements MessagingInterface {
...
-        const request: { message: Message; validate_only?: boolean } = { message: copy };
-        if (dryRun) {
-          request.validate_only = true;
-        }
-        return this.messagingRequestHandler.invokeRequestHandler(FCM_SEND_HOST, urlPath, request);
+
+    const clientMessage = this.convertToClientMessage(copy);
+
+    const client = new FcmServiceClient({
+      credentials: this.getAppKey(),
```
hiranya911, line 308: We should try to set `projectId` here too, and remove `parent` field in the request.

src/messaging/messaging.ts
```diff
@@ -245,15 +301,26 @@ export class Messaging implements MessagingInterface {
...
+
+        const request: ClientSendRequest = {
+          parent: parent,
+          message: clientMessage,
+          validateOnly: dryRun
+        };
+        return client.sendMessage(request);
       })
-      .then((response) => {
+      .then(([response]) => {
```
hiranya911, line 323: `response` has type `IMessage` with this implementation. We should be able to remove the following `any` cast.

src/messaging/messaging.ts
```diff
@@ -245,15 +301,26 @@ export class Messaging implements MessagingInterface {
...
-      .then((urlPath) => {
-        const request: { message: Message; validate_only?: boolean } = { message: copy };
-        if (dryRun) {
-          request.validate_only = true;
-        }
-        return this.messagingRequestHandler.invokeRequestHandler(FCM_SEND_HOST, urlPath, request);
+
+    const clientMessage = this.convertToClientMessage(copy);
+
+    const client = new FcmServiceClient({
```
hiranya911, line 307: We should use the same `client` instance for multiple calls. You can do something like:

```ts
private getFcmServiceClient(): Promise<FcmServiceClient> {
  if (this.fcmServiceClient) {
    return Promise.resolve(this.fcmServiceClient);
  }

  return this.getClientOptions()
    .then((options) => {
      this.fcmServiceClient = new FcmServiceClient(options);
      return this.fcmServiceClient;
    });
}
```

hiranya911 (changes requested):
Looks mostly good. Few minor changes suggested. Also let's work on getting the unit tests to pass.

hiranya911:
Also sync up/rebase with the base branch so the gulpfile changes are included in the revision history. Without it we cannot run integration tests on this branch directly.

I was able to run integration tests by manually merging the gulpfile changes to this branch on my local machine:

```
% npm run test:integration -- --grep messaging

> firebase-admin@9.9.0 test:integration /Users/hkj/Projects/firebase-admin-node/public
> mocha test/integration/*.ts --slow 5000 --timeout 20000 --require ts-node/register "--grep" "messaging"

  admin.messaging
    ✓ send(message, dryRun) returns a message ID
```

Nice work 👍

src/messaging/messaging.ts
lahirumaramba: Sorry if this has already been noted, but shouldn't we have converters for each type?
```
return {
  data: message.data,
  notification: convertToClientNotification(message.notification),
  fcmOptions: convertToFcmOptions(message.fcmOptions),
  token: message.token,
  topic: message.topic,
  condition: message.condition,
}
``
Otherwise we could just `return message as ClientMessage`?

File: src/messaging/messaging.ts (hunk 5 above)
Comment (maceonthompson) [reply src/messaging/messaging.ts:324]: Resolved!

File: src/messaging/messaging.ts (hunk 4 above)
Comment (maceonthompson) [reply src/messaging/messaging.ts:309]: The parent field is a required part of the request ([See here](https://github.com/firebase/firebase-admin-node/blob/cc435abbb87df2db98330afc620c3551d94ea7c6/src/generated/messaging/src/v1/fcm_service_client.ts#L293)), so it's unfortunately not possible to set in the constructor.

File: src/messaging/messaging.ts [hunk 7]
```diff
@@ -41,6 +42,13 @@ import MessagingConditionResponse = messaging.MessagingConditionResponse;
 import DataMessagePayload = messaging.DataMessagePayload;
 import NotificationMessagePayload = messaging.NotificationMessagePayload;
 
+// GAPIC Generated client types
+import * as protos from '../generated/messaging/protos/protos';
+import { FcmServiceClient } from '../generated/messaging/src/v1/fcm_service_client'
+import ClientMessage = protos.google.firebase.fcm.v1.IMessage;
+import ClientSendRequest = protos.google.firebase.fcm.v1.ISendMessageRequest;
+import { createFirebaseErrorFromGapicError } from './messaging-errors-internal';
```
hiranya911, line 50: Move up closer to other module imports.

test/unit/messaging/messaging.spec.ts
```diff
@@ -357,6 +344,12 @@ describe('Messaging', () => {
       httpsRequestStub.restore();
     }
     getTokenStub.restore();
+
+    if (generatedClientStub) {
+      generatedClientStub.restore();
+    }
+    generatedClientStub = null;
```
hiranya911, line 351: Probably not required.

test/unit/messaging/messaging.spec.ts
```diff
@@ -485,19 +487,25 @@ describe('Messaging', () => {
...
+              'errorCode':'INVALID_ARGUMENT'
+            }
+          ],
+          'note': 'Exception occurred in retry method that was not classified as transient'
+        });
       return messaging.send(
         { token: 'mock-token' },
-      ).should.eventually.be.rejectedWith('test error message')
-        .and.have.property('code', 'messaging/invalid-argument');
+      ).should.eventually.be.rejectedWith('The registration token is not ' +
```
hiranya911, line 505: Should also assert for the error code here.

test/unit/messaging/messaging.spec.ts
```diff
@@ -3534,42 +3545,60 @@ describe('Messaging', () => {
...
+    ];
+
     validMessages.forEach((config) => {
       it(`should serialize well-formed Message: ${config.label}`, () => {
         // Wait for the initial getToken() call to complete before stubbing https.request.
         return mockApp.INTERNAL.getToken()
           .then(() => {
-            const resp = utils.responseFrom({ message: 'test' });
-            httpsRequestStub = sinon.stub(HttpClient.prototype, 'send').resolves(resp);
+            httpsRequestStub = sinon.stub(FcmServiceClient.prototype, 'sendMessage').resolves(GAPIC_RESPONSE);
```
hiranya911, line 3559: This test can be further simplified now. The `mockApp.INTERNAL.getToken()` part should no longer be necessary.

hiranya911: Shouldn't we use `generatedClientStub` here?

test/unit/messaging/messaging.spec.ts
```diff
@@ -485,19 +487,25 @@ describe('Messaging', () => {
...
+        });
       return messaging.send(
         { token: 'mock-token' },
-      ).should.eventually.be.rejectedWith('test error message')
-        .and.have.property('code', 'messaging/invalid-argument');
+      ).should.eventually.be.rejectedWith('The registration token is not ' +
+          'a valid FCM registration token');
     });
-
+    /*
```
hiranya911, line 508: Just remove this entirely. Makes it easier to review the diff.

hiranya911 (changes requested):
Looks mostly good. Please address the open comments from this and past reviews, and we should be good to go.

src/messaging/messaging.ts
hiranya911: The doc comments say so, but it's really optional in the type declartion:

https://github.com/firebase/firebase-admin-node/blob/cc435abbb87df2db98330afc620c3551d94ea7c6/src/generated/messaging/protos/protos.d.ts#L76-L79

Shouldn't be required to specify it.

src/messaging/messaging.ts
```diff
@@ -228,6 +237,63 @@ export class Messaging implements MessagingInterface {
...
+
+  /**
+   * Method that returns a fully instantiated service client.
+   *
+   * @returns an instantiated service client with either the application's
+   * service account credentials or the default credentials
+   */
+  private getFcmServiceClient(): FcmServiceClient {
+    if (this.fcmServiceClient) {
+      this.fcmServiceClient;
```
hiranya911, line 278: missing return

src/messaging/messaging.ts
```diff
@@ -228,6 +237,63 @@ export class Messaging implements MessagingInterface {
...
+    }
+
+    const credential = this.app.options.credential;
+    const options: { credentials?: object; fallback: 'rest' } = {
+      fallback: 'rest'
+    };
+
+    if (credential instanceof ServiceAccountCredential) {
```
hiranya911, line 286: Pass project ID here instead of as the `parent` field.

test/unit/messaging/messaging.spec.ts
maceonthompson: Added

test/unit/messaging/messaging.spec.ts
maceonthompson: Yes! Thanks for catching that.

test/unit/messaging/messaging.spec.ts
maceonthompson: Removed.

src/messaging/messaging.ts
maceonthompson: data, notification, and fcmOptions are easily converted because they only contain primitive types. In the future, there will be converters for the types that have more complicated structuring (i.e `androidConfig` and `webPush`). 
The suggestion of just returning an object with `token`, `topic` and `condition` regardless of the type of the `message` parameter would be cleaner, but does not compile.

test/unit/messaging/messaging.spec.ts
maceonthompson: Added

src/messaging/messaging.ts
maceonthompson: Even when including the `projectId` in the constructor for the client, omitting the `parent` field (the project ID) in the message request results in the generated client throwing an error, specifically stating that it `Cannot build HTTP request`. I'm assuming that the generated type declaration is incorrect here.

src/messaging/messaging.ts
maceonthompson: Agreed, added.

test/unit/messaging/messaging.spec.ts
maceonthompson: Removed a bit of redundant logic here.

src/messaging/messaging.ts
maceonthompson: Added.

src/messaging/messaging.ts
maceonthompson: The same issue as [this comment](https://github.com/firebase/firebase-admin-node/pull/1354#discussion_r679461191) applies here, unfortunately.

src/messaging/messaging.ts
maceonthompson: My thought was that naming it `ClientMessage` and `ClientSendRequest` would more clearly differentiate between our public types and the generated ones (alternatively something like `GapicMessage` or `GeneratedMessage` etc.). If that isn't necessary I can definitely change to the original names.

hiranya911: Let's use the original generated names here. I think the naming convention of `IMessage` and `ISendMessageRequest` itself serves as a good differentiator since we don't use `I` prefixed names in our code.

src/messaging/messaging.ts
```diff
@@ -245,16 +311,26 @@ export class Messaging implements MessagingInterface {
...
-    return this.getUrlPath()
-      .then((urlPath) => {
-        const request: { message: Message; validate_only?: boolean } = { message: copy };
-        if (dryRun) {
-          request.validate_only = true;
-        }
-        return this.messagingRequestHandler.invokeRequestHandler(FCM_SEND_HOST, urlPath, request);
+
+    const clientMessage = this.convertToClientMessage(copy);
+
```
hiranya911, line 316: Nit: Redundant whitespace

src/messaging/messaging.ts
```diff
@@ -245,16 +311,26 @@ export class Messaging implements MessagingInterface {
...
+    const client = this.getFcmServiceClient();
+
+    return client.getProjectId()
```
hiranya911, line 319: Is this looking up the project ID from `FcmServiceClient`? Then we must pass `projectId` to the `FcmServiceClient` constructor to avoid having GAPIC do extra work.

test/unit/messaging/messaging.spec.ts
```diff
@@ -466,17 +458,26 @@ describe('Messaging', () => {
       { token: 'mock-token' }, { topic: 'mock-topic' },
       { topic: '/topics/mock-topic' }, { condition: '"foo" in topics' },
     ];
+    const SEND_SUCCESSFUL_RETURN = [
+      // An examle IMessage that is normally returned by the generated client
```
hiranya911, line 462: Typo: example

test/unit/messaging/messaging.spec.ts
```diff
@@ -3534,42 +3470,56 @@ describe('Messaging', () => {
       },
     ];
 
+    const GAPIC_RESPONSE = [
```
hiranya911, line 3473: This was duplicated somewhere earlier in this file. Might want to make this a module-level constant so all test cases in this file can use it.

test/unit/messaging/messaging.spec.ts
```diff
@@ -3534,42 +3470,56 @@ describe('Messaging', () => {
...
+        });
       });
     });
 
     it('should not throw when the message is addressed to the prefixed topic name', () => {
       return mockApp.INTERNAL.getToken()
         .then(() => {
-          const resp = utils.responseFrom({ message: 'test' });
-          httpsRequestStub = sinon.stub(HttpClient.prototype, 'send').resolves(resp);
+          //const resp = utils.responseFrom({ message: 'test' });
```
hiranya911, line 3509: Please remove.

Also the call to `INTERNAL.getToken()` should no longer be required.

hiranya911 (changes requested):
Thanks for making the changes and for the responses to earlier questions. Looks pretty good. I only pointed out a few nits, but this is almost done.

src/messaging/messaging.ts
maceonthompson: Understood, made the change!

test/unit/messaging/messaging.spec.ts
maceonthompson: Removed!

test/unit/messaging/messaging.spec.ts
maceonthompson: Addressed!

test/unit/messaging/messaging.spec.ts
maceonthompson: Fixed

src/messaging/messaging.ts
maceonthompson: Made the adjustment, using `utils.findProjectId()`

src/messaging/messaging.ts
maceonthompson: Now that the client is returned as a promise, this is the only variable declared. Is it better practice still to remove the line break between the declaration and the return statement?

src/messaging/messaging.ts
```diff
@@ -290,8 +290,22 @@ export class Messaging implements MessagingInterface {
...
+        throw new FirebaseMessagingError(
+          MessagingClientErrorCode.INVALID_ARGUMENT,
+          'Failed to determine project ID for Messaging. Initialize the '
+          + 'SDK with service account credentials or set project ID as an app option. '
+          + 'Alternatively set the GOOGLE_CLOUD_PROJECT environment variable.',
+        );
+      }
+
+      options.projectId = projectId;
+    }).then(() => {
```
hiranya911, line 305: You can drop this line safely, and it would still work.

hiranya911 (approved):
Thanks. LGTM 👍

PR #1401, opened 2021-08-04
fix: Throw error on user disabled and check revoked set true

xil222:
Add the logic to throw error in verifySessionCookie and verifyIdToken when user disabled and checkRevoked is true.
The logic might not look most efficient, but it is rather a simple approach without changing logic and breaking other tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,6 @@ jobs:
       run: npm run api-extractor
     - name: Run emulator-based integration tests
       run: |
-        npm install -g firebase-tools
+        npm install -g firebase-tools@9.16.0
         firebase emulators:exec --project fake-project-id --only auth,database,firestore \
           'npx mocha \"test/integration/{auth,database,firestore}.spec.ts\" --slow 5000 --timeout 20000 --require ts-node/register'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,6 @@ jobs:
       run: npm run api-extractor
     - name: Run emulator-based integration tests
       run: |
-        npm install -g firebase-tools@9.16.0
+        npm install -g firebase-tools
         firebase emulators:exec --project fake-project-id --only auth,database,firestore \
           'npx mocha \"test/integration/{auth,database,firestore}.spec.ts\" --slow 5000 --timeout 20000 --require ts-node/register'
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -745,7 +745,6 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
     // Get tokens valid after time for the corresponding user.
     return this.getUser(decodedIdToken.sub)
       .then((user: UserRecord) => {
-        // If user disabled, throw an auth/user-disabled error.
         if (user.disabled) {
           throw new FirebaseAuthError(
             AuthClientErrorCode.USER_DISABLED,
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -104,19 +104,20 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
   }
 
   /**
-   * Verifies a JWT auth token. Returns a Promise with the tokens claims. Rejects
-   * the promise if the token could not be verified.
-   * If checkRevoked is set to true, first verifies whether the corresponding userinfo.disabled
-   * is true:
+   * Verifies a JWT auth token. Returns a Promise with the tokens claims.
+   * Rejects the promise if the token could not be verified.
+   * If checkRevoked is set to true, first verifies whether the corresponding
+   * user is disabled.
    * If yes, an auth/user-disabled error is thrown.
    * If no, verifies if the session corresponding to the ID token was revoked.
-   * If the corresponding user's session was invalidated, an auth/id-token-revoked error is thrown.
+   * If the corresponding user's session was invalidated, an
+   * auth/id-token-revoked error is thrown.
    * If not specified the check is not applied.
    *
    * @param {string} idToken The JWT to verify.
    * @param {boolean=} checkRevoked Whether to check if the ID token is revoked.
-   * @return {Promise<DecodedIdToken>} A Promise that will be fulfilled after a successful
-   *     verification.
+   * @return {Promise<DecodedIdToken>} A Promise that will be fulfilled after
+   *   a successful verification.
    */
   public verifyIdToken(idToken: string, checkRevoked = false): Promise<DecodedIdToken> {
     const isEmulator = useEmulator();
@@ -509,20 +510,22 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
   }
 
   /**
-   * Verifies a Firebase session cookie. Returns a Promise with the tokens claims. Rejects
-   * the promise if the cookie could not be verified. 
-   * If checkRevoked is set to true, first verifies whether the corresponding userinfo.disabled
-   * is true:
+   * Verifies a Firebase session cookie. Returns a Promise with the tokens claims.
+   * Rejects the promise if the cookie could not be verified. 
+   * If checkRevoked is set to true, first verifies whether the corresponding
+   * user is true:
    * If yes, an auth/user-disabled error is thrown.
-   * If no, verifies if the session corresponding to the session cookie was revoked.
-   * If the corresponding user's session was invalidated, an auth/session-cookie-revoked error
-   * is thrown.
+   * If no, verifies if the session corresponding to the session cookie was
+   * revoked.
+   * If the corresponding user's session was invalidated, an
+   * auth/session-cookie-revoked error is thrown.
    * If not specified the check is not performed.
    *
    * @param {string} sessionCookie The session cookie to verify.
-   * @param {boolean=} checkRevoked Whether to check if the session cookie is revoked.
-   * @return {Promise<DecodedIdToken>} A Promise that will be fulfilled after a successful
-   *     verification.
+   * @param {boolean=} checkRevoked Whether to check if the session cookie is
+   *     revoked.
+   * @return {Promise<DecodedIdToken>} A Promise that will be fulfilled after
+   *     a successful verification.
    */
   public verifySessionCookie(
     sessionCookie: string, checkRevoked = false): Promise<DecodedIdToken> {
@@ -530,7 +533,7 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
     return this.sessionCookieVerifier.verifyJWT(sessionCookie, isEmulator)
       .then((decodedIdToken: DecodedIdToken) => {
         // Whether to check if the cookie was revoked.
-        if (checkRevoked || isEmulator) {         
+        if (checkRevoked || isEmulator) {
           return this.verifyDecodedJWTNotRevokedOrDisabled(
             decodedIdToken,
             AuthClientErrorCode.SESSION_COOKIE_REVOKED);
@@ -748,7 +751,7 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
         if (user.disabled) {
           throw new FirebaseAuthError(
             AuthClientErrorCode.USER_DISABLED,
-            'The user account has been disabled by an administrator');
+            'The user record is disabled.');
         }
         // If no tokens valid after time available, token is not revoked.
         if (user.tokensValidAfterTime) {
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -124,21 +124,7 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
       .then((decodedIdToken: DecodedIdToken) => {
         // Whether to check if the token was revoked.
         if (checkRevoked || isEmulator) {
-          // Whether user has been disabled.
-          if (checkRevoked) {
-            const uid = decodedIdToken.uid;
-            return this.getUser(uid)
-              .then((userInfo: UserRecord) => {
-                if (userInfo.disabled) {
-                  throw new FirebaseAuthError(
-                    AuthClientErrorCode.USER_DISABLED, 'The user account has been disabled by an administrator');
-                } 
-                return this.verifyDecodedJWTNotRevoked(
-                  decodedIdToken,
-                  AuthClientErrorCode.ID_TOKEN_REVOKED);
-              });
-          }
-          return this.verifyDecodedJWTNotRevoked(
+          return this.verifyDecodedJWTNotRevokedOrDisabled(
             decodedIdToken,
             AuthClientErrorCode.ID_TOKEN_REVOKED);
         }
@@ -524,11 +510,11 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
 
   /**
    * Verifies a Firebase session cookie. Returns a Promise with the tokens claims. Rejects
-   * the promise if the token could not be verified. 
+   * the promise if the cookie could not be verified. 
    * If checkRevoked is set to true, first verifies whether the corresponding userinfo.disabled
    * is true:
    * If yes, an auth/user-disabled error is thrown.
-   * If no, verifies if the session corresponding to the ID token was revoked.
+   * If no, verifies if the session corresponding to the session cookie was revoked.
    * If the corresponding user's session was invalidated, an auth/session-cookie-revoked error
    * is thrown.
    * If not specified the check is not performed.
@@ -543,23 +529,9 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
     const isEmulator = useEmulator();
     return this.sessionCookieVerifier.verifyJWT(sessionCookie, isEmulator)
       .then((decodedIdToken: DecodedIdToken) => {
-        // Whether to check if the token was revoked.
-        if (checkRevoked || isEmulator) {
-          // Whether user has been disabled.
-          if (checkRevoked) {
-            const uid = decodedIdToken.uid;
-            return this.getUser(uid)
-              .then((userInfo: UserRecord) => {
-                if (userInfo.disabled) {
-                  throw new FirebaseAuthError(
-                    AuthClientErrorCode.USER_DISABLED, 'The user account has been disabled by an administrator');
-                } 
-                return this.verifyDecodedJWTNotRevoked(
-                  decodedIdToken,
-                  AuthClientErrorCode.SESSION_COOKIE_REVOKED);
-              });
-          }          
-          return this.verifyDecodedJWTNotRevoked(
+        // Whether to check if the cookie was revoked.
+        if (checkRevoked || isEmulator) {         
+          return this.verifyDecodedJWTNotRevokedOrDisabled(
             decodedIdToken,
             AuthClientErrorCode.SESSION_COOKIE_REVOKED);
         }
@@ -758,8 +730,9 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
   }
 
   /**
-   * Verifies the decoded Firebase issued JWT is not revoked. Returns a promise that resolves
-   * with the decoded claims on success. Rejects the promise with revocation error if revoked.
+   * Verifies the decoded Firebase issued JWT is not revoked or disabled. Returns a promise that
+   * resolves with the decoded claims on success. Rejects the promise with revocation error if revoked
+   * or user disabled.
    *
    * @param {DecodedIdToken} decodedIdToken The JWT's decoded claims.
    * @param {ErrorInfo} revocationErrorInfo The revocation error info to throw on revocation
@@ -767,11 +740,17 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
    * @return {Promise<DecodedIdToken>} A Promise that will be fulfilled after a successful
    *     verification.
    */
-  private verifyDecodedJWTNotRevoked(
+  private verifyDecodedJWTNotRevokedOrDisabled(
     decodedIdToken: DecodedIdToken, revocationErrorInfo: ErrorInfo): Promise<DecodedIdToken> {
     // Get tokens valid after time for the corresponding user.
     return this.getUser(decodedIdToken.sub)
       .then((user: UserRecord) => {
+        // If user disabled, throw an auth/user-disabled error.
+        if (user.disabled) {
+          throw new FirebaseAuthError(
+            AuthClientErrorCode.USER_DISABLED,
+            'The user account has been disabled by an administrator');
+        }
         // If no tokens valid after time available, token is not revoked.
         if (user.tokensValidAfterTime) {
           // Get the ID token authentication time and convert to milliseconds UTC.
--- a/src/auth/auth.ts
+++ b/src/auth/auth.ts
@@ -105,10 +105,13 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
 
   /**
    * Verifies a JWT auth token. Returns a Promise with the tokens claims. Rejects
-   * the promise if the token could not be verified. If checkRevoked is set to true,
-   * verifies if the session corresponding to the ID token was revoked. If the corresponding
-   * user's session was invalidated, an auth/id-token-revoked error is thrown. If not specified
-   * the check is not applied.
+   * the promise if the token could not be verified.
+   * If checkRevoked is set to true, first verifies whether the corresponding userinfo.disabled
+   * is true:
+   * If yes, an auth/user-disabled error is thrown.
+   * If no, verifies if the session corresponding to the ID token was revoked.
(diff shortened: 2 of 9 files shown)


src/auth/auth.ts
```diff
@@ -121,6 +124,20 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
       .then((decodedIdToken: DecodedIdToken) => {
         // Whether to check if the token was revoked.
         if (checkRevoked || isEmulator) {
+          // Whether user has been disabled.
+          if (checkRevoked) {
+            const uid = decodedIdToken.uid;
```
hiranya911, line 129: I think we should handle this in `verifyDecodedJWTNotRevoked()`. Otherwise we will end up calling `getUser()` twice (i.e. 2 RPCs).

src/auth/auth.ts
```diff
@@ -524,6 +545,20 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
...
```
hiranya911, line 550: Same here. This logic should be executed at `verifyDecodedJWTNotRevoked()`.

test/integration/auth.spec.ts
```diff
@@ -833,53 +833,120 @@ describe('admin.auth', () => {
...
+          expect(decodedIdToken.uid).to.equal(uid);
+          expect(decodedIdToken.email).to.equal(email);
+          return admin.auth().updateUser(uid, { disabled: true })
+        })
+        .then((userRecord) => {
+          // Ensure disabled field has been updated.
+          expect(userRecord.uid).to.equal(uid);
+          expect(userRecord.email).to.equal(email);
+          expect(userRecord.disabled).to.equal(true);
+          return clientAuth().signInWithEmailAndPassword(email, password)
```
hiranya911, line 888: Shouldn't this fail now that the user account is disabled?

test/integration/auth.spec.ts
```diff
@@ -1982,6 +2049,50 @@ describe('admin.auth', () => {
...
+          admin.auth().verifySessionCookie(sessionCookie, true)
+        })
+        .then(() => {
+          return admin.auth().updateUser(uid, { disabled : true });
+        })
+        .then((userRecord) => {
+          // Ensure disabled field has been updated.
+          expect(userRecord.uid).to.equal(uid);
+          expect(userRecord.disabled).to.equal(true);
+          return clientAuth().signInWithCustomToken(currentCustomToken)
```
hiranya911, line 2086: Same here. I'd  expect this sign in attempt to fail since the user account is disabled.

hiranya911 (changes requested):
Needs a bit of refactoring to make sure we don't call `getUser()` more than once. But mostly looks pretty good.

test/integration/auth.spec.ts
xil222: Yes, I'll add a test case for updating a user by setting UserRecord.disabled = true, then signin get rejected.

For here, I think the logic is as expected that this shouldn't fail, because we've signed in once and get an 1-hour token, the 2nd signInWithEmail isn't the real 'sign in', but like a refresh, I need to do this to detect a second-call on verifyIdToken returns error.

xil222: An example of 2nd calling signInWithEmailAndPassword doing refresh [here](https://github.com/firebase/firebase-admin-node/blob/ae944f4d083a7087224073711bf7fbc0bac84da1/test/integration/auth.spec.ts#L471)

hiranya911: In that case can we just skip this step entirely, and just do the following?

```
return admin.auth().verifyIdToken(currentIdToken, true)
```

Then it's clear we are re-verifying the same ID token as before, but this time it's expected to fail.

xil222: It makes sense to me that the 2nd verifyIdTokens should fail without a second sign-in, however it doesn't, which doesn't make sense(I understand in Python, that is how it works). I haven't found the reason why it doesn't work, all I find out is the similar example linked above(calling sign in twice), so I suspect there's a refresh needed. It could be hard for me to find another reason explaining that.

src/auth/auth.ts
```diff
@@ -121,6 +124,20 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
...
+            return this.getUser(uid)
```
bojeil-google, line 130: You are calling `getUser` twice. Here and in `verifyDecodedJWTNotRevoked`. You need to refactor `verifyDecodedJWTNotRevoked` to do both checks.

See here: https://github.com/firebase/firebase-admin-node/blob/b88f8919c67c252b31d6b4b455fd7ba9b6226131/src/auth/auth.ts#L738

src/auth/auth.ts
```diff
@@ -507,10 +524,14 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
 
   /**
    * Verifies a Firebase session cookie. Returns a Promise with the tokens claims. Rejects
-   * the promise if the token could not be verified. If checkRevoked is set to true,
-   * verifies if the session corresponding to the session cookie was revoked. If the corresponding
-   * user's session was invalidated, an auth/session-cookie-revoked error is thrown. If not
-   * specified the check is not performed.
+   * the promise if the token could not be verified.
```
bojeil-google, line 527: Replace token with cookie

src/auth/auth.ts
```diff
@@ -507,10 +524,14 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
...
    * Verifies a Firebase session cookie. Returns a Promise with the tokens claims. Rejects
-   * the promise if the token could not be verified. If checkRevoked is set to true,
-   * verifies if the session corresponding to the session cookie was revoked. If the corresponding
-   * user's session was invalidated, an auth/session-cookie-revoked error is thrown. If not
-   * specified the check is not performed.
+   * the promise if the token could not be verified. 
+   * If checkRevoked is set to true, first verifies whether the corresponding userinfo.disabled
+   * is true:
+   * If yes, an auth/user-disabled error is thrown.
+   * If no, verifies if the session corresponding to the ID token was revoked.
```
bojeil-google, line 531: Replace ID token with session cookie

src/auth/auth.ts
```diff
@@ -524,6 +545,20 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
       .then((decodedIdToken: DecodedIdToken) => {
         // Whether to check if the token was revoked.
         if (checkRevoked || isEmulator) {
+          // Whether user has been disabled.
+          if (checkRevoked) {
+            const uid = decodedIdToken.uid;
+            return this.getUser(uid)
```
bojeil-google, line 551: Same here as above.

src/utils/error.ts
```diff
@@ -721,6 +721,10 @@ export class AuthClientErrorCode {
     code: 'not-found',
     message: 'The requested resource was not found.',
   };
+  public static USER_DISABLED = {
+    code: 'user-disabled',
+    message: 'The user account has been disabled by an administrator.',
```
bojeil-google, line 726: Maybe we shouldn't specify it is disabled by an admin
'The user record is disabled.'

test/integration/auth.spec.ts
bojeil-google: I think `return clientAuth().signInWithEmailAndPassword(email, password)` will throw `auth/user-disabled` error. It will jump the next then statements until it reaches the error block. This will also pass since the client op will fail with `auth/user-disabled`. Try it out:
```
return clientAuth().signInWithEmailAndPassword(email, password)
  .then(() => {
    console.log('1');
  })
  .then(() => {
    console.log('2');
  })
  .then(() => {
    throw new Error('Unexpected success');
  }, (error) => {
    console.log(error.code);
  });
```

It will not print '1' or '2' and just print the error code

bojeil-google (changes requested):

test/integration/auth.spec.ts
xil222: I remove the 2nd signInWithEmailAndPassword, and using currentUser.reload() to make sure 2nd call verifyIdToken after modification will fail(otherwise it won't).

src/auth/auth.ts
```diff
@@ -723,20 +730,27 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
...
    * @return {Promise<DecodedIdToken>} A Promise that will be fulfilled after a successful
    *     verification.
    */
-  private verifyDecodedJWTNotRevoked(
+  private verifyDecodedJWTNotRevokedOrDisabled(
     decodedIdToken: DecodedIdToken, revocationErrorInfo: ErrorInfo): Promise<DecodedIdToken> {
     // Get tokens valid after time for the corresponding user.
     return this.getUser(decodedIdToken.sub)
       .then((user: UserRecord) => {
+        // If user disabled, throw an auth/user-disabled error.
```
hiranya911, line 748: Nit: Comment unnecessary. Pretty obvious from the code what's going on here.

test/integration/auth.spec.ts
```diff
@@ -833,53 +847,120 @@ describe('admin.auth', () => {
...
+          return admin.auth().updateUser(uid, { disabled: true });
+        })
+        .then((userRecord) => {
+          // Ensure disabled field has been updated.
+          expect(userRecord.uid).to.equal(uid);
+          expect(userRecord.email).to.equal(email);
+          expect(userRecord.disabled).to.equal(true);
+          return admin.auth().verifyIdToken(currentIdToken, false);
+        })
+        .then(() => {
```
hiranya911, line 906: This test passed for me with this block commented out (i.e. no reload). I added some extra logging to the implementation to log the user record every time the token is verified. Here's my output:

```
> mocha test/integration/*.ts --slow 5000 --timeout 20000 --require ts-node/register



  admin.auth
    verifyIdToken()
{"uid":"2qvksybtzo8ecdnvu0de","email":"2qvksybtzo8ecdnvu0de@example.com","emailVerified":false,"disabled":false,"metadata":{"lastSignInTime":"Tue, 10 Aug 2021 18:11:43 GMT","creationTime":"Tue, 10 Aug 2021 18:11:43 GMT"},"tokensValidAfterTime":"Tue, 10 Aug 2021 18:11:43 GMT","providerData":[{"uid":"2qvksybtzo8ecdnvu0de@example.com","email":"2qvksybtzo8ecdnvu0de@example.com","providerId":"password"}]}
{"uid":"2qvksybtzo8ecdnvu0de","email":"2qvksybtzo8ecdnvu0de@example.com","emailVerified":false,"disabled":true,"metadata":{"lastSignInTime":"Tue, 10 Aug 2021 18:11:43 GMT","creationTime":"Tue, 10 Aug 2021 18:11:43 GMT"},"tokensValidAfterTime":"Tue, 10 Aug 2021 18:11:43 GMT","providerData":[{"uid":"2qvksybtzo8ecdnvu0de@example.com","email":"2qvksybtzo8ecdnvu0de@example.com","providerId":"password"}]}
      ✓ verifyIdToken() fails with checkRevoked set to true and corresponding user disabled


  1 passing (7s)
```

May be we can try commenting this block out, and pushing the change to GitHub? I've enabled staging on this PR, so future pushes to this PR will trigger integration tests (note: you will also have to remove the `currentUser` local variable if you comment out this block).

hiranya911 (commented):
The integration test passed for me without `currentUser.reload()`. Can we try that out with our CI and see what happens?

test/integration/auth.spec.ts
xil222: looks like it doesn't work as expected

hiranya911 (approved):
LGTM 👍

src/auth/auth.ts
```diff
@@ -105,10 +105,13 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
 
   /**
    * Verifies a JWT auth token. Returns a Promise with the tokens claims. Rejects
-   * the promise if the token could not be verified. If checkRevoked is set to true,
-   * verifies if the session corresponding to the ID token was revoked. If the corresponding
-   * user's session was invalidated, an auth/id-token-revoked error is thrown. If not specified
-   * the check is not applied.
+   * the promise if the token could not be verified.
+   * If checkRevoked is set to true, first verifies whether the corresponding userinfo.disabled
+   * is true:
```
bojeil-google, line 110: nit: If checkRevoked is set to true, first verifies whether the corresponding  user is disabled.

src/auth/auth.ts
```diff
@@ -507,10 +510,14 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
 
   /**
    * Verifies a Firebase session cookie. Returns a Promise with the tokens claims. Rejects
-   * the promise if the token could not be verified. If checkRevoked is set to true,
-   * verifies if the session corresponding to the session cookie was revoked. If the corresponding
-   * user's session was invalidated, an auth/session-cookie-revoked error is thrown. If not
-   * specified the check is not performed.
+   * the promise if the cookie could not be verified. 
+   * If checkRevoked is set to true, first verifies whether the corresponding userinfo.disabled
+   * is true:
```
bojeil-google, line 515: nit: If checkRevoked is set to true, first verifies whether the corresponding user is disabled.

src/auth/auth.ts
```diff
@@ -723,20 +730,26 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
...
-  private verifyDecodedJWTNotRevoked(
+  private verifyDecodedJWTNotRevokedOrDisabled(
     decodedIdToken: DecodedIdToken, revocationErrorInfo: ErrorInfo): Promise<DecodedIdToken> {
     // Get tokens valid after time for the corresponding user.
     return this.getUser(decodedIdToken.sub)
       .then((user: UserRecord) => {
+        if (user.disabled) {
+          throw new FirebaseAuthError(
+            AuthClientErrorCode.USER_DISABLED,
+            'The user account has been disabled by an administrator');
```
bojeil-google, line 751: I don't think we should specify how it was disabled. It could be disabled automatically due to suspicious activity.
'The user record is disabled.'

test/integration/auth.spec.ts
```diff
@@ -833,53 +847,113 @@ describe('admin.auth', () => {
...
+          return admin.auth().verifyIdToken(currentIdToken, true);
+        })
+        .then(() => {
+          throw new Error('Unexpected success');
+        }, (error) => {
+          expect(error).to.have.property('code', 'auth/user-disabled');
         });
```
bojeil-google, line 911: I think this check is still not reliable. You can't guarantee which promise is rejecting:
`return admin.auth().verifyIdToken(currentIdToken, true);` or
`return admin.auth().verifyIdToken(currentIdToken, false);`

Here is an example:
```
p1 // assume this throws error foo.
  .then(() => {
    // This will be skipped.
    return p2;
  })
  .then(() => {
    // This will be skipped.
    throw new Error('Unexpected success');
  }, (error) => {
    // You can't tell if p1 or p2 threw the error.
    expect(error.message).toEqual('foo');
  });
```

test/integration/auth.spec.ts
```diff
@@ -1982,6 +2056,48 @@ describe('admin.auth', () => {
...
+        .then(() => {
+          return admin.auth().updateUser(uid, { disabled : true });
+        })
+        .then((userRecord) => {
+          // Ensure disabled field has been updated.
+          expect(userRecord.uid).to.equal(uid);
+          expect(userRecord.disabled).to.equal(true);
+          return admin.auth().verifySessionCookie(currentSessioncookie, false);
+        }).then(() => {
+          return admin.auth().verifySessionCookie(currentSessioncookie, true);
```
bojeil-google, line 2093: Same issue here. we can't tell for sure which one is throwing the error.

test/unit/auth/auth.spec.ts
```diff
@@ -838,6 +857,25 @@ AUTH_CONFIGS.forEach((testConfig) => {
...
 
+      it('should be rejected with checkRevoked set to true and corresponding user disabled', () =>  {
+        const expectedAccountInfoResponse = getValidGetAccountInfoResponse(tenantId);
+        expectedAccountInfoResponse.users[0].disabled = true;
+        const expectedUserRecordDisabled = getValidUserRecord(expectedAccountInfoResponse);
+        const getUserStub = sinon.stub(testConfig.Auth.prototype, 'getUser')
+          .resolves(expectedUserRecordDisabled);
+        expect(expectedUserRecordDisabled.disabled).to.be.equal(true);
+        stubs.push(getUserStub);
+        return auth.verifyIdToken(mockSessionCookie, true)
```
bojeil-google, line 868: This is testing `verifyIdToken` and not `verifySessionCookie`

test/unit/auth/auth.spec.ts
```diff
@@ -482,6 +482,25 @@ AUTH_CONFIGS.forEach((testConfig) => {
           .should.eventually.be.rejectedWith('Decoding Firebase ID token failed');
       });
 
+      it('should be rejected with checkRevoked set to true and corresponding user disabled', () =>  {
```
bojeil-google, line 485: Can you add 1 more test when both token is revoked and user is disabled. We should throw user disabled error as we check that first.

test/unit/auth/auth.spec.ts
```diff
@@ -838,6 +857,25 @@ AUTH_CONFIGS.forEach((testConfig) => {
           });
       });
 
+      it('should be rejected with checkRevoked set to true and corresponding user disabled', () =>  {
```
bojeil-google, line 860: Can you add 1 more test when both token is revoked and user is disabled. We should throw user disabled error as we check that first.

bojeil-google (changes requested):

src/auth/auth.ts
```diff
@@ -104,24 +104,28 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
   }
 
   /**
-   * Verifies a JWT auth token. Returns a Promise with the tokens claims. Rejects
-   * the promise if the token could not be verified. If checkRevoked is set to true,
-   * verifies if the session corresponding to the ID token was revoked. If the corresponding
-   * user's session was invalidated, an auth/id-token-revoked error is thrown. If not specified
-   * the check is not applied.
+   * Verifies a JWT auth token. Returns a Promise with the tokens claims.
```
egilmorez, line 107: This should be lowercase unless it's a (backticked) literal, I think.

src/auth/auth.ts
```diff
@@ -104,24 +104,28 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
...
+   * Rejects the promise if the token could not be verified.
```
egilmorez, line 108: Suggest "cannot"

Present tense always preferred.

src/auth/auth.ts
```diff
@@ -104,24 +104,28 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
...
+   * If checkRevoked is set to true, first verifies whether the corresponding
```
egilmorez, line 109: Looks like a literal; therefore, backticks.

src/auth/auth.ts
```diff
@@ -104,24 +104,28 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
...
+   * If no, verifies if the session corresponding to the ID token was revoked.
+   * If the corresponding user's session was invalidated, an
+   * auth/id-token-revoked error is thrown.
+   * If not specified the check is not applied.
    *
    * @param {string} idToken The JWT to verify.
    * @param {boolean=} checkRevoked Whether to check if the ID token is revoked.
-   * @return {Promise<DecodedIdToken>} A Promise that will be fulfilled after a successful
-   *     verification.
+   * @return {Promise<DecodedIdToken>} A Promise that will be fulfilled after
```
egilmorez, line 119: Same here RE caps/literal.

src/auth/auth.ts
```diff
@@ -506,25 +510,31 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
   }
 
   /**
-   * Verifies a Firebase session cookie. Returns a Promise with the tokens claims. Rejects
-   * the promise if the token could not be verified. If checkRevoked is set to true,
-   * verifies if the session corresponding to the session cookie was revoked. If the corresponding
-   * user's session was invalidated, an auth/session-cookie-revoked error is thrown. If not
-   * specified the check is not performed.
+   * Verifies a Firebase session cookie. Returns a Promise with the tokens claims.
```
egilmorez, line 513: Literal?

src/auth/auth.ts
```diff
@@ -506,25 +510,31 @@ export class BaseAuth<T extends AbstractAuthRequestHandler> implements BaseAuthI
...
+   * Rejects the promise if the cookie could not be verified. 
+   * If checkRevoked is set to true, first verifies whether the corresponding
```
egilmorez, line 515: Backticks.

egilmorez (commented):
A few style things to look at, thanks!

test/integration/auth.spec.ts
```diff
@@ -890,6 +890,7 @@ describe('admin.auth', () => {
           return admin.auth().verifyIdToken(currentIdToken, true);
         })
         .then((decodedIdToken) => {
+          expect(decodedIdToken.sub).to.equal(uid);
```
bojeil-google, line 893: I am not sure how this solves the issue I pointed out earlier.

How about you use async/await instead of chained promises in these tests. It would be cleaner to read too.
```
....
// If this throws, it won't be caught and the test will fail.
const decodedIdToken = await admin.auth().verifyIdToken(currentIdToken, false);
expect(decodedIdToken.uid).to.equal(uid);
try {
  await admin.auth().verifyIdToken(currentIdToken, true);
} catch (e) {
  // You know for sure that this was thrown by the second call.
  expect(error).to.have.property('code', 'auth/user-disabled');
}
```

test/integration/auth.spec.ts
```diff
@@ -2081,15 +2085,19 @@ describe('admin.auth', () => {
           currentSessioncookie = sessionCookie;
           return admin.auth().verifySessionCookie(sessionCookie, true);
         })
-        .then(() => {
+        .then((decodedIdToken) => {
```
bojeil-google, line 2088: Same suggestion to switch to async/await.

bojeil-google (changes requested):

src/auth/auth.ts
xil222: Hey @egilmorez , I think there are 16 JWT being used in this file and 50 times+ in other files as well, along I agree with you, but since that wasn't my change, I am not I should change those.

xil222:
It seems like tests passing on my end can't pass the CI.

.github/workflows/ci.yml
```diff
@@ -28,6 +28,6 @@ jobs:
       run: npm run api-extractor
     - name: Run emulator-based integration tests
       run: |
-        npm install -g firebase-tools
+        npm install -g firebase-tools@9.16.0
```
hiranya911, line 31: The underlying issue has been resolved with latest firebase-tools (v9.16.5). This shouldn't be needed any more.

hiranya911 (approved):
Thanks. LGTM with a comment.

test/unit/auth/auth.spec.ts
```diff
@@ -501,6 +501,38 @@ AUTH_CONFIGS.forEach((testConfig) => {
...
+        const validSince = new Date(expectedUserRecordDisabled.tokensValidAfterTime!);
+        // One second before validSince.
+        const oneSecBeforeValidSince = Math.floor(validSince.getTime() / 1000 - 1);
+        const getUserStub = sinon.stub(testConfig.Auth.prototype, 'getUser')
+          .resolves(expectedUserRecordDisabled);
+        expect(expectedUserRecordDisabled.disabled).to.be.equal(true);
+        stubs.push(getUserStub);
+
+        const unsignedToken = mocks.generateIdToken({
+          algorithm: 'none',
```
bojeil-google, line 518: Can we not depend on algorithm being none which is specific to auth emulator.
You can just follow existing tests that touch on revocation:
https://github.com/firebase/firebase-admin-node/blob/a9d8b410848babfa9b1dfeedbf1ae00572deea57/test/unit/auth/auth.spec.ts#L541

test/unit/auth/auth.spec.ts
```diff
@@ -876,6 +908,39 @@ AUTH_CONFIGS.forEach((testConfig) => {
...
```
bojeil-google, line 925: same here

bojeil-google (changes requested):

egilmorez (approved):
LG regarding doc comments, thanks!

test/unit/auth/auth.spec.ts
```diff
@@ -914,23 +910,18 @@ AUTH_CONFIGS.forEach((testConfig) => {
         expectedAccountInfoResponseUserDisabled.users[0].disabled = true;
         const expectedUserRecordDisabled = getValidUserRecord(expectedAccountInfoResponseUserDisabled);
         const validSince = new Date(expectedUserRecordDisabled.tokensValidAfterTime!);
+        // Restore verifyIdToken stub.
```
bojeil-google, line 913: verifySessionCookie

bojeil-google (approved):
Just one incorrect comment.

PR #1409, opened 2021-08-13
chore: Add emulator tests to nightlies

lahirumaramba:
- Run emulator based integration tests as part of the nightly tests.


hiranya911 (approved):
LGTM 👍 

The emulator team is working on the test failures we are seeing here (regression introduced in firebase-tools@9.16.1). We might want to hold off on the merge until the fix is released.

PR #1411, opened 2021-08-16
chore: Merging main branch into modular-sdk

hiranya911:
Key changes include:

* Updates to the App Check API.
* Addition of the Installations API and the deprecation of InstanceId API.
* Upgrades to various dependencies and tools.


etc/firebase-admin.api.md
```diff
@@ -52,6 +56,10 @@ export namespace appCheck {
     export type AppCheck = AppCheck;
     // Warning: (ae-forgotten-export) The symbol "AppCheckToken" needs to be exported by the entry point default-namespace.d.ts
     export type AppCheckToken = AppCheckToken;
+    // Warning: (ae-forgotten-export) The symbol "AppCheckTokenOptions" needs to be exported by the entry point default-namespace.d.ts
```
lahirumaramba, line 59: Interesting... I thought these warning were fixed. :) Are these expected?

lahirumaramba (approved):
LGTM! Thanks!
Left one comment.

etc/firebase-admin.api.md
hiranya911: They will continue to appear in this file, which declares the old "admin.appCheck" namespace (see the other child namespaces in the same file). This is because the old namespaces only export type aliases, and not the source types themselves.

PR #1436, opened 2021-09-22
Update SignOptions to fix a type error

lahirumaramba:
In `@types/jsonwebtoken@8.5.5`  the data type of the property `header` was [changed](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53435/files#r699621093) from `header: object` to `header: JWTHeader`. This results in type errors in our test build. This PR fixes it by setting the `alg` in `header` and removing the optional `algorithm` from the top level.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "@types/chai": "^4.0.0",
     "@types/chai-as-promised": "^7.1.0",
     "@types/firebase-token-generator": "^2.0.28",
-    "@types/jsonwebtoken": "^8.5.0",
+    "@types/jsonwebtoken": "^8.5.5",
     "@types/lodash": "^4.14.104",
     "@types/minimist": "^1.2.0",
     "@types/mocha": "^8.2.2",
--- a/test/resources/mocks.ts
+++ b/test/resources/mocks.ts
@@ -25,11 +25,14 @@ import stream = require('stream');
 import * as _ from 'lodash';
 import * as jwt from 'jsonwebtoken';
 
+import * as validator from '../../src/utils/validator';
+
 import { AppOptions } from '../../src/firebase-namespace-api';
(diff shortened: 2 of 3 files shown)


test/resources/mocks.ts
```diff
@@ -309,3 +315,12 @@ export const messaging = {
     dryRun: true,
   },
 };
+
+function copyAlgorithmToHeader(options: SignOptions, overrides?: object): SignOptions {
+  // The following allows us to override the alg in header without replacing the kid.
+  if (validator.isNonNullObject(overrides) && validator.isNonNullObject(options.header) &&
+    Object.prototype.hasOwnProperty.call(overrides, 'algorithm')) {
```
hiranya911, line 322: Just `overrides.hasOwnProperty()`?

package.json
```diff
@@ -77,7 +77,7 @@
     "@types/chai": "^4.0.0",
     "@types/chai-as-promised": "^7.1.0",
     "@types/firebase-token-generator": "^2.0.28",
-    "@types/jsonwebtoken": "^8.5.0",
+    "@types/jsonwebtoken": "^8.5.5",
```
hiranya911, line 80: I'm wondering if we should just downgrade and pin this dependency to v8.5.1. Implementing all these workarounds to deal with a badly versioned dependency doesn't sound very good to me. WDYT?

hiranya911 (changes requested):
Workaround looks ok with a couple of suggestions. But I'm wondering if we should downgrade the faulty dependency instead.

package.json
lahirumaramba: Agreed. Created a new PR to pin the dependency. I will close this for now.

lahirumaramba:
Decided to pin the dependency to 8.5.1 for now #1438. Closing this PR.

Issue #1434, opened 2021-09-14
Error in firebase-admin dependency: '@firebase/app' does not contain a default export (imported as 'firebase').


taletski:
### Environment

  * Operating System version: macOS Big Sur 11.5.1 (20G80)
  * Firebase SDK version: 9.0.2
  * Firebase Product: database
  * Node.js version: 16.7.0
  * NPM version: 7.20.3
  * Yarn version: 1.22.11

### The problem

After the installation to the Next.js (v10.0.5) environment the project compiling fails with an error:
```
error - ./node_modules/firebase-admin/node_modules/@firebase/database/dist/index.esm.js
Attempted import error: '@firebase/app' does not contain a default export (imported as 'firebase').
```

I was only able to 'fix' it by changing `import firebase from '@firebase/app'` to `import * as firebase from '@firebase/app'` in `./node_modules/firebase-admin/node_modules/@firebase/database/dist/index.esm.js`.

Is that the `firebase-admin` obsolete dependency issue?

#### Steps to reproduce:

1. Install the firebase-admin to the Next.js project: `yarn add firebase-admin`;
2. Run next dev server: `yarn dev`;
3. See that it won't compile because of the following error:
```
error - ./node_modules/firebase-admin/node_modules/@firebase/database/dist/index.esm.js
Attempted import error: '@firebase/app' does not contain a default export (imported as 'firebase').
```

hiranya911:
@Feiyang1 do you have any idea what this is about?

Feiyang1:
We accidentally removed the export that the admin SDK uses in v9, and removed the condition that prevent `@firebase/app` from being evaluated when used in the admin SDK. I will add them back. 

Before the fix is released, the admin SDK has to use `@firebase/database` version `0.11.x` or below. You will need to manually remove `/node_modules/firebase-admin/node_modules/@firebase/database`, and install a compatible `@firebase/database` version. It can become complicated if you are also using the `firebase` package because of package hoisting. 

I can make an emergency release, so it may be better to just wait for the fix in the next `firebase` release.

Alex-Anev:
> We accidentally removed the export that the admin SDK uses in v9, and removed the condition that prevent `@firebase/app` from being evaluated when used in the admin SDK. I will add them back.
> 
[...]

PR #1437, opened 2021-09-22
fix(rtdb): Changed admin.database to use database-compat package

hiranya911:
The `@firebase/database` package now exposes the modular API since the v9 JS SDK release. Admin SDK should upgrade to the latest dependencies, using the `@firebase/database-compat` package as the new intermediary. @Feiyang1 is working on some changes in the JS SDK side to make this work smoothly.

RELEASE NOTE: Refactored the `admin.database()` API to use the new `@firebase/database-compat` package under the hood. This helps avoid several dependency conflicts, and type compilation errors when using Firebase Web SDK packages alongside the Admin SDK.

### DIFF
--- a/src/database/database-internal.ts
+++ b/src/database/database-internal.ts
@@ -19,7 +19,7 @@ import * as path from 'path';
 
 import { FirebaseApp } from '../firebase-app';
 import { FirebaseDatabaseError, AppErrorCodes, FirebaseAppError } from '../utils/error';
-import { Database as DatabaseImpl } from '@firebase/database-compat';
+import { Database as DatabaseImpl } from '@firebase/database-compat/standalone';
 import { database } from './index';
 
 import * as validator from '../utils/validator';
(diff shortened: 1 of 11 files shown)


hiranya911:
Resolves #1434

Feiyang1 (approved):
LGTM. Remember to replace the prerelease version numbers with the main version numbers.

src/database/database-internal.ts
```diff
@@ -89,7 +89,8 @@ export class DatabaseService {
 
     let db: Database = this.databases[dbUrl];
     if (typeof db === 'undefined') {
-      const rtdb = require('@firebase/database'); // eslint-disable-line @typescript-eslint/no-var-requires
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const rtdb = require('@firebase/database-compat/standalone');
```
Feiyang1, line 93: Can you just import `initStandalone` at the top of the file instead of using require here?

Feiyang1 (approved):

src/database/database-internal.ts
hiranya911: I believe this was necessary to defer loading RTDB dependencies until needed. The `database-internal` module is imported by the `firebase-app` module, which gets loaded pretty early on during `initializeApp()`. So we had to defer importing any values from RTDB till later. Looking at the latest code, however, I think it's safe to change this. I see that `firebase-app` now only imports a type from `database-internal`. But I'd prefer to not make any such changes at this point.

Once we switched to our modularized implementation, this will be lot easier to implement.

lahirumaramba (approved):

PR #1439, opened 2021-09-24
fix(fac): Change the jwks cache duration from 1 day to 6 hours

lahirumaramba:
- Change the jwks cache duration (used by the verify token API) from 1 day to 6 hours.
- b/200973487

### DIFF
--- a/test/unit/utils/jwt.spec.ts
+++ b/test/unit/utils/jwt.spec.ts
@@ -33,7 +33,7 @@ import {
 const expect = chai.expect;
 
 const ONE_HOUR_IN_SECONDS = 60 * 60;
-const ONE_DAY_IN_SECONDS = 86400;
+const SIX_HOURS_IN_SECONDS = ONE_HOUR_IN_SECONDS * 6;
 const publicCertPath = '/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com';
 const jwksPath = '/v1alpha/jwks';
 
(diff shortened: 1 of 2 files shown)


test/unit/utils/jwt.spec.ts
```diff
@@ -709,24 +709,24 @@ describe('JwksFetcher', () => {
 
       return keyFetcher.fetchPublicKeys().then(() => {
         expect(https.request).to.have.been.calledOnce;
-        clock!.tick((ONE_DAY_IN_SECONDS - 1) * 1000);
+        clock!.tick((SIX_HOURS_IN_SECONDS - 1) * 1000);
```
hiranya911, line 712: Time math is a bit strange here. It seems we are running assertions at:

00:00:00 (start; calledOnce)
05:59:59 (calledOnce)
11:59:59 (calledTwice)
17:59:58 (calledTwice)
23:59:58 (calledThrice)

Not really a problem. But just wanted to point it out.

test/unit/utils/jwt.spec.ts
```diff
@@ -709,24 +709,24 @@ describe('JwksFetcher', () => {
...
         return keyFetcher.fetchPublicKeys();
       }).then(() => {
         expect(https.request).to.have.been.calledOnce;
-        clock!.tick(ONE_DAY_IN_SECONDS * 1000); // 24 hours in milliseconds
+        clock!.tick(SIX_HOURS_IN_SECONDS * 1000); // 6 hours in milliseconds
```
hiranya911, line 716: I don't think the assertion on clock (`clock!`) is necessary here.

hiranya911 (approved):
LGTM with a couple of comments.

avolkovi (approved):

weixifan (approved):
Thanks Lahiru!

test/unit/utils/jwt.spec.ts
lahirumaramba: We are currently checking the boundaries at 6 hour steps. Should this be cleaned up a bit?

lahirumaramba: We could also check, 0, 5.59, 6, 11.59, 12, 17.59, 18 etc.

test/unit/utils/jwt.spec.ts
lahirumaramba: We need it because of the way `clock` is defined in these tests: `let clock: sinon.SinonFakeTimers | undefined;`
We set it to `undefined` after each test case:
```
  afterEach(() => {
    if (clock) {
      clock.restore();
      clock = undefined;
    }
  });
```
I think just doing `clock.restore();` should be enough here. Let me know if you would like me to clean up all the test cases.

hiranya911: Does the compiler actually throw for these instances? We initialize clock as follows in some these tests:

```
clock = sinon.useFakeTimers(1000);
```

After that compiler should see that `clock` is no longer `undefined`.

test/unit/utils/jwt.spec.ts
hiranya911: That's kind of what I expected to see. But it's not a major issue. Feel free to disregard or address in a future PR.

test/unit/utils/jwt.spec.ts
lahirumaramba: Yeah, even after initializing the compiler throws if the instance is used inside a promise chain.

PR #1444, opened 2021-09-30
fix: Remove CR (\r) from generated content

hiranya911:
Our API doc post processing script currently writes line endings with `\r\n`. Apparently devsite doesn't like that. Just using `\n` which seems to be acceptable to the devsite publishing tools.


egilmorez (approved):
Thanks!

lahirumaramba (approved):
LGTM! Thanks!

PR #1489, opened 2021-11-07
Update link typo  in base-auth.ts

cyrrill:
Updates a link in function documentation for `updateUser` in /auth/base-auth.ts


lahirumaramba:
Thank you @cyrrill!
It looks like this has been fixed in #1501. Closing this PR because of that.

PR #1540, opened 2022-01-12
chore: Update ESLint and @typescript-eslint dependencies

lahirumaramba:
This is the first PR of a series of changes to update `@google-cloud/firestore` dependency to 5.x.

`@google-cloud/firestore` 5.x requires typescript 4.x, which breaks the current `typescript-eslint` versions we use. An immediate workaround is to replace tuple types with arrays. The proper fix is to update `eslint` and `typescript-eslint` dependencies.

We have decided to fix this properly by updating the dependencies. We are addressing this in multiple PRs and as the first step this change handles the `eslint` update.

- Updated `eslint` to `^7.1.0`, `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` to `^5.0.0`.
- In ESlint 7 `camelcase` check [has been replaced](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/naming-convention.md) with `@typescript-eslint/naming-convention`. Updated the config to reflect this change and removed the previously used `disable camelcase` rules.
- The new `@typescript-eslint/naming-convention` works quite differently from the previous `@typescript-eslint/camelcase` check. I disabled a few checks for now and we will enable them fixing the errors as we go in the future.


test/unit/remote-config/remote-config-api-client.spec.ts
```diff
@@ -664,6 +661,7 @@ describe('RemoteConfigApiClient', () => {
     });
   });
 
+  // eslint-disable-next-line @typescript-eslint/ban-types
```
hsubox76, line 664: Plans to fix these two in the future? (either by specifying the function signature or allowing `Function`?)

lahirumaramba: Yeah was going to keep it for a future PR but turns out it was an easy fix. Updated with a function signature. Thanks!

hsubox76 (approved):

Issue #1466, opened 2021-10-15
Firebase does not set projectID property when setting up Storage object using application default credentials


guibirow:
### [READ] Step 1: Are you in the right place?
yes

### [REQUIRED] Step 2: Describe your environment

  * Operating System version: MacOs BigSur
  * Firebase SDK version: "firebase-admin”: “^9.5.0"
  * Firebase Product: storage
  * Node.js version: 14
  * NPM version: _____

### [REQUIRED] Step 3: Describe the problem

When using Firebase admin sdk locally and authenticated with personal account and application default credentials, 
the storage api does not work properly because the project Id provided in the Firebase application Initialisation is not forwarded to the Auth library used by the storage, causing an error because the project it was not found.

#### Steps to reproduce:

- login with personal account using `gcloud auth login --update-adc`
- create a storage object from the app
- try create a file in the bucket
- an error will be thrown

#### Relevant Code:
```
const app = admin.initializeApp({
    projectId: "abc",
    storageBucket: "def",
});
const storage = app.storage();
const bucket = storage.bucket("fghij");
bucket.file("filename").save("something")

Error:
(node:2750) UnhandledPromiseRejectionWarning: Error: Unable to detect a Project Id in the current environment.

```
When debugging the code above, is visible that the projectId is nor set for `storage.storageClient.projectId` and `storage.storageClient.authClient._cachedProjectId`

#### Workaround:

Set the project Id into internal objects manually
```
const storage = admin.storage();

//dodgy workaround to make it work
storage.storageClient.projectId = "abc";
storage.storageClient.authClient._cachedProjectId = "abc"; //this is the required one

const bucket = storage.bucket("fghij");
bucket.file("filename").save("something")
```

# Updated workaround

Passing the project id via [environment variable](https://github.com/googleapis/google-auth-library-nodejs/blob/e69308f58c3972b71326ba48ad5d3f1d94200c69/src/auth/googleauth.ts#L194) `GOOGLE_CLOUD_PROJECT` did the trick, but ideally I would expect it to use the information I passed in the `initializeApp()` method

guibirow:
I suspect this issue might be related to https://github.com/googleapis/google-auth-library-nodejs/issues/755

hiranya911:
Just tested the following code, and it worked fine:

```js
const admin = require('firebase-admin');

const app = admin.initializeApp({
  projectId: "*******",
});

const storage = app.storage();
const bucket = storage.bucket("******.appspot.com");
[...]

PR #1567, opened 2022-02-04
Fixed storage ignoring projectId issue

AMFTech512:
### Discussion

In the following piece of code:
```
initializeApp({ projectId: 'my-project-id` })
```

The library should initialize the app with the specified project id. However, when using the storage library, such as with this example:
```
getStorage().bucket().upload(filePath)
```
The application throws the following error:
```
Error: Unable to detect a Project Id in the current environment. 
To learn more about authentication and Google APIs, visit: 
https://cloud.google.com/docs/authentication/getting-started
```

This issue has been mentioned in https://github.com/firebase/firebase-admin-node/issues/1466.

This PR fixes the issue by setting the `process.env.GCLOUD_PROJECT` variable to the value passed in to `initializeApp`.

### Testing

All unit tests pass after making the changes in this PR.
Integration tests were not performed.

### DIFF
--- a/src/storage/storage.ts
+++ b/src/storage/storage.ts
@@ -45,6 +45,9 @@ export class Storage {
       });
     }
 
+    // make sure to pass along `projectId` from when the user calls `initializeApp`
+    if (app.options.projectId) process.env.GCLOUD_PROJECT = app.options.projectId;
+
     if (!process.env.STORAGE_EMULATOR_HOST && process.env.FIREBASE_STORAGE_EMULATOR_HOST) {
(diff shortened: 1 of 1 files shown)


lahirumaramba:
Hi @AMFTech512 The Admin SDK initializes the storage client with the project id set in `app` ([ref](https://github.com/firebase/firebase-admin-node/blob/b0a7434e2d9f7bca199655292351f791e28b14ea/src/storage/storage.ts#L79)).
I don't think the right approach here is to set the environment variable. The SDK probably shouldn't change/set the environment variable at all.

Let's use #1466 to discuss this. Thanks!

PR #1593, opened 2022-03-03
get/update project config on recaptcha level

Xiaoshouzi-gh:
Implemented getProjectConfig and updateProjectConfig on recaptcha level.

### Discussion

### Testing

### API Changes

    can discuss it together.


src/utils/error.ts
```diff
@@ -729,6 +729,14 @@ export class AuthClientErrorCode {
     code: 'user-not-disabled',
     message: 'The user must be disabled in order to bulk delete it (or you must pass force=true).',
   };
+  public static INVALID_RECAPTCHA_ACTION = {
+    code: 'invalid-recaptcha-action',
+    message: 'ReCAPTCHA must be "BLOCK".'
```
renkelvin, line 734: I think we should use "reCAPTCHA" here and elsewhere.

Also, the message should be "reCAPTCHA **action** must be "BLOCK"."

src/auth/auth-api-request.ts
```diff
@@ -2035,7 +2058,7 @@ const CREATE_TENANT = new ApiSettings('/tenants', 'POST')
  */
 export class AuthRequestHandler extends AbstractAuthRequestHandler {
 
-  protected readonly tenantMgmtResourceBuilder: AuthResourceUrlBuilder;
+  protected readonly v2ResourceBuilder: AuthResourceUrlBuilder;
```
renkelvin, line 2061: What's the benefit changing the name here?

src/auth/project-config.ts
```diff
@@ -0,0 +1,124 @@
...
+ */
+import * as validator from '../utils/validator';
+import { AuthClientErrorCode, FirebaseAuthError } from '../utils/error';
+import {
+  RecaptchaConfig,
+  RecaptchaConfigAuth,
+} from './auth-config';
+
+/**
+ * Interface representing the properties to update on the provided tenant.
```
renkelvin, line 24: config

src/auth/project-config-manager.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+  /**
+   * Updates an existing project configuration.
+   *
+   * @param projectConfigOptions - The properties to update on the project.
+   *
+   * @returns A promise fulfilled with the update project config.
+   */
+  public updateProjectConfig(projectConfigOptions: UpdateProjectConfigRequest): Promise<ProjectConfig> {
+    return this.authRequestHandler.updateConfig(projectConfigOptions)
```
renkelvin, line 66: Shall we update all the names to `updatePojectConfig`, config itself here is not informative enough

Xiaoshouzi-gh: Done

src/auth/project-config.ts
Xiaoshouzi-gh: Done

src/auth/auth-api-request.ts
Xiaoshouzi-gh: seems like authRequestHandler works with all V2 apis, it will be duplicate to implemented another authHandler.  With the naming `tenantMgmtResourceBuilder` it seems narrow down it's usage. Hence, renaming to be used by getProjectConfig/updateProjectConfig APIs.  For a specific tenant level access, there is another authHandler called `TenantAwareAuthRequestHandler` which left un-touched.

src/utils/error.ts
Xiaoshouzi-gh: Done

src/auth/auth-api-request.ts
renkelvin: Let's call it `authResourceUrlBuilder` then, so that we don't need to update the name when we starts using v3

renkelvin: Please update the comments of the class as well

PR #1599, opened 2022-03-09
Recapcha integ test

Xiaoshouzi-gh:
Integration test for UpdateTenant, getProjectConfig, updateProjectConfig on reCAPTCHA config.
Tests ran passed against a real project.

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/test/integration/auth.spec.ts
+++ b/test/integration/auth.spec.ts
@@ -31,7 +31,7 @@ import { deepExtend, deepCopy } from '../../src/utils/deep-copy';
 import {
   AuthProviderConfig, CreateTenantRequest, DeleteUsersResult, PhoneMultiFactorInfo,
   TenantAwareAuth, UpdatePhoneMultiFactorInfoRequest, UpdateTenantRequest, UserImportOptions,
-  UserImportRecord, UserRecord, getAuth,
+  UserImportRecord, UserRecord, getAuth, UpdateProjectConfigRequest,
 } from '../../lib/auth/index';
 
 const chalk = require('chalk'); // eslint-disable-line @typescript-eslint/no-var-requires
@@ -1154,6 +1154,55 @@ describe('admin.auth', () => {
     });
   });
 
+  describe('Project config management operations', () => {
+    before(function() {
+      if (authEmulatorHost) {
+        this.skip(); // getConfig is not supported in Auth Emulator
+      }
+    });
+    const projectOption1: UpdateProjectConfigRequest = {
+      recaptchaConfig: {
+        emailPasswordEnforcementState:  'AUDIT',
+        managedRules: [{endScore: 0.1, action: 'BLOCK'}],
+      }
+    }
+    const projectOption2: UpdateProjectConfigRequest = {
+      recaptchaConfig: {
+        emailPasswordEnforcementState:  'OFF',
+      }
+    }
+    const expectedProjectConfig1: any = {
+      recaptchaConfig: {
+        emailPasswordEnforcementState:  'AUDIT',
+        managedRules: [{endScore: 0.1, action: 'BLOCK'}],
+      }
+    }
+    const expectedProjectConfig2: any = {
+      recaptchaConfig: {
+        emailPasswordEnforcementState:  'OFF',
+        managedRules: [{endScore: 0.1, action: 'BLOCK'}],
+      }
+    }
+    it('updateProjectConfig() should resolve with the updated project config', () => {
+      return getAuth().projectConfigManager().updateProjectConfig(projectOption1)
+        .then((actualProjectConfig) => {
+          expect(actualProjectConfig.toJSON()).to.deep.equal(expectedProjectConfig1);
+          return getAuth().projectConfigManager().updateProjectConfig(projectOption2);
+        })
+        .then((actualProjectConfig) => {
+          expect(actualProjectConfig.toJSON()).to.deep.equal(expectedProjectConfig2);
+        });
+    });
+    it('getProjectConfig() should resolve with expected project config', () => {
+      return getAuth().projectConfigManager().getProjectConfig()
+        .then((actualConfig) => {
+          const actualConfigObj = actualConfig.toJSON();
+          expect(actualConfigObj).to.deep.equal(expectedProjectConfig2);
+        });
+    });
+
+  });
+
   describe('Tenant management operations', () => {
     let createdTenantId: string;
     const createdTenants: string[] = [];
@@ -1210,6 +1259,13 @@ describe('admin.auth', () => {
       testPhoneNumbers: {
         '+16505551234': '123456',
       },
+      recaptchaConfig: {
+        emailPasswordEnforcementState:  'AUDIT',
+        managedRules: [{
+          endScore: 0.3,
+          action: 'BLOCK'
+        }]
+      }
     };
     const expectedUpdatedTenant2: any = {
       displayName: 'testTenantUpdated',
@@ -1222,6 +1278,13 @@ describe('admin.auth', () => {
         state: 'ENABLED',
         factorIds: ['phone'],
       },
+      recaptchaConfig: {
+        emailPasswordEnforcementState:  'OFF',
+        managedRules: [{
+          endScore: 0.3,
+          action: 'BLOCK'
+        }]
+      }
     };
 
     // https://mochajs.org/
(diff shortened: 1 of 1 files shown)


lahirumaramba:
@Xiaoshouzi-gh is the BE API ready to test these? If so I can trigger the integration tests on this PR. Thanks!

Xiaoshouzi-gh:
> @Xiaoshouzi-gh is the BE API ready to test these? If so I can trigger the integration tests on this PR. Thanks!

The Prod endpoint is not ready.  I used staging endpoint earlier for testing.

@lahirumaramba

test/integration/auth.spec.ts
```diff
@@ -1154,6 +1154,55 @@ describe('admin.auth', () => {
     });
   });
 
+  describe('Project config management operations', () => {
+    before(function() {
+      if (authEmulatorHost) {
+        this.skip(); // getConfig is not supported in Auth Emulator
+      }
+    });
+    const projectOption1: UpdateProjectConfigRequest = {
```
lahirumaramba, line 1163: Nit: would `projectConfig1` be a better name for this?

test/integration/auth.spec.ts
```diff
@@ -1154,6 +1154,55 @@ describe('admin.auth', () => {
...
+      recaptchaConfig: {
+        emailPasswordEnforcementState:  'AUDIT',
+        managedRules: [{ endScore: 0.1, action: 'BLOCK' }],
+      }
```
lahirumaramba, line 1167:
```suggestion
      },
```

test/integration/auth.spec.ts
```diff
@@ -1154,6 +1154,55 @@ describe('admin.auth', () => {
...
+    }
```
lahirumaramba, line 1168:
```suggestion
    };
```

test/integration/auth.spec.ts
```diff
@@ -1154,6 +1154,55 @@ describe('admin.auth', () => {
...
+    const projectOption2: UpdateProjectConfigRequest = {
+      recaptchaConfig: {
+        emailPasswordEnforcementState:  'OFF',
+      }
+    }
```
lahirumaramba, line 1173: Nit: please add a trailing `,` and a `;` here and below.

test/integration/auth.spec.ts
```diff
@@ -1154,6 +1154,55 @@ describe('admin.auth', () => {
...
+      return getAuth().projectConfigManager().updateProjectConfig(projectOption1)
+        .then((actualProjectConfig) => {
+          expect(actualProjectConfig.toJSON()).to.deep.equal(expectedProjectConfig1);
+          return getAuth().projectConfigManager().updateProjectConfig(projectOption2);
+        })
+        .then((actualProjectConfig) => {
+          expect(actualProjectConfig.toJSON()).to.deep.equal(expectedProjectConfig2);
+        });
+    });
+    it('getProjectConfig() should resolve with expected project config', () => {
```
lahirumaramba, line 1196:
```suggestion

    it('getProjectConfig() should resolve with expected project config', () => {
```

test/integration/auth.spec.ts
```diff
@@ -1154,6 +1154,55 @@ describe('admin.auth', () => {
...
+      return getAuth().projectConfigManager().getProjectConfig()
+        .then((actualConfig) => {
+          const actualConfigObj = actualConfig.toJSON();
+          expect(actualConfigObj).to.deep.equal(expectedProjectConfig2);
```
lahirumaramba, line 1200: Can we guarantee that this test always gets executed after the previous one? Otherwise, you might want to update the config in a `before` block.

test/integration/auth.spec.ts
```diff
@@ -1154,6 +1154,55 @@ describe('admin.auth', () => {
...
+        });
+    });
+
```
lahirumaramba, line 1203:
```suggestion
```

test/integration/auth.spec.ts
```diff
@@ -1210,6 +1259,13 @@ describe('admin.auth', () => {
       testPhoneNumbers: {
         '+16505551234': '123456',
       },
+      recaptchaConfig: {
+        emailPasswordEnforcementState:  'AUDIT',
+        managedRules: [{
+          endScore: 0.3,
+          action: 'BLOCK'
+        }]
```
lahirumaramba, line 1267: Super minor thing, but I think this should be formatted as follows:
```
       managedRules: [
          {
            endScore: 0.3,
            action: 'BLOCK',
          },
      ],
```

test/integration/auth.spec.ts
```diff
@@ -1210,6 +1259,13 @@ describe('admin.auth', () => {
...
+      }
```
lahirumaramba, line 1268:
```suggestion
      },
```

test/integration/auth.spec.ts
```diff
@@ -1222,6 +1278,13 @@ describe('admin.auth', () => {
         state: 'ENABLED',
         factorIds: ['phone'],
       },
+      recaptchaConfig: {
+        emailPasswordEnforcementState:  'OFF',
+        managedRules: [{
+          endScore: 0.3,
+          action: 'BLOCK'
+        }]
```
lahirumaramba, line 1286: Please see the comment above on formatting values in an array.

test/integration/auth.spec.ts
```diff
@@ -1222,6 +1278,13 @@ describe('admin.auth', () => {
...
+      }
```
lahirumaramba, line 1287:
```suggestion
      },
```

test/integration/auth.spec.ts
```diff
@@ -1643,6 +1707,7 @@ describe('admin.auth', () => {
         multiFactorConfig: deepCopy(expectedUpdatedTenant2.multiFactorConfig),
         // Test clearing of phone numbers.
         testPhoneNumbers: null,
+        recaptchaConfig: deepCopy(expectedUpdatedTenant2.recaptchaConfig),
```
lahirumaramba, line 1710: Does updating a config with an `undefined` `recaptchaConfig` deletes the current config? Do we need to add a test (unit or integration) for this scenario?

lahirumaramba (changes requested):
Thanks! Added a few comments.

test/integration/auth.spec.ts
Xiaoshouzi-gh: Updated

test/integration/auth.spec.ts
Xiaoshouzi-gh: Mocha runs tests in serial unless indicates otherwise by using `--parallel` parameter.

test/integration/auth.spec.ts
Xiaoshouzi-gh: Good point.  With recaptchaConfig being undefined, the current config will stay unchanged.  Added integration test for updateTenants.  As of project config, since it's the only config, it got covered by unit test.

test/integration/auth.spec.ts
lahirumaramba: Great! Thanks.

test/integration/auth.spec.ts
```diff
@@ -1672,6 +1742,34 @@ describe('admin.auth', () => {
...
+    it('updateTenant() should not update tenant reCAPTCHA config is undefined', () => {
+      expectedUpdatedTenant.tenantId = createdTenantId;
+      const updatedOptions2: UpdateTenantRequest = {
+        emailSignInConfig: {
+          enabled: true,
+          passwordRequired: false,
+        },
+        multiFactorConfig: deepCopy(expectedUpdatedTenant2.multiFactorConfig),
+        // Test clearing of phone numbers.
+        testPhoneNumbers: null,
```
lahirumaramba, line 1754: I feel like this is testing multiple things. Would it make more sense to just limit this test case only for `recaptchaConfig: undefined` ?

test/integration/auth.spec.ts
```diff
@@ -1672,6 +1742,34 @@ describe('admin.auth', () => {
...
+        recaptchaConfig: undefined,
```
lahirumaramba, line 1755: `recaptchaConfig: null` would throw a runtime error, correct?

lahirumaramba (approved):
Thank you! LGTM!
Left a couple minor comments.

renkelvin (approved):

test/integration/auth.spec.ts
Xiaoshouzi-gh: Yes.

test/integration/auth.spec.ts
Xiaoshouzi-gh: I kept only recaptchaConfig and an extra displayName in the new version so that the request data sent is not empty. Let me know your thought on this.

PR #1740, opened 2022-06-03
chore: Unpin @types/jsonwebtoken

lahirumaramba:
- Fixes #1723


lahirumaramba:
Looks like the issue is now fixed in https://github.com/firebase/firebase-admin-node/pull/1747

Issue #1726, opened 2022-05-25
Invalid type declaration after upgrading to TS 4.7


DebkanchanSamadder:
### [READ] Step 1: Are you in the right place?

  * For issues related to __the code in this repository__ file a Github issue.
  * If the issue pertains to Cloud Firestore, read the instructions in the "Firestore issue"
    template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Operating System version: MacOS Monterey 12.3.1
  * Firebase SDK version: 10.2.0
  * Firebase Product: app, firestore
  * Node.js version: 16
  * NPM version: 8.5.5

#### Steps to reproduce:

After upgrading to Typescript 4.7, `tsc` throws the following error for all firebase admin SDK imports

```bash
error TS7016: Could not find a declaration file for module 'firebase-admin/firestore'. '/Users/debkanchan/Library/Mobile Documents/com~apple~CloudDocs/Work/Ride/Code/Backend/services/wallet-service/node_modules/firebase-admin/lib/esm/firestore/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/firebase-admin` if it exists or add a new declaration (.d.ts) file containing `declare module 'firebase-admin/firestore';`

2 import { Firestore, FieldValue, getFirestore } from "firebase-admin/firestore";
```

#### Relevant Code:
tsconfig.json

```jsonc
{
	"extends": "@tsconfig/node16/tsconfig.json",
	"compilerOptions": {
		"module": "Node16",
		"outDir": "build",
		"removeComments": true
		// "noImplicitAny": false
	},
	"compileOnSave": true,
	"include": ["src"]
}
```

package.json
```jsonc
{
	"type": "module",
	"engines": {
		"node": "16"
	},
       ...
	"dependencies": {
		"firebase-admin": "^10.2.0",
                ...
	},
	"devDependencies": {
		"@tsconfig/node16": "^1.0.2",
                 ...
		"ts-node": "^10.8.0",
		"typescript": "^4.7.2"
	}
}

```

lahirumaramba:
[...]

Issue #1718, opened 2022-05-20
firebase-admin snyk issues


harshagarwal00:
Snyk is throwing a warning how to fix it ?
Link: https://snyk.io/test/npm/firebase-admin/10.2.0

node: v16.14.2
npm: 8.5.0

harshagarwal00:
This is duplicate of https://github.com/firebase/firebase-admin-node/issues/1512
But needs a fix.. cant believe above issue got ignored for 5 months :)

lahirumaramba:
Thank you @harshagarwal00 for reporting this issue.
The Admin Node.js SDK uses `dicer` to parse multipart responses from Firebase and GCP backend servers.
https://github.com/firebase/firebase-admin-node/blob/94dd7c3efb9ff00b0462cf772b803d6abecc2dcc/src/utils/api-request.ts#L592
IIUC the threat introduced by the vulnerability in `dicer` is pretty low here as we can trust the responses from the BE servers. However, if you are using the `dicer` package to parse responses in your own code then you might be at a higher risk. Having said that, we are looking into the fixes mentioned in #1512 and the potential to use the fixed version of dicer in Admin SDK. I will use this issue to update the progress.

fbartho:
Thanks @lahirumaramba -- I recognize this is a low vulnerability point, but given that Github/Dependabot are alerting, it'd be great to address this issue permanently.

Corrob:
Another reason to fix this quickly is that some people block our releases on "npm audit" results to make sure we don't have high vulnerabilities in our release build. Currently, our releases are blocked or must be worked around until firebase-admin stops being reported as being with a high vulnerability in the audit.

samarpanB:
Rightly said @Corrob . We are also facing the same issue. We are blocked for deployment on production environment because of `npm audit` failing and snyk also failing. Reason is this issue only. How soon can we fix it ?

harshagarwal00:
@lahirumaramba : brilliant but its not tagged in release... by when will we get that :) 
waiting to import the fix
thanks!!!

lahirumaramba:
This should be fixed in v10.3.0. Thanks everyone for you patience while we were working on the fix.

Diribar:
The alert is still showng, both in Github and nodejs. Do you have any news about the solution?

ellie-ochieno:
[...]

Issue #1512, opened 2021-12-05
Security Issues in dicer package


Uzlopak:
Hi,

we forked busboy and fixed two critical bugs in the package, which could cause the node-process to crash or to hang. Those bugs originated in the dicer package, which we integrated in our busboy fork and, as I already mentioned, fixed. Is there any interest in a separate @fastify/dicer package, were we only ship the bugfixed dicer package?

for tracking reasons:
fastify/busboy#68

lahirumaramba:
Hi @Uzlopak Thank you for reaching out. Apologies for not responding to this earlier. Do you know id `fastify` is still interested in shipping a bug fixed dicer package?

harshagarwal00:
@lahirumaramba :
can we use busyboy instead do see comments on : https://github.com/mscdex/dicer/pull/22
As mentioned we can use fastify/busyboy : https://github.com/fastify/fastify-multipart/issues/297
link to fastify/busyboy: https://github.com/fastify/busboy

if we cant use busboy 
then we can ask fastify/busboy team to fork a dicer for use... as mentioned in PRs they seem open to it..

samarpanB:
We are also facing the same issue. We are blocked for deployment on production environment because of `npm audit` failing and snyk also failing. Reason is this issue only. How soon can we fix it ?

Uzlopak:
We did not plan to create a separate fork of dicer. But I could export dicer class from our busboy-fork.

lahirumaramba:
@Uzlopak thank you! I think that would the fastest way to address this issue. Do you know what your current release schedule is? I am also looking into replacing `dicer` with `busyboy`, though I am not sure if that would be an overkill for the current use case we have in the admin sdk.

Uzlopak:
@kibertoad
Should we?

samschurter:
We are running into this security warning on several of our repos. Thanks for working on this!

rafaelmaeuer:
any news on this?

Uzlopak:
Exports dicer from fastify busboy

https://github.com/fastify/busboy/pull/90

As busboy is already default import, it would be necessary to use the named export Dicer

samschurter:
@lahirumaramba Now that there is a fixed fork of Dicer available, do you need someone to open a PR with a fix here or are you planning on doing that?

kibertoad:
will do tomorrow

lahirumaramba:
Thank you @Uzlopak , @kibertoad !
[...]

PR #1759, opened 2022-06-09
[chore] Release 10.3.0

lahirumaramba:
- Includes fixes for #1512, #1718, and #1726
- Node.js 12 support is deprecated. We strongly encourage you to use Node.js 14 or higher as we will drop support for Node.js 12 in the next major version. We will also upgrade Typescript to 4.x in the next major version.


egilmorez (approved):
LG, thanks!

Issue #1198, opened 2021-03-17
securityRules() methods to release rulesets cannot be used on clean project


wvanderdeijl:
### Describe your environment

  * Operating System version: MacOS 11.2.3
  * Firebase SDK version: 9.4.2
  * Firebase Product: securityRules
  * Node.js version: 14.15.4
  * NPM version: 6.14.10

### Describe the problem

The `securityRules` within firebase-admin offer a number of methods to change the security rules of your firebase project. However, they fail with an error "Requested entity was not found." when invoked on a freshly created project. The problem resolves itself if you create an initial set of rules through the console or CLI.

Requested entity was not found.

#### Steps to reproduce:

We are creating our projects through infrastructure as code (terraform) and create an AppEngine that way (to enable firestore) and then enable the firebaserules.googleapis.com service. We then try to programmatically deploy rules (as there is no terraform support for rules) using firebase-admin:
```
app.securityRules().releaseFirestoreRulesetFromSource(requested.rules);
```

#### Relevant Code:

I feel the problem lies in
https://github.com/firebase/firebase-admin-node/blob/2f6da8925b953a116f1e44bb6a2b9213db5d2681/src/security-rules/security-rules-api-client-internal.ts#L165
That always does a PATCH assuming a ruleset already exists.

All the methods in firebase-admin.securityRules to change the ruleset eventually end up in that `updateRelease` method.

If I look at the source code of `firebase-tools` you can see it has a method to first create a release: https://github.com/firebase/firebase-tools/blob/1633f4fccbbc1bcbc6216fe13b8e888c8940bde4/src/gcp/rules.ts#L232

It would be nice if firebase-admin detects that no rules release exists yet, and create one first using a POST, before attempting the PATCH. Alternative would be to catch the "Requested entity was not found" error, then do a POST followed by a retry of the PATCH.

hiranya911:
I believe the release resource is created automatically upon enabling Firestore. But this happens asynchronously, and therefore there's no guarantee as to when it will be available. 

@rachelmyers do you have any insights on this? Does it make sense for the SDK to try and create the release if an update fails?

wvanderdeijl:
[...]

PR #1779, opened 2022-06-21
fix(rules): Add update or create release method

lahirumaramba:
If updating rules fails with a NOT_FOUND error, create a new release instead.

Fixes: #1198


hsubox76 (approved):
LG!

PR #1828, opened 2022-07-21
PR fixes from #1812

joehan:
Realized that I never actually pushed the commit containing the requested fixes from #1812 🤦


lahirumaramba (approved):
Thank You!

PR #1882, opened 2022-08-22
Support updating custom claims in updateUser()

### LINKED ISSUE #1876: [FR] Auth: Update user details and custom claims in the same request
Currently, the custome user claims can be updated with [`setCustomUserClaims()`](https://firebase.google.com/docs/reference/admin/node/firebase-admin.auth.baseauth.md#baseauthsetcustomuserclaims), and the user details can be updated with [`updateUser()`](https://firebase.google.com/docs/reference/admin/node/firebase-admin.auth.baseauth.md#baseauthupdateuser).

Both use same [`/accounts:update`](https://cloud.google.com/identity-platform/docs/reference/rest/v1/accounts/update) API under the hood.

Is there a particular reason why the `updateUser()` method does not allow updating the custom claims even though it is possible in the underlying API? A follow up point is: if the `updateUser()` method has this functionalitly, then the `setCustomUserClaims()` would become obsolete (just a convenience wrapper).

I'd like to be able to include a `customClaims` field in the [`UpdateRequest`](https://firebase.google.com/docs/reference/admin/node/firebase-admin.auth.updaterequest.md#updaterequest_interface) interface, so that, if needed, we can update the user's details and claims in a single request. Currently we need two separate network requests to do this.

Let me know, and I can help with a PR if you want.

brettwillis:
Closes #1876

### DIFF
--- a/src/auth/auth-api-request.ts
+++ b/src/auth/auth-api-request.ts
@@ -1349,7 +1349,7 @@ export abstract class AbstractAuthRequestHandler {
     if (customUserClaims === null) {
       customUserClaims = {};
     }
-    // Construct custom user attribute editting request.
+    // Construct custom user attribute editing request.
     const request: any = {
       localId: uid,
       customAttributes: JSON.stringify(customUserClaims),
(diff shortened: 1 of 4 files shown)


brettwillis:
@lahirumaramba any thoughts on this?

brettwillis:
@lahirumaramba @prameshj I completed the test suite adding integration tests and unit tests. All relevant tests that were applied to `setCustomUserClaims()` are applied to the relevant claims functionality in `updateUser()`. Would appreciate a moment to review 🙏🏼 .

Issue #1900, opened 2022-09-10
Bug: FCM - Duplicate Topic Notifications


roybeitner:
### [READ] Step 1: Are you in the right place?

  * For issues related to __the code in this repository__ file a Github issue.
  * If the issue pertains to Cloud Firestore, read the instructions in the "Firestore issue"
    template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Operating System version: **Any iOS or Android running Flutter 3.00 with firebase_messaging 13.0.0**
  * Firebase SDK version: **11.0.1**
  * Firebase Product: **FCM** (auth, database, storage, etc)
  * Node.js version: **v16.x AWS Lambda**
  * NPM version: **AWS Lambda**

### [REQUIRED] Step 3: Describe the problem
Android and iOS users receive at least once a day a duplicate topic notification sent from the Lambda service running the Firebase-Admin SDK.
No duplicate notifications we're generated by our service itself (used extensive logging for each notification sent).
We use the sendAll function (of the SDK) and send the notifications in batches.
The client app does not create a local notification in any part of the code.
#### Steps to reproduce:

What happened? How can we make the problem occur?
This could be a description, log/console output, etc.

Log from the client receiving a duplicate notification:
```
D/FLTFireMsgReceiver(11547): broadcast received for message
I/flutter (11547): ----------------------------
I/flutter (11547): {data, blablabla}
I/flutter (11547): 0:1662825724904819%88b82b6688b82b66
I/flutter (11547): Subtitle bla bla bla
I/flutter (11547): Title bla bla bla
I/flutter (11547): 2022-09-10 19:02:04.313
I/flutter (11547): ----------------------------
W/FirebaseMessaging(11547): Unable to log event: analytics library is missing
W/FirebaseMessaging(11547): Missing Default Notification Channel metadata in AndroidManifest. Default value will be used.
I/FirebaseMessaging(11547): Starting download of: imageurl
D/SurfaceView@bee2085(11547): updateSurface: surface is not valid
D/FLTFireMsgReceiver(11547): broadcast received for message
[...]

PR #1947, opened 2022-10-19
fix(fcm): Increase FCM timeout to 15s

lahirumaramba:
- Increase the FCM timeout to 15s to match the recommended timeout in the BE. This will prevent duplicate notifications due to the Admin SDK retrying too early.

Fixes: #1900


chong-shao:
LGTM. Thanks!

chong-shao (approved):

thopedam:
The batch request timeout is not increased. Still duplicate push notifications.

PR #1949, opened 2022-10-21
Internal implementation of multiDb

tom-andersen:


wu-hui:
please take `package-lock.json` out of this PR. Otherwise LGTM.

tom-andersen:
> please take `package-lock.json` out of this PR. Otherwise LGTM.

Done

src/firestore/firestore-internal.ts
```diff
@@ -25,14 +25,22 @@ import { App } from '../app';
...
+  private databases: Map<string, Firestore> = new Map();
 
   constructor(app: App) {
-    this.firestoreClient = initFirestore(app);
     this.appInternal = app;
   }
 
+  getDatabase(databaseId: string): Firestore {
+    let database = this.databases.get(databaseId);
+    if (database === undefined) {
```
lahirumaramba, line 37:
```suggestion
    if (typeof database === undefined) {
```

src/firestore/firestore-internal.ts
```diff
@@ -25,14 +25,22 @@ import { App } from '../app';
 
 export class FirestoreService {
 
-  private appInternal: App;
-  private firestoreClient: Firestore;
+  private readonly appInternal: App;
+  private databases: Map<string, Firestore> = new Map();
```
lahirumaramba, line 29: I think this also can be `readonly databases: Map<string, Firestore> = new Map()`

src/firestore/firestore-internal.ts
lahirumaramba: Or we could also do `if (!this.databases.has(databaseId)) {` just a suggestion really. I will leave it up to you. :)

src/firestore/index.ts
```diff
@@ -96,13 +110,29 @@ export {
  *   service if no app is provided or the `Firestore` service associated with the
  *   provided app.
  */
-export function getFirestore(app?: App): Firestore {
-  if (typeof app === 'undefined') {
-    app = getApp();
-  }
-
+export function getFirestore(app: App): Firestore;
```
lahirumaramba, line 113: nit: new line

src/firestore/index.ts
```diff
@@ -96,13 +110,29 @@ export {
...
+/**
+ * @param databaseId
+ * @internal
+ */
+export function getFirestore(databaseId: string): Firestore;
```
lahirumaramba, line 118: nit: new line

src/firestore/index.ts
```diff
@@ -96,13 +110,29 @@ export {
...
+ * @param app
+ * @param databaseId
+ * @internal
+ */
+export function getFirestore(app: App, databaseId: string): Firestore;
+export function getFirestore(
+  appOrDatabaseId?: App | string,
+  optionalDatabaseId?: string
+): Firestore {
+  const app: App = typeof appOrDatabaseId === 'object' ? appOrDatabaseId : getApp();
```
lahirumaramba, line 129: I think `type of null` is also `object`... we might want to check if `app !== null` as well. Alternatively, you can also move the implementations into the respective functions separately.

src/firestore/index.ts
```diff
@@ -96,13 +110,29 @@ export {
...
+  const databaseId =
+    typeof appOrDatabaseId === 'string'
+      ? appOrDatabaseId
+      : optionalDatabaseId || '(default)';
```
lahirumaramba, line 133: Is `'(default)'` a const?

src/firestore/index.ts
```diff
@@ -96,13 +110,29 @@ export {
...
+ * @internal
+ */
+export function getFirestore(app: App, databaseId: string): Firestore;
+export function getFirestore(
+  appOrDatabaseId?: App | string,
+  optionalDatabaseId?: string
+): Firestore {
+  const app: App = typeof appOrDatabaseId === 'object' ? appOrDatabaseId : getApp();
+  const databaseId =
+    typeof appOrDatabaseId === 'string'
```
lahirumaramba, line 131: Do we want to check if `appOrDatabaseId` is a non-empty string? We a validator util if this is needed `validator.isNonEmptyString()`.

test/unit/firestore/firestore.spec.ts
```diff
@@ -98,23 +98,25 @@ describe('Firestore', () => {
       it(`should throw given invalid app: ${ JSON.stringify(invalidApp) }`, () => {
         expect(() => {
           const firestoreAny: any = FirestoreService;
-          return new firestoreAny(invalidApp);
+          const firestoreService: FirestoreService = new firestoreAny(invalidApp);
+          return firestoreService.getDatabase('(default)')
```
lahirumaramba, line 102: Should we make this `'(default)'` a constant?

test/unit/firestore/firestore.spec.ts
```diff
@@ -158,19 +160,6 @@ describe('Firestore', () => {
     });
   });
 
-  describe('client', () => {
```
lahirumaramba, line 161: Could this be a potential breaking change in the public API?

lahirumaramba (changes requested):
Thank you @tom-andersen ! Overall LGTM!
I left a few comments and questions.

src/firestore/index.ts
```diff
@@ -96,13 +110,29 @@ export {
...
+export function getFirestore(app: App): Firestore;
+/**
+ * @param databaseId
+ * @internal
+ */
+export function getFirestore(databaseId: string): Firestore;
+/**
+ * @param app
+ * @param databaseId
+ * @internal
```
lahirumaramba, line 122: `@internal` will hide this from both reference docs and autocomplete (it will remove the typescript annotation). Is that what we expect here? If we only want to hide from reference docs and would like to keep autocomplete we can use `@alpha` instead.

src/firestore/index.ts
tom-andersen: They would need to bypass Typescript type checking for this to happen, since `appOrDatabaseId?: App | string`.

According to this doc, we should not guard against invalid types.
https://g3doc.corp.google.com/firebase/g3doc/aip/3270.md?cl=head#input-validation

src/firestore/index.ts
tom-andersen: Good point. I think we should consider an empty string equivalent to the default database id.

test/unit/firestore/firestore.spec.ts
tom-andersen: Imported from Server SDK

lahirumaramba (approved):
LGTM. Thanks!

PR #1971, opened 2022-11-09
TOTP MFA changes

pragatimodi:


src/auth/auth-config.ts
```diff
@@ -599,6 +659,42 @@ export class MultiFactorAuthConfig implements MultiFactorConfig {
...
+      if (!validator.isArray(options.providerConfigs)) {
+        throw new FirebaseAuthError(
+          AuthClientErrorCode.INVALID_CONFIG,
+          '"MultiFactorConfig.providerConfigs" must be an array of valid "MultiFactorProviderConfigs."',
+        );
+      }
+
+      //Validate content of array.
+      options.providerConfigs.forEach((multiFactorProviderConfig) => {
+        if (typeof multiFactorProviderConfig == 'undefined') {
```
prameshj, line 673: nit: ===

src/auth/auth-config.ts
```diff
@@ -623,6 +719,18 @@ export class MultiFactorAuthConfig implements MultiFactorConfig {
...
+    this.providerConfigs = [];
+    (response.providerConfigs || []).forEach((providerConfig) => {
+      if (typeof providerConfig !== 'undefined') {
+        if (typeof providerConfig.state === 'undefined' ||
+          typeof providerConfig.totpProviderConfig === 'undefined') {
+          throw new FirebaseAuthError(
+            AuthClientErrorCode.INTERNAL_ERROR,
+            'INTERNAL ASSERT FAILED: Invalid multi-factor configuration response');
+        }
+        this.providerConfigs?.push(providerConfig);
```
prameshj, line 732: Do you need "?" since you initialized in line 723?

src/auth/project-config.ts
```diff
@@ -57,6 +62,15 @@ export class ProjectConfig {
    * This is based on the calling code of the destination phone number.
    */
   public readonly smsRegionConfig?: SmsRegionConfig;
+  private readonly multiFactorConfig_?: MultiFactorConfig;
```
prameshj, line 65: don't you need multiFactorConfig in ProjectConfigServerResponse and ClientRequest?

src/auth/user-import-builder.ts
```diff
@@ -350,7 +355,21 @@ export function convertMultiFactorInfoToServerFormat(multiFactorInfo: UpdateMult
       }
     }
     return authFactorInfo;
-  } else {
+  } else if (isTotpFactor(multiFactorInfo)) {
+    const authFactorInfo: AuthFactorInfo = {
+      mfaEnrollmentId: multiFactorInfo.uid,
+      displayName: multiFactorInfo.displayName,
+      sharedSecretKey: multiFactorInfo.sharedSecretKey,
+      enrolledAt,
```
prameshj, line 363: nit: `enrolledAt: enrolledAt`?

src/auth/user-import-builder.ts
```diff
@@ -350,7 +355,21 @@ export function convertMultiFactorInfoToServerFormat(multiFactorInfo: UpdateMult
       }
     }
     return authFactorInfo;
-  } else {
+  } else if (isTotpFactor(multiFactorInfo)) {
```
prameshj, line 358: nit: since you have a return in the previous if block, this can also be if instead of else if.

src/auth/user-import-builder.ts
```diff
@@ -350,7 +355,21 @@ export function convertMultiFactorInfoToServerFormat(multiFactorInfo: UpdateMult
...
+      enrolledAt,
+    };
+    for (const objKey in authFactorInfo) {
+      if (typeof authFactorInfo[objKey] === 'undefined') {
+        delete authFactorInfo[objKey];
+      }
+    }
+    return authFactorInfo;
+  }
+  else {
```
prameshj, line 372: Same comment as above, the else keyword is optional

src/auth/user-import-builder.ts
```diff
@@ -363,6 +382,11 @@ function isPhoneFactor(multiFactorInfo: UpdateMultiFactorInfoRequest):
   return multiFactorInfo.factorId === 'phone';
 }
 
+function isTotpFactor(multiFactorInfo: UpdateMultiFactorInfoRequest):
+  multiFactorInfo is UpdateTotpMultiFactorInfoRequest {
+  return multiFactorInfo.factorId == 'totp';
```
prameshj, line 387: === ?

test/unit/auth/user-import-builder.spec.ts
```diff
@@ -825,7 +835,7 @@ describe('UserImportBuilder', () => {
                 uid: 'enrollmentId2',
                 secret: 'SECRET',
                 displayName: 'Google Authenticator on personal phone',
-                factorId: 'totp',
+                factorId: 'factor',
```
prameshj, line 838: should be totp for a valid testcase? Maybe add another user with an invalid factor id and update the display name also?

test/unit/auth/tenant.spec.ts
```diff
@@ -431,10 +460,17 @@ describe('Tenant', () => {
       expect(tenant.emailSignInConfig).to.deep.equal(expectedEmailSignInConfig);
     });
 
+    //TODO
     it('should set readonly property multiFactorConfig', () => {
       const expectedMultiFactorConfig = new MultiFactorAuthConfig({
         state: 'ENABLED',
-        enabledProviders: ['PHONE_SMS'],
+        enabledProviders: ['PHONE_SMS', 'TOTP'],
```
prameshj, line 467: As mentioned in anoher comment, TOTP might not be allowed as an "enabledProvider". We should change unit tests accordinglty.

Also can you expand on the TODO?

src/auth/auth-config.ts
pragatimodi: Yes, the compiler shows an error otherwise

PR #1974, opened 2022-11-14
change(ml): Drop the feature to publish from an AutoML Model

pavelgj:


lahirumaramba (approved):
This change will be included in the upcoming major (v12) release

PR #1989, opened 2022-11-23
feat(auth): Add TOTP support in Project and Tenant config

pragatimodi:
~**DO_NOT_MERGE** (waiting for BE changes)~

PR for TOTP Config for Project and Tenant

RELEASE NOTES: Added support for TOTP (Time-based One Time Password) as a Multi-Factor Authentication option in Tenant and Project config.


src/auth/auth-config.ts
```diff
@@ -506,6 +507,27 @@ export interface MultiFactorConfig {
    * Currently only ‘phone’ is supported.
    */
   factorIds?: AuthFactorType[];
+  /**
+   * A list of multi-factor provider specific config.
```
prameshj, line 511: Extend the comment with "new MFA providers (except phone) will indicate enablement/disablement through this field."

src/auth/auth-config.ts
```diff
@@ -599,6 +631,41 @@ export class MultiFactorAuthConfig implements MultiFactorConfig {
...
+          AuthClientErrorCode.INVALID_CONFIG,
+          '"MultiFactorConfig.providerConfigs" must be an array of valid "MultiFactorProviderConfigs."',
+        );
+      }
+      //Validate content of array.
+      options.providerConfigs.forEach((multiFactorProviderConfig) => {
+        if (typeof multiFactorProviderConfig === 'undefined') {
+          throw new FirebaseAuthError(
+            AuthClientErrorCode.INVALID_CONFIG,
+            `"${multiFactorProviderConfig}" is not a valid "MultiFactorProviderConfigType".`
```
prameshj, line 647: nit: space between MultiFactorProviderConfig and Type.

src/auth/auth-config.ts
```diff
@@ -599,6 +631,41 @@ export class MultiFactorAuthConfig implements MultiFactorConfig {
...
+          )
+        }
+        if (typeof multiFactorProviderConfig.state !== 'undefined' &&
```
prameshj, line 650: Should we instead check:

`if (typeof multiFactorProviderConfig.state === 'undefined' || ( multiFactorProviderConfig.state !== 'ENABLED' && multiFactorProviderConfig.state !== 'DISABLED'))`

src/auth/auth-config.ts
```diff
@@ -599,6 +631,41 @@ export class MultiFactorAuthConfig implements MultiFactorConfig {
...
+          multiFactorProviderConfig.state !== 'ENABLED' &&
+          multiFactorProviderConfig.state !== 'DISABLED') {
+          throw new FirebaseAuthError(
+            AuthClientErrorCode.INVALID_CONFIG,
+            '"MultiFactorConfig.providerConfigs.state" must be either "ENABLED" or "DISABLED".',
+          )
+        }
+        if (typeof multiFactorProviderConfig.totpProviderConfig !== 'undefined') {
```
prameshj, line 658: not specifying totpProviderConfig is an error right now, since we don't have any other options in MultiFactorProviderConfig.

src/auth/auth-config.ts
```diff
@@ -624,13 +691,28 @@ export class MultiFactorAuthConfig implements MultiFactorConfig {
         this.factorIds.push(AUTH_FACTOR_SERVER_TO_CLIENT_TYPE[enabledProvider]);
       }
     })
+    this.providerConfigs = [];
```
prameshj, line 694: Asking for my own clarification - this validation is for the response from the backend and the above code in "validate()" is for user/customer input, correct?

src/auth/project-config.ts
```diff
@@ -57,6 +62,13 @@ export class ProjectConfig {
    * This is based on the calling code of the destination phone number.
    */
   public readonly smsRegionConfig?: SmsRegionConfig;
+  private readonly multiFactorConfig_?: MultiFactorConfig;
```
prameshj, line 65: Don't you need this field in ProjectConfigClientRequest and ProjectConfigServerResponse?

prameshj:
Also adding @Xiaoshouzi-gh to take a look.

src/auth/auth-config.ts
```diff
@@ -506,6 +507,27 @@ export interface MultiFactorConfig {
...
+  /**
+   * TOTP MultiFactor provider config.
+   */
+  totpProviderConfig?: TotpMultiFactorProviderConfig;
+}
+
+export interface TotpMultiFactorProviderConfig {
+  /**
+   *  The allowed number of adjacent intervals that will be used for verification
+  *  to avoid clock skew.
```
Xiaoshouzi-gh, line 528: nit: add a space before *

test/unit/auth/auth-config.spec.ts
```diff
@@ -206,10 +206,26 @@ describe('MultiFactorAuthConfig', () => {
       const config = new MultiFactorAuthConfig({
         state: 'ENABLED',
         enabledProviders: ['PHONE_SMS'],
+        providerConfigs: [
```
Xiaoshouzi-gh, line 209: Do you have any tests on these field for the validation you wrote, they could be either located in auth-config.spec or auth.spec

src/auth/auth-config.ts
```diff
@@ -624,13 +691,28 @@ export class MultiFactorAuthConfig implements MultiFactorConfig {
         this.factorIds.push(AUTH_FACTOR_SERVER_TO_CLIENT_TYPE[enabledProvider]);
       }
     })
+    this.providerConfigs = [];
+    (response.providerConfigs || []).forEach((providerConfig) => {
+      if (typeof providerConfig !== 'undefined') {
+        if (typeof providerConfig.state === 'undefined' ||
+          typeof providerConfig.totpProviderConfig === 'undefined' ||
+          (typeof providerConfig.totpProviderConfig.adjacentIntervals !== 'undefined' &&
+            typeof providerConfig.totpProviderConfig.adjacentIntervals !== 'number')) {
```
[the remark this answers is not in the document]
pragatimodi: Yes, it is the same thing, just consistent with the rest of the validation in this if statement.

test/unit/auth/tenant.spec.ts
```diff
@@ -105,7 +121,7 @@ describe('Tenant', () => {
           .to.deep.equal(tenantOptionsServerRequest);
       });
 
-      it('should return the expected server request with multi-factor and phone config', () => {
+      it('should return the expected server request with multi-factor, phone and provider config', () => {
```
prameshj, line 124: nit: multi-factor and provider config are part of the same "multiFactorConfig" struct. So maybe call this:

"'should return the expected server request with multi-factor (SMS, TOTP) and testPhoneNumber config".

I assume thats what phone config means here.

test/integration/auth.spec.ts
```diff
@@ -1302,6 +1325,12 @@ describe('admin.auth', () => {
       multiFactorConfig: {
         state: 'ENABLED',
         factorIds: ['phone'],
+        providerConfigs: [
```
prameshj, line 1328: I think we need one more struct -expectedUpdatedTenantSmsEnabledTotpDisabled to cover the case where TOTP is disabled via provider config, but it preserves the phone MFA settings.

test/integration/auth.spec.ts
```diff
@@ -1247,6 +1247,13 @@ describe('admin.auth', () => {
       multiFactorConfig: {
         state: 'ENABLED',
         factorIds: ['phone'],
+        // TODO: Once createTenant() is implemented on the backend, populate providerConfigs
```
prameshj, line 1250: Shouldn't we add test cases to the project config update flow also , like line 1189?

src/auth/auth-config.ts
pragatimodi: Yes

test/unit/auth/auth-config.spec.ts
```diff
@@ -296,6 +312,148 @@ describe('MultiFactorAuthConfig', () => {
...
+          {
+            state: 'ENABLED',
+            totpProviderConfig: {
+              adjacentIntervals: 5,
+            },
+          },
+        ],
+      });
+    });
```
Xiaoshouzi-gh, line 376: Please refactor the buildServerRequest to use one base object, and deepcopy the object in each tests with needed field value updated specifically for the current test.
e.g
```const expectedRequest = { ...};
const expectedDisabledRequest = deepCopy(expectedRequest);
expectedDisabledRequest.state = 'ENABLED';
```
This way it added readability and removed redundant code.

Xiaoshouzi-gh (changes requested):

src/auth/project-config.ts
```diff
@@ -37,6 +43,7 @@ export interface UpdateProjectConfigRequest {
  */
 export interface ProjectConfigServerResponse {
   smsRegionConfig?: SmsRegionConfig;
+  mfa?: MultiFactorConfig;
```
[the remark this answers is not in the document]
pragatimodi: multiFactorConfig is approved in the API proposal but mfa is the field that corresponds to backend while sending the request on the project side.

test/integration/auth.spec.ts
```diff
@@ -1264,6 +1271,14 @@ describe('admin.auth', () => {
       multiFactorConfig: {
         state: 'ENABLED',
         factorIds: ['phone'],
+        providerConfigs: [],
+        // TODO: Once createTenant() is implemented on the backend, populate providerConfigs
```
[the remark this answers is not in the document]
pragatimodi: This is the integration test suite

src/auth/auth-config.ts
```diff
@@ -599,6 +627,48 @@ export class MultiFactorAuthConfig implements MultiFactorConfig {
         }
       });
     }
+
+    if (typeof options.providerConfigs !== 'undefined') {
+      if (!validator.isArray(options.providerConfigs)) {
```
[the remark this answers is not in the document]
pragatimodi: Empty array just means no multi factor config which is the default value.

src/auth/project-config.ts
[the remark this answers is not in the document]
pragatimodi: Updated the field to be multiFactorConfig on the user end and translate to mfa for the backend request and response.

src/auth/auth-api-request.ts
```diff
@@ -34,7 +34,7 @@ import { ActionCodeSettings, ActionCodeSettingsBuilder } from './action-code-set
 import { Tenant, TenantServerResponse, CreateTenantRequest, UpdateTenantRequest } from './tenant';
 import {
   isUidIdentifier, isEmailIdentifier, isPhoneIdentifier, isProviderIdentifier,
-  UserIdentifier, UidIdentifier, EmailIdentifier,PhoneIdentifier, ProviderIdentifier,
+  UserIdentifier, UidIdentifier, EmailIdentifier, PhoneIdentifier, ProviderIdentifier,
```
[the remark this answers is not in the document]
pragatimodi: Hiding whitespace changes for now as reverting files with whitespace changes might also revert new code

src/auth/auth-config.ts
```diff
@@ -506,6 +507,28 @@ export interface MultiFactorConfig {
...
    */
   factorIds?: AuthFactorType[];
+  /**
+   * A list of multi-factor provider specific config. 
+   * New MFA providers (except phone) will indicate enablement/disablement through this field.
+   */
+  providerConfigs?: MultiFactorProviderConfig[];
+}
+
+export interface MultiFactorProviderConfig {
```
lahirumaramba, line 517: Let's add documentation to this and all the new types below:

example see: `MultiFactorConfig`
```
/**
 * Interface representing a multi-factor configuration.
 * This can be used to define whether multi-factor authentication is enabled
 * or disabled and the list of second factor challenges that are supported.
 */
 ```

File: src/auth/auth-config.ts [hunk 19]
```diff
@@ -506,6 +507,28 @@ export interface MultiFactorConfig {
...
+  /* indicates whether this multi-factor provider is enabled/disabled. */
```
lahirumaramba, line 518: I think this might need `/*** ....` style for the documentation to work correctly.

src/auth/project-config.ts
```diff
@@ -57,13 +66,19 @@ export class ProjectConfig {
...
+   */
+  public readonly multiFactorConfig?: MultiFactorAuthConfig;
 
   /**
    * Validates a project config options object. Throws an error on failure.
    *
    * @param request - The project config options object to validate.
    */
-  private static validate(request: ProjectConfigClientRequest): void {
+  private static validate(request: any): void {
```
lahirumaramba, line 81: It is better if we can keep the strong type here. Can we stick to `ProjectConfigClientRequest` over `any`?

src/auth/project-config.ts
```diff
@@ -97,7 +118,12 @@ export class ProjectConfig {
    */
   public static buildServerRequest(configOptions: UpdateProjectConfigRequest): ProjectConfigClientRequest {
     ProjectConfig.validate(configOptions);
-    return configOptions as ProjectConfigClientRequest;
+    const request = configOptions as any;
+    if (configOptions.multiFactorConfig !== undefined) {
+      request.mfa = MultiFactorAuthConfig.buildServerRequest(configOptions.multiFactorConfig);
+    }
+    delete request.multiFactorConfig;
```
lahirumaramba, line 125: I must have missed this, but whey do we have both `multiFactorConfig` and `mfa` in the request type?

lahirumaramba (changes requested):
Thank you! Left a few comments. I will take a look at the tests next. Thanks!

src/auth/project-config.ts
pragatimodi: Because of the inconsistent naming of types between backend and SDK, we can't keep a strong type to have flexibility in converting. The same thing has been done in tenant.ts : ```private static validate(request: any, createRequest: boolean): void {```

src/auth/project-config.ts
pragatimodi: Because the backend accepts `mfa` but we have `multiFactorConfig` for the field in the SDK to maintain consistency with tenant flow. So this is just a conversion.

test/integration/auth.spec.ts
```diff
@@ -1799,6 +1883,58 @@ describe('admin.auth', () => {
...
+    it('updateTenant() should not update SMS MFA related settings when TOTP is disabled', () => {
+      expectedUpdatedTenantSmsEnabledTotpDisabled.tenantId = createdTenantId;
+      const updatedOptions3: UpdateTenantRequest = {
+        displayName: expectedUpdatedTenant2.displayName,
+        multiFactorConfig: {
+          state: 'ENABLED',
+          factorIds: ['phone'],
+          providerConfigs: [
+            {
+              state: 'DISABLED',
```
prameshj, line 1917: interesting that TOTP gets disabled with a providerConfig that does not have totpProviderConfig{} inside it.

test/integration/auth.spec.ts
```diff
@@ -1241,14 +1275,17 @@ describe('admin.auth', () => {
         })
         .then((actualProjectConfig) => {
           expect(actualProjectConfig.toJSON()).to.deep.equal(expectedProjectConfig2);
+          return getAuth().projectConfigManager().updateProjectConfig(projectConfigOptionSmsEnabledTotpDisabled);
+        }).then((actualProjectConfig) => {
+          expect(actualProjectConfig.toJSON()).to.deep.equal(expectedProjectConfigSmsEnabledTotpDisabled);
         });
     });
```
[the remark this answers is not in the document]
pragatimodi: It's not quite clear to me what you mean by this. I think the flow intends to start by updating TOTP and SMS options and then disables TOTP.

src/auth/project-config.ts
```diff
@@ -86,6 +102,11 @@ export class ProjectConfig {
     if (typeof request.smsRegionConfig !== 'undefined') {
       SmsRegionsAuthConfig.validate(request.smsRegionConfig);
     }
+
+    // Validate Multi Factor Config if provided
+    if (typeof request.mfa !== 'undefined') {
```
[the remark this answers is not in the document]
pragatimodi: Yes, thanks for catching this miss. Looks like it did not break any tests because the validation function is being tested out separately and this if statement would always be false so the code block wouldn't have been executed. The validKeys has `multiFactorConfig` as the required key in request so the `mfa` field wasn't caught there either.

src/auth/project-config.ts
```diff
@@ -57,13 +66,19 @@ export class ProjectConfig {
    * This is based on the calling code of the destination phone number.
    */
   public readonly smsRegionConfig?: SmsRegionConfig;
+  /**
+   * The Multi Factor Config for the project.
+   * Configures the multi factor settings for the project.
+   * Supports only phone and TOTP
```
prameshj, line 72: nit: missing "." at the end.

Also, let's add a comment here - "This maps to the "mfa" field in the backend request."

src/auth/project-config.ts
prameshj: Reopening since this still shows the "any" type. Can we use the UpdateProjectConfigRequest type here?

src/auth/project-config.ts
prameshj: Reopening since the comment is not here yet. IMO, it is ok to repeat the comment in multiple places.

src/auth/project-config.ts
```diff
@@ -111,6 +137,9 @@ export class ProjectConfig {
     if (typeof response.smsRegionConfig !== 'undefined') {
       this.smsRegionConfig = response.smsRegionConfig;
     }
+    if (typeof response.mfa !== 'undefined') {
```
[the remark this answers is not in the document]
prameshj: Super nit: Add a space "/** Backend..." and also no single quote at the end of the comment string.

test/integration/auth.spec.ts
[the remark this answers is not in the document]
prameshj: Sorry, i must have misread it. I see the test performs this flow as you described.

test/integration/auth.spec.ts
```diff
@@ -1801,6 +1892,59 @@ describe('admin.auth', () => {
...
+            expect(actualTenantObj).to.deep.equal(expectedUpdatedTenant2);
+          });
+      }
+      return getAuth().tenantManager().updateTenant(createdTenantId, updateRequestNoMfaConfig)
+        .then((actualTenant) => {
+          expect(actualTenant.toJSON()).to.deep.equal(expectedUpdatedTenant2);
+        });
+    });
+
+    it('updateTenant() should not update SMS MFA related settings when TOTP is disabled', () => {
```
prameshj, line 1917: nit: can you update the description to "updateTenant() should not disable SMS MFA when TOTP is disabled"

Also, it is slightly inconsistent with the project config tests - there we verify this flow as part of the same test, but we split it here. Just pointing it out, no need to fix it.

src/auth/project-config.ts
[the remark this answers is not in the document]
prameshj: Hmm.. shouldn't it have failed the tests where invalid provider config was being passed in, like the unit tests with `const invalidProviderConfigs = [NaN, 0, 1, '', 'a', {}, { a: 1 }, _.noop, true, false,]`?

prameshj (commented):
Mostly looks good to me. Some nits about rewording comments and one question about using a strong type rather than any.

src/auth/project-config.ts
prameshj: This somehow seems to be getting auto-resolved. @pragatimodi can you look into making this type change?

src/auth/project-config.ts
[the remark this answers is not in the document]
prameshj: @pragatimodi

src/auth/project-config.ts
prameshj: This doesn't reflect in the diff yet.

src/auth/project-config.ts
prameshj: @pragatimodi

test/integration/auth.spec.ts
```diff
@@ -1218,19 +1231,40 @@ describe('admin.auth', () => {
         }
       },
     };
+    const projectConfigOptionSmsEnabledTotpDisabled: UpdateProjectConfigRequest = {
+      multiFactorConfig: {
+        state: 'ENABLED',
+        factorIds: ['phone'],
+        providerConfigs: [
+          {
+            state: 'DISABLED',
```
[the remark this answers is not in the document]
prameshj: @pragatimodi the diff doesn't show totpProviderConfig still.

src/auth/project-config.ts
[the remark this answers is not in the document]
prameshj: I think a comment like `'Backend API returns "mfa" in case of project config and "multiFactorConfig" in case of tenant config. The SDK exposes it as multiFactorConfig always.'` would be useful. @pragatimodi

prameshj (commented):
Tagged you in a few comments that aren't getting reflected in the diffs yet. LGTM once those are implemented.

src/auth/project-config.ts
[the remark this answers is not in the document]
pragatimodi: Well, invalidProviderConfigs was not tested because the field itself was marked as `nil` considering the `mfa` values weren't set.

src/auth/project-config.ts
pragatimodi: `link to the docs in the comment` Not sure which doc you are talking about here

prameshj: i meant https://cloud.google.com/identity-platform/docs/reference/rest/v2/projects.tenants#resource:-tenant and https://cloud.google.com/identity-platform/docs/reference/rest/v2/Config

prameshj (approved):

src/auth/project-config.ts
Xiaoshouzi-gh: Adding to this thread - the reason why SDK is using `multiFactorConfig ` is because this is the approved name by API council.

Xiaoshouzi-gh (commented):
Will LGTM after the  documentation fix.

lahirumaramba:
Thank you, @pragatimodi!
Added `release:stage` to trigger integration tests CI. Please fix the CI issues. Thanks!

src/auth/auth-config.ts
```diff
@@ -1503,12 +1632,12 @@ export interface AllowByDefault {
...
-   * https://cldr.unicode.org/
-   * The full list of these region codes is here:
-   * https://github.com/unicode-cldr/cldr-localenames-full/blob/master/main/en/territories.json
-   */
+  /**
+ * Two letter unicode region codes to allow as defined by
+ * https://cldr.unicode.org/
+ * The full list of these region codes is here:
+ * https://github.com/unicode-cldr/cldr-localenames-full/blob/master/main/en/territories.json
+ */
```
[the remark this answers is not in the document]
pragatimodi: Will address these issues in a different PR. Want to keep this one related to MFA changes.

src/auth/auth-config.ts
```diff
@@ -599,6 +648,70 @@ export class MultiFactorAuthConfig implements MultiFactorConfig {
...
+        };
+        for (const key in multiFactorProviderConfig.totpProviderConfig) {
+          if (!(key in validTotpProviderConfigKeys)) {
+            throw new FirebaseAuthError(
+              AuthClientErrorCode.INVALID_CONFIG,
+              `"${key}" is not a valid TotpProviderConfig parameter.`,
+            );
+          }
+        }
+        var adjIntervals = multiFactorProviderConfig.totpProviderConfig.adjacentIntervals
```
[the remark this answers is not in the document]
pragatimodi: It would not be null. I think adjIntervals will not be defined in that case which is being checked in the next line. Thanks for double checking!

etc/firebase-admin.auth.api.md
```diff
@@ -264,9 +264,19 @@ export interface ListUsersResult {
     users: UserRecord[];
 }
 
+// @public
+export class MultiFactorAuthConfig implements MultiFactorConfig {
```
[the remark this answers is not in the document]
pragatimodi: Good point. Yes these types don't exist in the API proposal. Should we update it? @prameshj @Xiaoshouzi-gh @renkelvin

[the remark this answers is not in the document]
renkelvin: Yes, we should stick to the API proposals.

Maybe just make MultiFactorAuthConfig internal?

src/auth/project-config.ts
```diff
@@ -122,8 +122,10 @@ export class ProjectConfig {
     if (configOptions.multiFactorConfig !== undefined) {
       request.mfa = MultiFactorAuthConfig.buildServerRequest(configOptions.multiFactorConfig);
     }
-    /**Backend API returns "mfa" in case of project config and "mfaConfig" in case of tenant config.
-    The SDK exposes it as multiFactorConfig always.*/
+    //Backend API returns "mfa" in case of project config and "mfaConfig" in case of tenant config.
+    //The SDK exposes it as multiFactorConfig always. 
+    //See https://cloud.google.com/identity-platform/docs/reference/rest/v2/projects.tenants#resource:-tenant 
+    //and https://cloud.google.com/identity-platform/docs/reference/rest/v2/Config
```
Xiaoshouzi-gh, line 128: nit: space after //

etc/firebase-admin.auth.api.md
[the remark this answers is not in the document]
Xiaoshouzi-gh: to make it internal, please check `SmsRegionsAuthConfig` for reference.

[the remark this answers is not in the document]
pragatimodi: This class isn't user facing. It just implements the MultiFactorConfig interface which is public. @lahirumaramba should i remove /** */ documentation to undo this?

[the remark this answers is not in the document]
lahirumaramba: Since we export it, it will still show up in the public API. I think we can mark it as `@internal` similar to here: https://github.com/firebase/firebase-admin-node/blob/81672d6e67bf36008fd9ecd4921e7ed4269b0ff6/src/auth/auth-config.ts#L1520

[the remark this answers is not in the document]
pragatimodi: Updated.

Xiaoshouzi-gh:
LGTM, waiting on the test cases explanation docs (internal).

src/auth/auth-config.ts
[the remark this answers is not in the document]
prameshj: Looks like this has been addressed.. Can this be resolved, @pragatimodi ?

src/auth/project-config.ts
```diff
@@ -121,10 +162,14 @@ export class ProjectConfig {
     // JSON serialization
     const json = {
       smsRegionConfig: deepCopy(this.smsRegionConfig),
+      multiFactorConfig: deepCopy(this.multiFactorConfig),
```
prameshj, line 165: "deepCopy(this.multiFactorConfig_)" ?

pragatimodi: Since `multiFactorConfig` is the getter function, shouldn't we use it that way to fetch `multiFactorConfig_` values in that?

Xiaoshouzi-gh (approved):

lahirumaramba (approved):
LGTM! Thank you!

**Please get the docs reviewed before you merge this.** @kevinthecheung might be able to help. Thanks!

src/auth/auth-config.ts
```diff
@@ -506,16 +507,63 @@ export interface MultiFactorConfig {
...
+ * This config is used to set second factor auth except for SMS. 
+ * Currently, only TOTP is supported.
+ */
+export interface MultiFactorProviderConfig {
+  /**
+   * Indicates whether this multi-factor provider is enabled/disabled. 
+   */
+  state: MultiFactorConfigState;
+  /**
+   * TOTP MultiFactor provider config.
```
kevinthecheung, line 529:
```suggestion
   * TOTP multi-factor provider config.
```

src/auth/auth-config.ts
```diff
@@ -506,16 +507,63 @@ export interface MultiFactorConfig {
...
+  totpProviderConfig?: TotpMultiFactorProviderConfig;
+}
+
+/**
+ * Interface representing configuration settings for TOTP second factor auth. 
+ */
+export interface TotpMultiFactorProviderConfig {
+  /**
+    *  The allowed number of adjacent intervals that will be used for verification
+    *  to avoid clock skew.
```
kevinthecheung, line 540:
```suggestion
    *  to compensate for clock skew.
```

src/auth/project-config.ts
```diff
@@ -57,13 +66,25 @@ export class ProjectConfig {
    * This is based on the calling code of the destination phone number.
    */
   public readonly smsRegionConfig?: SmsRegionConfig;
+  /**
+   * The Multi Factor Config for the project.
+   * Configures the multi factor settings for the project.
+   * Supports only phone and TOTP.
+   */
```
kevinthecheung, line 73:
```suggestion
  /**
   * The project's multi-factor auth configuration.
   * Supports only phone and TOTP.
   */
```

kevinthecheung (changes requested):

src/auth/auth-config.ts
```diff
@@ -506,16 +507,63 @@ export interface MultiFactorConfig {
    * Currently only ‘phone’ is supported.
```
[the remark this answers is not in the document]
pragatimodi: This seems to be correct. `factorIds` only supports phone right now.

kevinthecheung (approved):

renkelvin (approved):

PR #1999, opened 2022-12-02
fix(fcm): Increase batch send timeout to 15 seconds

lahirumaramba:
- Increase batch send timeout to 15 seconds
- Extension to: #1947 
Resolves: https://github.com/firebase/firebase-admin-node/issues/1900


chong-shao (approved):

superwoou:
I think you should change the name of constant :(

PR #2085, opened 2023-02-23
Fixing links to externally defined RTDB APIs.

egilmorez:
Issues here discovered via doc feedback: https://listnr.corp.google.com/report/90770486778

I *think* these may have been broken for a while, since we updated the JS v9 docgen process.


hsubox76 (approved):

lahirumaramba (approved):
Thank you, Eric!

PR #2090, opened 2023-02-28
Deprecate sendToDevice and sendToDeviceGroup and their response classes

Doris-Ge:
* Deprecate `sendToDevice` and its response class, `MessagingDeviceResponse`
* Deprecate `sendToDeviceGroup` and its response class, `MessagingDeviceGroupResponse`


lahirumaramba (approved):
Thank you @Doris-Ge ! LGTM!
Let's get a TW review as well.

src/messaging/messaging-api.ts
```diff
@@ -940,7 +940,11 @@ export interface MessagingOptions {
   [key: string]: any | undefined;
 }
 
-/* Individual status response payload from single devices */
+/**
+ * Individual status response payload from single devices
+ *
+ * @deprecated Returned by {@link Messaging#sendToDevice} which is also deprecated.
```
egilmorez, line 946: Nit, but "which" is generally preceded by a comma.  Suggest one here and for similar/identical instances below.

egilmorez (approved):
LG with a nit, thanks!

src/messaging/messaging-api.ts
Doris-Ge: Good catch! Thanks! Done.

PR #2094, opened 2023-03-03
Deprecate sendAll and sendMulticast

Doris-Ge:
1. Deprecate sendAll and sendMulticast
2. Add dummy implementation for sendEach and sendEachForMulticast to avoid errors reported by api-extractor

### DIFF
--- a/etc/firebase-admin.messaging.api.md
+++ b/etc/firebase-admin.messaging.api.md
@@ -184,7 +184,11 @@ export type Message = TokenMessage | TopicMessage | ConditionMessage;
 export class Messaging {
     get app(): App;
     send(message: Message, dryRun?: boolean): Promise<string>;
+    // @deprecated
     sendAll(messages: Message[], dryRun?: boolean): Promise<BatchResponse>;
+    sendEach(messages: Message[], dryRun?: boolean): Promise<BatchResponse>;
+    sendEachForMulticast(message: MulticastMessage, dryRun?: boolean): Promise<BatchResponse>;
+    // @deprecated
     sendMulticast(message: MulticastMessage, dryRun?: boolean): Promise<BatchResponse>;
     sendToCondition(condition: string, payload: MessagingPayload, options?: MessagingOptions): Promise<MessagingConditionResponse>;
     sendToDevice(registrationTokenOrTokens: string | string[], payload: MessagingPayload, options?: MessagingOptions): Promise<MessagingDevicesResponse>;
--- a/src/messaging/messaging.ts
+++ b/src/messaging/messaging.ts
@@ -250,6 +250,56 @@ export class Messaging {
       });
   }
 
+  // TODO: Update the comment based on the implementation
+  /**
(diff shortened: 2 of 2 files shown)


lahirumaramba:
Thanks, @Doris-Ge !
nit: It's better to avoid placeholder methods whenever we can and only add new methods when they are implemented. Since these changes are going to a feature branch, we should be okay. Let's try to avoid empty methods in future PRs if we can.

lahirumaramba (approved):

Doris-Ge:
Got it! I was trying to keep a PR smaller. Sorry about that!

patdx:
Why were sendAll and sendMulticast deprecated (#2094) and sendEach and sendEachForMulticast introduced (#2097)?

Is there a deprecation schedule for sendAll and sendMulticast? Will they be removed from a future version of the SDK or disabled on the server side? 

I tried searching online, including the Firebase docs and SDKs for other languages but having trouble to find any more information about this change. Would appreciate any guidance about this change. Thank you.

brettwillis:
Same here. From #2097:

> Therefore, unlike sendAll, sendEach does not always throw an error for a total failure. It can also return a BatchResponse with only errors in it.

Is this the only reason for the deprecation? As I catch and handle any errors anyway, regardless of using `sendAll()` or `sendEach()`.

Especially as per the `sendAll()` documentation: 

> Employs batching to send the entire list as a single RPC call. Compared to the `send()` method, this method is a significantly more efficient way to send multiple messages.

Is there some other context for this decision? Why should we use `sendEach()` instead of `sendAll()` if it is by implication significantly less efficient? It seems like a step backwards.

brettwillis:
Ah I see there's just been an announcement that the `https://fcm.googleapis.com/batch` batch send REST API is deprecated and will be removed on this day next year (21-June-2024).

See [here](https://firebase.google.com/docs/cloud-messaging/send-message#send-a-batch-of-messages):

> The batch send methods described in this section were deprecated on June 21, 2023, and will be removed in June 2024. For protocol, instead use the [standard HTTP v1 API send method](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages/send), implementing your own batch send by iterating through the list of recipients and sending to each recipient's token. For Admin SDK methods, make sure to update to the next major version.

And technical info [here](https://firebase.google.com/support/faq#fcm-depr-multiple-tokens):

> For use cases where end-to-end latency is critical, or where total fanout size is small (fewer than 1 million), Google recommends the HTTP V1 API. The HTTP V1 API over HTTP/2 can achieve up to ~1/3 of total message throughput rate compared to legacy multicast when sending on a single connection.
> For use cases where throughput and egress bandwidth are prioritized or where total fanout size is large (greater than 1 million), Google recommends topic messaging. While topic messaging requires a one-time action to subscribe recipients to a topic, it offers up to a [10,000 QPS per project fanout rate](https://firebase.google.com/docs/cloud-messaging/concept-options#fanout_throttling) without a maximum limit on topic size.

PR #2097, opened 2023-03-03
Implement `sendEach` and `sendEachForMulticast`

Doris-Ge:
`sendEach` vs `sendAll`
1. `sendEach` sends one HTTP request to V1 Send endpoint for each message in the array. `sendAll` sends only one HTTP request to V1 Batch Send endpoint to send all messages in the array.
2. `sendEach` calls `Promise.allSettled` to wait for all `httpClient.send` calls to complete and construct a `BatchResponse`. An `httpClient.send` call to V1 Send endpoint either completes with a success or throws an error. So if an error is thrown out, the error will be caught in `sendEach` and turned into a `SendResponse` with an error. Therefore, unlike `sendAll`, `sendEach` does not always throw an error for a total failure. It can also return a `BatchResponse` with only errors in it.

`sendEachForMulticast` calls `sendEach` under the hood.

### DIFF
--- a/src/messaging/messaging.ts
+++ b/src/messaging/messaging.ts
@@ -39,6 +39,7 @@ import {
   MessagingConditionResponse,
   DataMessagePayload,
   NotificationMessagePayload,
+  SendResponse,
 } from './messaging-api';
 
 // FCM endpoints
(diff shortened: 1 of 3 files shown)


src/messaging/messaging.ts
```diff
@@ -250,17 +251,16 @@ export class Messaging {
       });
   }
 
-  // TODO: Update the comment based on the implementation
   /**
   * Sends each message in the given array via Firebase Cloud Messaging.
   *
-  * Unlike {@link Messaging.sendAll}, this method makes a single RPC call for each message in the given array.
+  * Unlike {@link Messaging.sendAll}, this method makes a single RPC call for each message
```
Doris-Ge, line 257: Should this be a RPC call or an HTTP call?

lahirumaramba (approved):
Thank you @Doris-Ge ! Looks great!

Let's get the docs reviewed. Feel free to do it now or later when you make the merge to the main branch. Thanks!

src/messaging/messaging.ts
lahirumaramba: I think we use RPC in other places in the SDK. Let's get the docs reviewed so we can address all these naming issues. Thanks!

Doris-Ge: Sounds good! Based on the implementation, I think we actually send an HTTP call instead of an internal RPC call. I can leave this as it is now.

src/messaging/messaging.ts
```diff
@@ -270,22 +270,67 @@ export class Messaging {
...
+      // a regular array here before calling deepCopy(). See issue #566 for details.
+      messages = Array.from(messages);
+    }
+
+    const copy: Message[] = deepCopy(messages);
+    if (!validator.isNonEmptyArray(copy)) {
+      throw new FirebaseMessagingError(
+        MessagingClientErrorCode.INVALID_ARGUMENT, 'messages must be a non-empty array');
+    }
+    if (copy.length > FCM_MAX_BATCH_SIZE) {
```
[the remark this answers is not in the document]
Doris-Ge: I think we're okay to send 500 requests using `v1` API. Let me double check this with Xin in our group chat.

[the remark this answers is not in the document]
Doris-Ge: Discussed with Xin and Chong offline. The consensus is to keep this number to be 500. I'll mark this as resolved.

PR #2104, opened 2023-03-08
Allow bypassing getUser request when creating or updating a user

### LINKED ISSUE #1885: [FR] Auth: Provide an API to create/update users without unneeded API requests
The [`updateUser()`](https://firebase.google.com/docs/reference/admin/node/firebase-admin.auth.baseauth.md#baseauthupdateuser) and [`createUser()`](https://firebase.google.com/docs/reference/admin/node/firebase-admin.auth.baseauth.md#baseauthcreateuser) methods always perform two API requests under the hood: (1) the create/update request, and (2) a subsequent [`getUser()`](https://firebase.google.com/docs/reference/admin/node/firebase-admin.auth.baseauth.md#baseauthgetuser) request. 

In latency-sensitive applications where the result of `getUser()` is not required, this means an extra un-needed sequential round-trip network request. This fact is not apparent from the API or documentation.

### Proposal

Add a second, optional parameter to `createUser()` and `updateUser()` called something like `noFetchUserRecord` which defaults to `false` (existing behaviour). If `true`, the `getUser()` request will be skipped, and the function will return `Promise<void>`, avoiding the un-needed round-trip request.

brettwillis:
Closes #1885 

### Summary

The [`updateUser()`](https://firebase.google.com/docs/reference/admin/node/firebase-admin.auth.baseauth.md#baseauthupdateuser) and [`createUser()`](https://firebase.google.com/docs/reference/admin/node/firebase-admin.auth.baseauth.md#baseauthcreateuser) methods always perform two API requests under the hood: (1) the create/update request, and (2) a subsequent [`getUser()`](https://firebase.google.com/docs/reference/admin/node/firebase-admin.auth.baseauth.md#baseauthgetuser) request. 

This PR adds an optional parameter `noFetchUserRecord ` that allows bypassing the `getUser()` request when not needed, so that it returns void.

### DIFF
--- a/etc/firebase-admin.auth.api.md
+++ b/etc/firebase-admin.auth.api.md
@@ -73,7 +73,9 @@ export abstract class BaseAuth {
     createCustomToken(uid: string, developerClaims?: object): Promise<string>;
     createProviderConfig(config: AuthProviderConfig): Promise<AuthProviderConfig>;
     createSessionCookie(idToken: string, sessionCookieOptions: SessionCookieOptions): Promise<string>;
-    createUser(properties: CreateRequest): Promise<UserRecord>;
+    createUser(properties: CreateRequest, noFetchUserRecord?: boolean): Promise<UserRecord>;
+    // (undocumented)
+    createUser(properties: CreateRequest, noFetchUserRecord: true): Promise<void>;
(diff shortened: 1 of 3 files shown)


brettwillis:
@lahirumaramba  @prameshj please check thanks 🙏🏼

PR #2138, opened 2023-04-05
feat(fcm): Add `sendEach` and `sendEachForMulticast` for FCM batch send

Doris-Ge:
- Deprecate `sendAll()` and `sendMulticast()` APIs.
- Add new `sendEach()` and `sendEachForMulticast()` APIs.

### DIFF
--- a/src/messaging/messaging.ts
+++ b/src/messaging/messaging.ts
@@ -258,9 +258,9 @@ export class Messaging {
   * in the given array.
   *
   * The responses list obtained from the return value corresponds to the order of `messages`.
-  * An error from this method or a `BatchResponse` with all failures indicates a total failure
-  * -- i.e. none of the messages in the list could be sent. Partial failures or no failures
-  * are only indicated by a `BatchResponse` return value.
+  * An error from this method or a `BatchResponse` with all failures indicates a total failure,
+  * meaning that none of the messages in the list could be sent. Partial failures or no
+  * failures are only indicated by a `BatchResponse` return value.
   *
   * @param messages - A non-empty array
   *   containing up to 500 messages.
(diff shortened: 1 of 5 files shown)


src/messaging/messaging.ts
```diff
@@ -250,6 +251,124 @@ export class Messaging {
...
 
+  /**
+  * Sends each message in the given array via Firebase Cloud Messaging.
+  *
+  * Unlike {@link Messaging.sendAll}, this method makes a single RPC call for each message
+  * in the given array.
+  *
+  * The responses list obtained from the return value corresponds to the order of `messages`.
+  * An error from this method or a `BatchResponse` with all failures indicates a total failure
+  * -- i.e. none of the messages in the list could be sent. Partial failures or no failures
```
egilmorez, line 262: We avoid i. e., so for this I'd suggest just "...indicates a total failure, mean that none of the messages in the list could be sent.

src/messaging/messaging.ts
```diff
@@ -250,6 +251,124 @@ export class Messaging {
...
+
+  /**
+   * Sends the given multicast message to all the FCM registration tokens
+   * specified in it.
+   *
+   * This method uses the {@link Messaging.sendEach} API under the hood to send the given
+   * message to all the target recipients. The responses list obtained from the
+   * return value corresponds to the order of tokens in the `MulticastMessage`.
+   * An error from this method or a `BatchResponse` with all failures indicates a total failure
+   * -- i.e. none of the messages in the list could be sent. Partial failures or no failures
```
egilmorez, line 332: Same suggestion as above RE "-- i.e."

egilmorez (approved):
LG with one optional nit, thanks!

Doris-Ge:
Thanks Eric for reviewing the PR! @lahirumaramba I tried to push the fix to the fcm-batch-send branch but it got rejected because "Protected branch update failed for refs/heads/fcm-batch-send". Should I create another PR from dorisge/fcm-batch-send for the fix and merge it to fcm-batch-send?

src/messaging/messaging.ts
egilmorez: Oh I'm kind of glad you got that failure, as I see I have a typo in my comment!!

It should read:

"We avoid i. e., so for this I'd suggest just "...indicates a total failure, meaning that none of the messages in the list could be sent."

Note "meaning" with "ing," thanks!

Doris-Ge: Thank you for noticing that! I also had the question and should've asked.  I'll fix that.

lahirumaramba (approved):
Thank you! LGTM

PR #2148, opened 2023-04-12
feat(appcheck): Added replay protection feature to App Check `verifyToken()` API

lahirumaramba:
RELEASE NOTE: Added replay protection feature to App Check `verifyToken()` API.

### DIFF
--- a/src/app-check/app-check-api-client-internal.ts
+++ b/src/app-check/app-check-api-client-internal.ts
@@ -27,6 +27,7 @@ import { AppCheckToken } from './app-check-api'
 
 // App Check backend constants
 const FIREBASE_APP_CHECK_V1_API_URL_FORMAT = 'https://firebaseappcheck.googleapis.com/v1/projects/{projectId}/apps/{appId}:exchangeCustomToken';
+const ONE_TIME_USE_TOKEN_VERIFICATION_URL_FORMAT = 'https://firebaseappcheck.googleapis.com/v1beta/projects/{projectId}:verifyAppCheckToken';
 
 const FIREBASE_APP_CHECK_CONFIG_HEADERS = {
   'X-Firebase-Client': `fire-admin-node/${utils.getSdkVersion()}`
@@ -86,6 +87,31 @@ export class AppCheckApiClient {
       });
   }
 
+  public verifyOneTimeProtection(token: string): Promise<boolean> {
+    if (!validator.isNonEmptyString(token)) {
+      throw new FirebaseAppCheckError(
+        'invalid-argument',
+        '`token` must be a non-empty string.');
+    }
+    return this.getVerifyTokenUrl()
+      .then((url) => {
+        const request: HttpRequestConfig = {
+          method: 'POST',
+          url,
+          headers: FIREBASE_APP_CHECK_CONFIG_HEADERS,
+          data: { token }
+        };
+        return this.httpClient.send(request);
+      })
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      .then((resp) => {
+        return true;
+      })
+      .catch((err) => {
+        throw this.toFirebaseError(err);
+      });
+  }
+
   private getUrl(appId: string): Promise<string> {
     return this.getProjectId()
       .then((projectId) => {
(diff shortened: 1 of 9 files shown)


src/app-check/app-check.ts
```diff
@@ -75,17 +78,41 @@ export class AppCheck {
...
    * @returns A promise fulfilled with the token's decoded claims
    *   if the App Check token is valid; otherwise, a rejected promise.
    */
-  public verifyToken(appCheckToken: string): Promise<VerifyAppCheckTokenResponse> {
+  public verifyToken(appCheckToken: string, options?: VerifyAppCheckTokenOptions)
+    : Promise<VerifyAppCheckTokenResponse> {
+    this.validateVerifyAppCheckTokenOptions(options);
     return this.appCheckTokenVerifier.verifyToken(appCheckToken)
       .then((decodedToken) => {
+        if (options?.consume) {
```
weixifan, line 91: If the initial verifyToken(...) fails due to an invalid token, do we skip this step? I don't know TS that well -- if we can save the network call and throw right away that would be ideal.

src/app-check/app-check-api-client-internal.ts
```diff
@@ -86,6 +87,31 @@ export class AppCheckApiClient {
...
+          data: { token }
+        };
+        return this.httpClient.send(request);
+      })
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      .then((resp) => {
+        return true;
+      })
+      .catch((err) => {
+        throw this.toFirebaseError(err);
```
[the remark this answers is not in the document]
lahirumaramba: Ah good point! Since we check the validity of the token prior to making the call to BE, #1 should be addressed already. How do we check #2, is it a custom claim encoded in the JWT?

src/app-check/app-check.ts
lahirumaramba: Yep! We fail fast here. If `verifyToken` throws then the code won't continue an we will not make a network call even if `consume` is set to true. I added unit tests to cover this usecase.

test/unit/app-check/app-check-api-client-internal.spec.ts
```diff
@@ -235,4 +235,7 @@ describe('AppCheckApiClient', () => {
         });
       });
   });
+
+  //describe verifyReplayProtection
+
```
lahirumaramba, line 240: TODO: Need to add unit tests for `verifyReplayProtection`

src/app-check/app-check-api-client-internal.ts
```diff
@@ -86,6 +87,30 @@ export class AppCheckApiClient {
       });
   }
 
+  public verifyReplayProtection(token: string): Promise<boolean> {
```
lahirumaramba, line 90: @weixifan this is an internal method, but I renamed it to `verifyReplayProtection` from `verifyOneTimeUseProtection`. WDYT?

taeold (approved):

src/app-check/app-check-api.ts
```diff
@@ -41,6 +41,27 @@ export interface AppCheckTokenOptions {
   ttlMillis?: number;
 }
 
+/**
+ * Interface representing options for {@link AppCheck.verifyToken} method.
```
kevinthecheung, line 45:
```suggestion
 * Interface representing options for the {@link AppCheck.verifyToken} method.
```

src/app-check/app-check-api.ts
```diff
@@ -41,6 +41,27 @@ export interface AppCheckTokenOptions {
...
+ */
+export interface VerifyAppCheckTokenOptions {
+  /**
+   * To use the replay protection feature, set this to true to mark the token as consumed.
+   * Tokens that are found to be already consumed will be marked as such in the response.
```
kevinthecheung, line 50:
```suggestion
   * To use the replay protection feature, set this to true. The {@link AppCheck.verifyToken}
   * method will mark the token as consumed after verifying it.
   *
   * Tokens that are found to be already consumed will be marked as such in the response.
```

kevinthecheung (approved):

src/app-check/app-check-api-client-internal.ts
weixifan: Yeah that sounds fine to me -- in any case since this is internal it can be changed in the future too.

weixifan (approved):

src/app-check/app-check-api-client-internal.ts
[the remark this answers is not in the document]
weixifan: For (2), yes, it is actually a private claim named `provider` in the JWT, but now that I think about this more, I think a more robust method for detecting whether the token supports Replay Protection is to see whether its `jti` claim is populated with a nonempty string rather than checking the `provider` claim against a hardcoded list of allowed providers.

PR #2151, opened 2023-04-17
fix: Export error classes

IchordeDionysos:
### Discussion

Expose the `FirebaseError` classes to enable code like this:

```ts
catch (e) {
  if (e instanceof FirebaseError) {
    if (e.code === 'auth/' + AuthClientErrorCode.USER_NOT_FOUND.code) {
      throw new UserNotFoundException({email});
    }
  }
  throw e;
}
```

### Testing

### API Changes

    can discuss it together.


jketcham:
Would love to see this added, it doesn't appear there's another way to easily access the error types from this package?

Bullfrog1234:
I agree this is needed and follows the standard JS pattern for checking for a particular error. Error handling is important to good quality code and this library needs to support that also.

@IchordeDionysos this has got out of data with the base branch may need an update

lahirumaramba (approved):
Thank you @IchordeDionysos ! LGTM!

PR #2164, opened 2023-04-25
AppPlatform exported enum string checks

DellaBitta:
### Discussion

Add tests to ensure that the AppPlatforms enumerated type is being exported properly on builds.

### Testing
Updated the unit test to test for the validity of the enum strings.

### API Changes
None.


DellaBitta:
PR isn't required. The original issue had already been resolved, and these tests are no longer needed.

PR #2167, opened 2023-04-27
Add Firestore Filter export

danReynolds:
Filter from Firestore isn't exposed on the exports. It'd be helpful to add I think so that folks can use it directly from this package.


lahirumaramba:
Thank you @danReynolds for the contribution!
We need to update the api docs. Please do `npm run api-extractor:local` and push the new changes.

Please also update the unit and integration tests. For example:
https://github.com/firebase/firebase-admin-node/blob/1342b458667866f7ee6399713ddf1f306519a54a/test/integration/firestore.spec.ts#L114

jhuleatt:
#2192 has been merged! Thanks for starting this work, @danReynolds!

PR #2197, opened 2023-05-26
feat(auth): Add `TotpInfo` field to `UserRecord`

pragatimodi:
Adding a `TotpInfo` field to `UserRecord` for Admin users to view a user's MFA configuration. TOTP enrollment for users is enabled via client SDKs only. 
Testing was done by manually enrolling a user on TOTP using curl and creating a test app to get the corresponding user's `UserRecord` configuration
Manual testing output:
```
MultiFactorSettings {
  enrolledFactors: [
    TotpMultiFactorInfo {
      uid: 'bd8fa8e0-194c-443e-82e9-77b39868aed7',
      factorId: 'totp',
      displayName: 'enroll-totp-test',
      enrollmentTime: 'Wed, 31 May 2023 20:39:18 GMT',
      totpInfo: {}
    }
  ]
}
```

### DIFF
--- a/src/auth/user-record.ts
+++ b/src/auth/user-record.ts
@@ -134,6 +134,8 @@ export abstract class MultiFactorInfo {
         multiFactorInfo = new PhoneMultiFactorInfo(response);
       } else if (response.totpInfo !== undefined) {
         multiFactorInfo = new TotpMultiFactorInfo(response);
+      } else {
+        // Ignore the other SDK unsupported MFA factors to prevent blocking developers using the current SDK.
       }
     } catch (e) {
       // Ignore error.
@@ -237,6 +239,7 @@ export class PhoneMultiFactorInfo extends MultiFactorInfo {
         phoneNumber: this.phoneNumber,
       });
   }
+
   /**
    * Returns the factor ID based on the response provided.
    *
--- a/src/auth/user-record.ts
+++ b/src/auth/user-record.ts
@@ -49,11 +49,11 @@ export interface MultiFactorInfoResponse {
   phoneInfo?: string;
   totpInfo?: TotpInfoResponse;
   enrolledAt?: string;
-  [key: string]: any;
+  [key: string]: unknown;
 }
 
 export interface TotpInfoResponse {
-  [key: string]: any;
+  [key: string]: unknown;
 }
 
 export interface ProviderUserInfoResponse {
--- a/src/auth/user-record.ts
+++ b/src/auth/user-record.ts
@@ -47,7 +47,12 @@ export interface MultiFactorInfoResponse {
   mfaEnrollmentId: string;
   displayName?: string;
   phoneInfo?: string;
+  totpInfo?: TotpInfoResponse;
   enrolledAt?: string;
+  [key: string]: any;
+}
+
+export interface TotpInfoResponse {
   [key: string]: any;
 }
 
@@ -84,6 +89,7 @@ export interface GetAccountInfoUserResponse {
 
 enum MultiFactorId {
   Phone = 'phone',
+  Totp = 'totp',
 }
 
 /**
@@ -102,7 +108,9 @@ export abstract class MultiFactorInfo {
   public readonly displayName?: string;
 
   /**
-   * The type identifier of the second factor. For SMS second factors, this is `phone`.
+   * The type identifier of the second factor.
+   * For SMS second factors, this is `phone`.
+   * For TOTP second factors, this is `totp`.
    */
   public readonly factorId: string;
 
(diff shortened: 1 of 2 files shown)


test/unit/auth/user-record.spec.ts
```diff
@@ -439,6 +584,12 @@ describe('MultiFactorSettings', () => {
       enrolledAt: now.toISOString(),
       phoneInfo: '+16505556789',
     }),
+    new TotpMultiFactorInfo({
+      mfaEnrollmentId: 'enrollmentId5',
```
renkelvin, line 588: why 'enrollmentId5' not 'enrollmentId3'?

src/auth/user-record.ts
```diff
@@ -47,10 +47,15 @@ export interface MultiFactorInfoResponse {
   mfaEnrollmentId: string;
   displayName?: string;
   phoneInfo?: string;
+  totpInfo?: TotpInfoResponse;
   enrolledAt?: string;
   [key: string]: any;
 }
 
+export interface TotpInfoResponse {
+  [key: string]: any;
```
renkelvin, line 56: Try use `unknown` instead of `any` for better type safety.

src/auth/user-record.ts
```diff
@@ -240,14 +251,68 @@ export class PhoneMultiFactorInfo extends MultiFactorInfo {
   }
 }
 
+/**
+ * TotpInfo struct associated with a second factor
+ */
+export class TotpInfo {
```
renkelvin, line 257: Is this neccessary?

test/unit/auth/user-record.spec.ts
pragatimodi: Because there are 2 invalid configs in the `mfaInfo` list called `enrollmentId3` and `enrollmentId4`.
But seems like we can regroup or write better variable names for those.

src/auth/user-record.ts
pragatimodi: The backend proto is an empty struct so in case we need to add more information about this later, it might be easier to have a strong type struct than `any` defined in the `MultiFactorInfo` type in case something changes in the future. Happy to reconsider in case there's a suggestion.

src/auth/user-record.ts
```diff
@@ -47,8 +47,13 @@ export interface MultiFactorInfoResponse {
...
   displayName?: string;
   phoneInfo?: string;
+  totpInfo?: TotpInfoResponse;
   enrolledAt?: string;
-  [key: string]: any;
+  [key: string]: unknown;
+}
+
+export interface TotpInfoResponse {
+  [key: string]: unknown;
```
prameshj, line 56: Can we leave this empty for now? We can always add fields as they come up.

src/auth/user-record.ts
```diff
@@ -47,8 +47,13 @@ export interface MultiFactorInfoResponse {
   mfaEnrollmentId: string;
   displayName?: string;
   phoneInfo?: string;
+  totpInfo?: TotpInfoResponse;
   enrolledAt?: string;
-  [key: string]: any;
+  [key: string]: unknown;
```
prameshj, line 52: Let's leave this existing field type as-is

prameshj (commented):
Can you update the PR description to describe the change and also include the manual testing done? Thanks!

src/auth/user-record.ts
pragatimodi: We cannot declare an empty interface as per lint rules.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "firebase-admin",
-  "version": "11.8.0",
+  "version": "11.9.0",
```
prameshj, line 3: 11.8.1 is sufficient since this is a bug fix/patch? @lahirumaramba

prameshj (approved):
Can you leave a snippet of the user record output in the PR description? Thanks!

renkelvin (approved):

src/auth/user-record.ts
```diff
@@ -120,9 +128,13 @@ export abstract class MultiFactorInfo {
...
+    // PhoneMultiFactorInfo, TotpMultiFactorInfo currently available.
     try {
-      multiFactorInfo = new PhoneMultiFactorInfo(response);
+      if (response.phoneInfo !== undefined) {
+        multiFactorInfo = new PhoneMultiFactorInfo(response);
+      } else if (response.totpInfo !== undefined) {
+        multiFactorInfo = new TotpMultiFactorInfo(response);
+      }
     } catch (e) {
       // Ignore error.
```
Xiaoshouzi-gh, line 139: nit: can we add an empty else block with comment like - 
Ignore the other SDK unsupported MFA factors to prevent blocking developers using the current SDK.

Xiaoshouzi-gh (approved):
LGTM

src/auth/user-record.ts
lahirumaramba, line 56: Why is this currently empty?
Could you also share an example of how a developer calls the Admin SDK API to obtain this response, please?

pragatimodi:
```
getAuth().getUserByEmail("abc@example.com").then(user => {
    console.log(user.multiFactor);
})
```
Here's an example of how this response is obtained. The user record will display `TotpInfo` in case TOTP has been enrolled by the users. It doesn't cause any changes in creating/updating a user.

lahirumaramba: Thank you!

lahirumaramba (approved):
Thank you! LGTM!

PR #2209, opened 2023-06-09
Expose MultiDB within Firestore

tom-andersen:


etc/firebase-admin.firestore.api.md
```diff
@@ -100,11 +100,20 @@ export function getFirestore(): Firestore;
 // @public
 export function getFirestore(app: App): Firestore;
 
+// @beta
+export function getFirestore(databaseId: string): Firestore;
+
+// @beta
+export function getFirestore(app: App, databaseId: string): Firestore;
+
```
lahirumaramba, line 108: I think if we combine the two methods the docs will be more readable.

```
export function getFirestore(app?: App, databaseId: string): Firestore;
```

etc/firebase-admin.firestore.api.md
```diff
@@ -100,11 +100,20 @@ export function getFirestore(): Firestore;
...
+export function getFirestore(app: App, databaseId: string): Firestore;
+
 export { GrpcStatus }
 
 // @public
 export function initializeFirestore(app: App, settings?: FirestoreSettings): Firestore;
 
+// @beta
+export function initializeFirestore(app: App, settings: FirestoreSettings, databaseId: string): Firestore;
+
```
lahirumaramba, line 116: Would it make more sense to include the `databaseId` in `FirestoreSettings`?

src/firestore/index.ts
```diff
@@ -75,58 +75,72 @@ export {
 export { FirestoreSettings };
 
 /**
- * Gets the {@link https://googleapis.dev/nodejs/firestore/latest/Firestore.html | Firestore}
+ * Gets the default {@link https://googleapis.dev/nodejs/firestore/latest/Firestore.html | Firestore}
  * service for the default app.
  *
- * `getFirestore()` can be called with no arguments to access the default
```
lahirumaramba, line 81: Please get these documentation changes reviewed by a tech writer before merging the PR. Thanks!

lahirumaramba (changes requested):
Thank you! Added some comments

etc/firebase-admin.firestore.api.md
tom-andersen: TLDR; No, I I think this is a bad idea.

The admin SDK manages lifecycle of Firestore instances, such that subsequent requests will return the same instance. The `app` and `databaseId` form the key to this cache, and should really be held separate from settings in my opinion.

Other methods such as `getFirestore`, do not take a settings object. Adding a settings object here would conflict with intent to have `getFirestore` simply be a method to return an existing instance (if one exists).

`initializeFirestore` cannot be called with different settings, since settings cannot be changed after Firestore instance has been started. The semantic difference that "some" settings such as `database id` are allowed to change, will simply add confusion to interface. A simple rule that no settings are allowed to change can be maintained by excluding databaseId from settings.

etc/firebase-admin.firestore.api.md
tom-andersen: I think inviting customers to pass null as a parameter is a bad idea. 

The problem is in doc generation that doesn't include parameters in signature:

image

src/firestore/index.ts
tom-andersen: @markarndt

markarndt (approved):

src/firestore/index.ts
tom-andersen: Thanks Mark!

etc/firebase-admin.firestore.api.md
lahirumaramba: What I meant to say was to keep both as optional parameters so you don't need three method signatures. It also simplifies the docs.

```
export function getFirestore(app?: App, databaseId?: string): Firestore;

export function getFirestore(
  appOrDatabaseId?: App | string,
  optionalDatabaseId?: string
): Firestore {
```

lahirumaramba (approved):

etc/firebase-admin.firestore.api.md
tom-andersen: I still think calling `getFirestore(null, 'myDB')` is worse than having overload `getFirestore('myDB')`.

TypeScript doesn't expose the last signature, nor would I want to. The signature suggests someone could write code `getFirestore('myDB', 'myDB')` which isn't allowed.

```
export function getFirestore(
  appOrDatabaseId?: App | string,
  optionalDatabaseId?: string
): Firestore
```

PR #2216, opened 2023-06-14
feat(functions): Add features to task queue functions

blidd-google:
Since CF3 released support for Task Queue Functions at Google I/O 2022, users have made several feature requests. This PR enhances the Task Queue functions API to enable the following:

* [Headers passed to TaskQueue HTTPS handlers](https://cloud.google.com/tasks/docs/creating-http-target-tasks#handler) are now available to the context/event.
* Users can name their tasks by including an `id` to TaskOptions when enqueueing tasks.
* Users can cancel their tasks by providing the task `id` to the new `delete` API.

See https://github.com/firebase/firebase-functions/pull/1423 for the sister changes in the functions SDK.


test/unit/functions/functions-api-client-internal.spec.ts
```diff
@@ -137,63 +138,72 @@ describe('FunctionsApiClient', () => {
...
+        expect(apiClient.enqueue({}, invalidName as any))
+          .to.eventually.throw('Function name must be a single string or a qualified resource name');
       });
     }
 
     for (const invalidOption of [null, 'abc', '', [], true, 102, 1.2]) {
       it(`should throw if options is ${invalidOption}`, () => {
-        expect(() => apiClient.enqueue({}, FUNCTION_NAME, '', invalidOption as any))
-          .to.throw('TaskOptions must be a non-null object');
+        expect(apiClient.enqueue({}, FUNCTION_NAME, '', invalidOption as any))
```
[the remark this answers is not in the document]
blidd-google: Hm I deliberately chose to use chai's `eventually` to wait on the Promise returned by `enqueue` calls instead of async/await, to align better with the existing implementation which generally avoids async/await. Should I go ahead and refactor the unit tests to use async/await instead?

package-lock.json
lahirumaramba, line 1: Since we don't make any changes to the `package.json` in this PR, let's revert the lock file changes and let dependabot take care of the `package-lock.json` updates.

lahirumaramba (changes requested):

src/utils/validator.ts
```diff
@@ -278,3 +278,19 @@ export function isTopic(topic: any): boolean {
...
+ * @returns Whether the provided task ID is valid. 
+ */
+export function isTaskId(taskId: any): boolean {
+  if (typeof taskId !== 'string') {
+    return false;
+  }
+
+  const VALID_TASK_ID_REGEX = /^[A-Za-z0-9_-]+$/;
+  return VALID_TASK_ID_REGEX.test(taskId);
+}
```
lahirumaramba, line 296: nit: new line

src/functions/functions-api.ts
```diff
@@ -58,6 +60,40 @@ export type TaskOptions = DeliverySchedule & TaskOptionsExperimental & {
...
+   * prefixes, for example using a timestamp, causes an increase in latency and error rates in
+   * all task commands. The infrastructure relies on an approximately uniform distribution of
+   * task ids to store and serve tasks efficiently.
+   *
+   * "Push IDs" from the Firebase Realtime Database make poor IDs because they are based on
+   * timestamps and will cause contention (slow downs) in your task queue. Reversed push IDs
+   * however form a perfect distribution and are an ideal key. To reverse a string in
+   * javascript use `someString.split("").reverse().join("")`
+   */
+  id?: string;
```
lahirumaramba, line 84: Let's get the documentation reviewed by someone from the tech writing team

lahirumaramba (changes requested):
Thank you for the PR!
Please get a TW review on the documentation changes before merging this.

src/functions/functions-api.ts
```diff
@@ -70,3 +106,15 @@ export interface TaskOptionsExperimental {
  */
   uri?: string;
 }
+
+export class TaskAlreadyExists extends FirebaseError {
```
lahirumaramba, line 110: I would suggest we reuse the `FirebaseFunctionsError` type for this and add a new error code for `task-already-exists` error response to keep it consistent with the rest of the services.

src/functions/functions-api-client-internal.ts
```diff
@@ -54,6 +55,61 @@ export class FunctionsApiClient {
     }
     this.httpClient = new AuthorizedHttpClient(app as FirebaseApp);
   }
+  /**
+   * Deletes a task from a queue.
+   *
+   * @param id - The ID of the task to delete.
+   * @param functionName - The functionName of the queue.
```
egilmorez, line 62: Is functionName a literal?

But it probably makes more sense to just say "function name" in the definition/description.

src/functions/functions-api.ts
```diff
@@ -58,6 +60,40 @@ export type TaskOptions = DeliverySchedule & TaskOptionsExperimental & {
    * The default is 10 minutes. The deadline must be in the range of 15 seconds and 30 minutes.
    */
   dispatchDeadlineSeconds?: number;
+
+  /**
+   * The id to use for the enqueued event.
```
egilmorez, line 63: Suggest caps for "ID"

src/functions/functions-api.ts
```diff
@@ -58,6 +60,40 @@ export type TaskOptions = DeliverySchedule & TaskOptionsExperimental & {
...
+   * If not provided, one will be automatically generated.
+   * If provided explicitly specifying a task ID enables task de-duplication. If a task's ID is
```
egilmorez, line 65: iiuc, this reads better as, "If provided, an explicitly specified task ID enables . . ."

src/functions/functions-api.ts
```diff
@@ -58,6 +60,40 @@ export type TaskOptions = DeliverySchedule & TaskOptionsExperimental & {
...
+   * identical to that of an existing task or a task that was deleted or executed recently then
+   * the call will throw a TaskAlreadyExists error. Another task with the same id can't be
```
egilmorez, line 67: Suggest `TaskAlreadyExists` and capped "ID" in the next line.

src/functions/functions-api.ts
```diff
@@ -58,6 +60,40 @@ export type TaskOptions = DeliverySchedule & TaskOptionsExperimental & {
...
+   * created for ~1hour after the original task was deleted or executed.
+   * 
+   * Because there is an extra lookup cost to identify duplicate task ids, setting id
```
egilmorez, line 70: Yup, please do a scrub throughout for "ID," which should be capped -- unless you're referring to a lowercase literal, in which case it should be backticked.  Thanks!

src/functions/functions-api.ts
```diff
@@ -58,6 +60,40 @@ export type TaskOptions = DeliverySchedule & TaskOptionsExperimental & {
...
+   * significantly increases latency. Using hashed strings for the task id or for the prefix of
+   * the task id is recommended. Choosing task ids that are sequential or have sequential
+   * prefixes, for example using a timestamp, causes an increase in latency and error rates in
+   * all task commands. The infrastructure relies on an approximately uniform distribution of
+   * task ids to store and serve tasks efficiently.
+   *
+   * "Push IDs" from the Firebase Realtime Database make poor IDs because they are based on
+   * timestamps and will cause contention (slow downs) in your task queue. Reversed push IDs
```
egilmorez, line 78: Suggest "slowdowns" (otherwise it looks like verb or command).

egilmorez (commented):
I got a certain way through this and GH told me my diff was outdated . . . submitting these comments so far and will look again, thanks!

src/functions/functions-api.ts
blidd-google: Actually wanted to get your thoughts on how to document this — instead of throwing a TaskAlreadyExists error, we changed the behavior to throw a FirebaseFunctionsError with code "functions/task-already-exists". So if a user wanted to check if a thrown error is due to the task already existing, they would check like so:

```
try {
  await queue.enqueue(...);
} catch (err: unknown) {
  if (err.code === "functions/task-already-exists") {
    ...
  }
}
```
Is this comment in the functions API sufficient?

PR #2227, opened 2023-06-27
Fixes to password policy validation

kevinthecheung:
- Reorder checks so that you can pass undefined to use defaults
- Fix some 'undefined' comparisons


pragatimodi (approved):

pragatimodi:
LGTM, thank you for finding these!

PR #2236, opened 2023-07-07
fix(auth): Memory Leak in AsyncHttpCall affecting auth.listUsers

### LINKED ISSUE #2235: Memory Leak in getAuth(...).listUsers(...)
### Environment

  * Operating System version: macOS 13.4.1 (but also in within a Google Cloud Run NodeJS environment)
  * Firebase SDK version:
    * firebase: 9.23.0 (also 9.21.0)
    * firebase-admin: 11.9.0 (also 11.7.0)
    * firebase-functions: 4.4.1 (also 4.3.1)
  * Firebase Product: auth
  * Node.js version: 16, 18, 20
  * NPM version: 9.5.1
 
#### Steps to reproduce:

We discovered a Memory Leak when running the list_all_users example code from https://firebase.google.com/docs/auth/admin/manage-users#list_all_users.
The data loaded from `getAuth(app).listUsers(1000, nextPageToken)` is not garbage collected and causes the memory footprint of our authentication backup to grow and eventually run out of memory.

node20-plain-new-dep

Every blue bar is memory that is still used with is ~1MB/1000 Users making it almost impossible to use this function to backup multiple GB of user data.
The problematic path seems to be within: `node_modules/firebase-admin/lib/utils/api-request.js`.

Side-note: We only started noticing this issue with node v20, with node v18 and below we did not have the issue when using our production bundle that is generated with typescript and webpack. We could go back to previous versions and can see that the memory is correctly cleaned up in older node versions. Unfortunately we couldn't figure out what exactly the difference to the simple issue demo script I've added below is, which has the issue even on older versions. 
It just shows us, that there is now an issue that didn't exist before. 

We could also see the same when not using a loop and just calling `auth.listUsers(1000)` once, so the issue is not in the boilerplate around. Also the issue goes away when we mock this function call.

Here you can see a Memory Profile of the issue with our webpack output. You can clearly see, that in node20 the blue bars that are kept are way larger. In node18 and before the data was immediately cleaned up.

node 20 | node 18 (16 looks similar)
--- | ---
node20-webpack-new-dep | node18-webpack-new-dep

#### Relevant Code:

```js
const { initializeApp } = require("firebase-admin/app");
const { getAuth } = require("firebase-admin/auth");

const app = initializeApp({
  projectId: /* TODO */,
});

const auth = getAuth(app);

const listAllUsers = (nextPageToken) => {
  auth
    .listUsers(1000, nextPageToken)
    .then((listUsersResult) => {
      listUsersResult.users.forEach((userRecord) => {
        console.log("user", userRecord.toJSON());
      });
      if (listUsersResult.pageToken) {
        listAllUsers(listUsersResult.pageToken);
      }
    })
    .catch((error) => {
      console.log("Error listing users:", error);
    });
};

listAllUsers();
```

adrianjost:
fixes #2235

The socket code was only introduced to fix a bug in NodeJS 12 (https://github.com/firebase/firebase-admin-node/commit/db8be2620a5c9cfc5b9824d9d2487ab9815f5baf) and since then has even needed a patch (https://github.com/firebase/firebase-admin-node/commit/1a34bc49373a5a34bb0efc006ea156d66168103c). However the code does not seem necessary at all anymore. Removing this code resolved the described memory leak.

### Discussion

 - [x] Read the contribution guidelines (CONTRIBUTING.md).
 - [x] If this has been discussed in an issue, make sure to link to the issue here. 

### Testing

 - [ ] Make sure all existing tests in the repository pass after your change. - CI not started yet, but we are running it in production without issues 🤷‍♂️
 - [x] ~If you fixed a bug or added a feature, add a new test to cover your code.~ Not sure how this could be tested

### API Changes

 - [x] At this time we cannot accept changes that affect the public API.  If you'd like to help 
    can discuss it together.

### DIFF
--- a/src/utils/api-request.ts
+++ b/src/utils/api-request.ts
@@ -504,14 +504,6 @@ class AsyncHttpCall {
     if (timeout) {
       // Listen to timeouts and throw an error.
       req.setTimeout(timeout, timeoutCallback);
-      req.on('socket', (socket) => {
-        socket.setMaxListeners(socket.getMaxListeners() + 1);
-        socket.setTimeout(timeout, timeoutCallback);
-        socket.on('end', () => {
-          socket.setTimeout(0);
-          socket.setMaxListeners(Math.max(socket.getMaxListeners() - 1, 0));
-        });
-      });
     }
 
     // Send the request


lahirumaramba (approved):
Thank you for the fix! LGTM!

PR #2340, opened 2023-10-18
adding integration tests

pragatimodi:


test/integration/auth.spec.ts
```diff
@@ -1758,6 +1758,46 @@ describe('admin.auth', () => {
       });
     });
 
+    describe('Passkey config management operations', () => {
```
Xiaoshouzi-gh, line 1761: can we also add a disable test case?

pragatimodi: Disabling is not supported by the backend as of now.

Xiaoshouzi-gh: confirmed with backend disabling is an undone feature.  Resolving this.

Xiaoshouzi-gh (approved):
LGTM

PR #2341, opened 2023-10-18
user record changes for getAccountInfo()

pragatimodi:


src/auth/user-record.ts
```diff
@@ -357,6 +365,55 @@ export class MultiFactorSettings {
   }
 }
 
+/**
+ * Information about passkeys
```
kevinthecheung, line 369:
```suggestion
 * Interface representing a user-enrolled passkey.
```

src/auth/user-record.ts
```diff
@@ -582,6 +639,11 @@ export class UserRecord {
    */
   public readonly multiFactor?: MultiFactorSettings;
 
+  /**
+   * Information about Passkeys
```
kevinthecheung, line 643:
```suggestion
   * Passkey-related properties for the current user, if available.
```

kevinthecheung (approved):

src/auth/user-record.ts
```diff
@@ -56,6 +56,13 @@ export interface TotpInfoResponse {
   [key: string]: unknown;
 }
 
+export interface PasskeyInfoResponse {
+  name?: string;
+  credentialId?: string;
+  displayName?: string;
+  [key: string]: unknown;
```
renkelvin, line 63: What's this field used for?

renkelvin (approved):
LGTM with 1 issue.

src/auth/user-record.ts
pragatimodi: I followed convention looking at `MultiFactorInfoResponse` and `GetAccountInfoUserResponse`. Can be removed if we think it is redundant.

renkelvin: Yeah, it's useful for parsing unknown fields from backend which might not be the case here. Let's add it back when needed.

src/auth/user-record.ts
```diff
@@ -56,6 +56,12 @@ export interface TotpInfoResponse {
   [key: string]: unknown;
 }
 
+export interface PasskeyInfoResponse {
+  name?: string;
+  credentialId?: string;
```
Xiaoshouzi-gh, line 61: credentialId and name should never be empty correct?

test/unit/auth/user-record.spec.ts
```diff
@@ -185,6 +197,18 @@ function getUserJSON(tenantId?: string): object {
         },
       ],
     },
+    passkeyInfo: [
```
Xiaoshouzi-gh, line 200: can you add a undefined displayName test case?

Xiaoshouzi-gh (approved):
LGTM with nit comments

src/auth/user-record.ts
pragatimodi: yes

PR #2351, opened 2023-10-25
@google-cloud/firestore optional dependency ^6.6.0 -> ^6.8.0

joehan:
Bumping up @google-cloud/firestore dependency to the latest 6.x version. 

Per https://github.com/googleapis/nodejs-firestore/releases/tag/v6.7.0, multidb support was only added in 6.7.0. https://firebase.google.com/docs/reference/admin/node/firebase-admin.firestore.md#getfirestore says that we support this, so we should require at least 6.7.0. Currently, this silently write/reads from the default DB if you try to use a different DB and have firestore@6.6.0 in your node modules.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -101,21 +101,93 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.21.4",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.21.4.tgz",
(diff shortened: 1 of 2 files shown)


The-Don-Himself:
Actually would be great if this would be bumped to the latest 7.x versions but those contain a breaking change. I guess that would require a new major release. Not sure.

lahirumaramba:
Fixed in #2385 
We will bump `@google-cloud/firestore` to `v7` in the upcoming major release.

PR #2432, opened 2024-01-26
[chore] Bump `@actions/core` to ^1.10.1 to remove `set-output` warning and set action to use Node 20

jonathanedey:
- `@actions/core` ^1.9.1 -> ^1.10.1
- Set send-email action to use Node 20


lahirumaramba (approved):
Thanks!

PR #2451, opened 2024-02-07
[chore] Bump `mailgun.js` to `v10.1.0`

jonathanedey:
- Verified in https://github.com/firebase/firebase-admin-node/actions/runs/7821364789/job/21338134512?pr=2451


lahirumaramba (approved):
Thanks! LGTM

PR #2469, opened 2024-02-24
issue 2467: add async to send each loop to prevent local validation f…

DyosTheNerd:
### Discussion
adds async to send each loop to prevent local validation from throwing errors which could result in messages being sent and accepted by the api, while the function would still throw an error.

### Testing

adds a test with a successful message and one that does not pass local validation. Previously, this would throw an error.

### DIFF
--- a/src/messaging/messaging.ts
+++ b/src/messaging/messaging.ts
@@ -294,7 +294,7 @@ export class Messaging {
 
     return this.getUrlPath()
       .then((urlPath) => {
-        const requests: Promise<SendResponse>[] = copy.map((message) => {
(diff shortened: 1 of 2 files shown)


Doris-Ge (approved):
Thank you for adding this!

jonathanedey (approved):
LGTM! Thank you @DyosTheNerd for the contribution.

DyosTheNerd:
no probs :) @jonathanedey as branch is out of date, do you need me to update the branch, or will it automatically be done on merge from your side?

jonathanedey:
Yes could you update the branch. Updating on your side makes it easier in case there are any conflicts. Thanks!

PR #2458, opened 2024-02-13
Add public SSRC methods

erikeldridge:
Internal client updates.

This is based on PR #2457, which updated the API client.

### Discussion

Working with @lahirumaramba and @trekforever.

### Testing

Ran npm test and all tests pass.

PR #2478, opened 2024-02-27
Remove SSRC async initializer

erikeldridge:
Clarifies the name of in-app default when used in a server context.

PR https://github.com/firebase/firebase-admin-node/pull/2458 defined an async convenience method `getServerTemplate`. In practice, we find the sync alternative `initServerTemplate` much more user-friendly and use it exclusively, so this PR removes `getServerTemplate`.

### Discussion

Working with @lahirumaramba and @trekforever.

### Testing

Ran npm test and all tests pass.

Functionally tested using a local server.

### DIFF
--- a/src/remote-config/remote-config.ts
+++ b/src/remote-config/remote-config.ts
@@ -177,16 +177,6 @@ export class RemoteConfig {
   }
 
   /**
-   * Instantiates {@link RemoteConfigServerTemplate} and then fetches and caches the latest
-   * template version of the project.
-   */
-  public async getServerTemplate(options?: RemoteConfigServerTemplateOptions): Promise<RemoteConfigServerTemplate> {
-    const template = this.initServerTemplate(options);
-    await template.load();
-    return template;
-  }
-
-  /**
    * Synchronously instantiates {@link RemoteConfigServerTemplate}.
    */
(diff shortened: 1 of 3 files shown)


src/remote-config/remote-config.ts
```diff
@@ -176,16 +176,6 @@ export class RemoteConfig {
...
-  /**
-   * Instantiates {@link RemoteConfigServerTemplate} and then fetches and caches the latest
-   * template version of the project.
-   */
-  public async getServerTemplate(options?: RemoteConfigServerTemplateOptions): Promise<RemoteConfigServerTemplate> {
-    const template = this.initServerTemplate(options);
-    await template.load();
-    return template;
-  }
-
```
lahirumaramba, line 188: do we need to update the internal API proposals to reflect this change?

lahirumaramba (approved):
LGTM! thanks!

src/remote-config/remote-config.ts
erikeldridge: I was thinking this is a subset of the approved API, ie I'm not adding an unreviewed method, so it wasn't worth a full amendment, but I'll add a note to the doc, comment-in the approvers and see if they have any objections.

jenh (approved):

src/remote-config/remote-config.ts
erikeldridge: Amendment approved!

PR #2487, opened 2024-03-05
Support SSRC conditions

erikeldridge:
Add support for condition evaluation.

Note this is based on https://github.com/firebase/firebase-admin-node/pull/2496.

### Discussion

Working with @lahirumaramba and @trekforever.

### Testing

Ran npm test and all tests pass.

Functionally tested using a local server.

PR #2496, opened 2024-03-15
Remove product prefix from SSRC types

erikeldridge:
We have a Firebase AIP to not include product prefixes in type names.

### Discussion

There are already some types with the prefix, but we can avoid adding new types, as discussed with @lahirumaramba.

This change also renames the unused RemoteConfigServerCondition -> NamedServerCondition, in anticipation of https://github.com/firebase/firebase-admin-node/pull/2487.

### Testing

Ran npm test and all tests pass.

Functionally tested using a local server.


jenh (approved):
Looks good to me; also tested with a local server client--it successfully fetched all configured parameters from Remote Config without error.

lahirumaramba (approved):
LGTM!

PR #2505, opened 2024-03-22
Remove defaultConfig from public SSRC API

erikeldridge:
### Discussion

In practice, we only set default config using the initialization methods, so make the internal field private to simplify the API.

This wasn't a controversial field in API review.

### Testing

Tested by running `npm test`.

### API Changes

Makes `defaultConfigs` private.


lahirumaramba (approved):
LGTM! Thanks!

PR #2506, opened 2024-03-22
Differentiate options for sync and async SSRC initialization

erikeldridge:
### Discussion

It doesn't make sense to pass a cached template to `getServerTemplate` because the method loads a new template immediately.

I see a recommendation to define distinct interfaces for method options: https://github.com/microsoft/tsdoc/issues/19.

So, this change splits `ServerTemplateOptions` into `GetServerTemplateOptions` and `InitServerTemplateOptions`.

### Testing

Tested with `npm test`.

### API Changes

Splits `ServerTemplateOptions` into `GetServerTemplateOptions` and `InitServerTemplateOptions`.

### DIFF
--- a/src/remote-config/remote-config-api.ts
+++ b/src/remote-config/remote-config-api.ts
@@ -354,7 +354,21 @@ export interface ServerTemplateData {
 /**
  * Represents optional arguments that can be used when instantiating {@link ServerTemplate}.
  */
-export interface ServerTemplateOptions {
+export interface GetServerTemplateOptions {
+
+  /**
+   * Defines in-app default parameter values, so that your app behaves as
+   * intended before it connects to the Remote Config backend, and so that
+   * default values are available if none are set on the backend.
+   */
+  defaultConfig?: ServerConfig,
+}
+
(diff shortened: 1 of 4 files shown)


src/remote-config/remote-config-api.ts
```diff
@@ -354,7 +354,21 @@ export interface ServerTemplateData {
 /**
  * Represents optional arguments that can be used when instantiating {@link ServerTemplate}.
  */
-export interface ServerTemplateOptions {
+export interface GetServerTemplateOptions {
```
lahirumaramba, line 357: Does it makes sense to create a base type and extend from it?

```
export interface InitServerTemplateOptions extends GetServerTemplateOptions {
      template?: ServerTemplateData,
}
```

lahirumaramba (approved):
LGTM! Thanks!
Left one comment

src/remote-config/remote-config-api.ts
erikeldridge: Makes sense. I'll update it.

PR #2519, opened 2024-04-02
Define type-specific getters for SSRC

erikeldridge:
### Discussion

RC's existing SDKs define type-specific getters, like `getBoolean`. These aren't idiomatic for TS/JS, but have a couple advantages:
* RC param names, values and types are mutable remotely, so a simple object can’t guarantee a strict type for application logic. A formal schema would address this, but feels excessive for the common case. Type-specific methods are consistent with RC's current SDKs and ensure appropriate types for application logic.
* RC Android and iOS SDKs log events when personalized values are used. A method interface facilitates such additional functionality

This change defines an "internal" src directory for the new `ValueImpl` class, since we now have three internal classes. We can move the other two in there in a follow-up change.

### Testing

Unit tested via `npm test`. Functionally tested using a test server.

### API Changes

Updates the return type of the `evaluate` method to be a rich object, modeled after the [getters on the Web SDK](https://firebase.google.com/docs/reference/js/remote-config#functionremoteconfig_).


src/remote-config/internal/value-impl.ts
```diff
@@ -0,0 +1,54 @@
...
+ * @internal
+ */
+export class ValueImpl implements Value {
+  public static BOOLEAN_TRUTHY_VALUES = ['1', 'true', 't', 'yes', 'y', 'on'];
+  public static DEFAULT_VALUE_FOR_NUMBER = 0;
+  public static DEFAULT_VALUE_FOR_STRING = '';
+  constructor(
+        private readonly source: ValueSource,
+        private readonly value = ValueImpl.DEFAULT_VALUE_FOR_STRING) { }
+  asBoolean(): boolean {
```
trekforever, line 38: should we have a 'DEFAULT_VALUE_FOR_BOOLEAN' for the static behavior? Similar to https://github.com/firebase/firebase-js-sdk/blob/master/packages/remote-config/src/value.ts#L36-L39?

trekforever (approved):

src/remote-config/internal/value-impl.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Implements type-safe getters for parameter values.
+ * 
+ * Visible for testing.
+ * 
+ * @internal
+ */
+export class ValueImpl implements Value {
+  public static BOOLEAN_TRUTHY_VALUES = ['1', 'true', 't', 'yes', 'y', 'on'];
```
lahirumaramba, line 32: should we mark these as `readonly`?

src/remote-config/remote-config-api.ts
```diff
@@ -539,4 +539,98 @@ export interface ListVersionsOptions {
...
+ *   <li>"remote" indicates the value was defined by config produced by
+ *   evaluating a template.</li>
+ * </ul>
+ */
+export type ValueSource = 'static' | 'default' | 'remote';
+
+/**
+ * Defines the format for in-app default parameter values.
+ */
+export type DefaultConfig = { [key: string]: string | number | boolean };
```
lahirumaramba, line 636: Could this also contain `object` or `Value`?

lahirumaramba (approved):
LGTM! Thanks!

jenh (approved):

src/remote-config/remote-config-api.ts
erikeldridge: We could add `object`. I think it'd pair well with a `getObject<T>`, so those could be the same review. I think it'd be safe to add these after the initial launch.

I don't think `Value` would make sense because it's a tuple of the string value and an indication of where it came from (the "source"). If the customer is setting it, then the its source will always be "default". Setting a string value in `defaultConfig` accomplishes the same thing in a more user-friendly way.

src/remote-config/internal/value-impl.ts
erikeldridge: Makes sense. Will do.

PR #2520, opened 2024-04-02
Add functionality to accept a JSON template string when initializing a RC server template

trekforever:
### Discussion
During testing, we commonly stored a template as JSON. It was then inconvenient to parse that into an object of a specific type (TemplateData) just to accommodate the type expected by the option. 

This change updates `initServerTemplate` to also accept a JSON string, in addition to the existing ServerTemplateData.

### Testing

Unit tested via npm test

### API Changes

The `InitServerTemplateOptions.template` now accepts a union type `ServerTemplateData|string` instead of just `ServerTemplateData`

### DIFF
--- a/src/remote-config/remote-config-api.ts
+++ b/src/remote-config/remote-config-api.ts
@@ -354,7 +354,7 @@ export interface ServerTemplateData {
 /**
  * Represents optional arguments that can be used when instantiating {@link ServerTemplate}.
  */
-export interface GetServerTemplateOptions {
+export interface ServerTemplateOptions {
 
   /**
    * Defines in-app default parameter values, so that your app behaves as
@@ -362,13 +362,6 @@ export interface GetServerTemplateOptions {
    * default values are available if none are set on the backend.
    */
   defaultConfig?: ServerConfig,
-}
-
-/**
- * Represents optional arguments that can be used when instantiating
- * {@link ServerTemplate} synchonously.
- */
-export interface InitServerTemplateOptions extends GetServerTemplateOptions {
 
   /**
    * Enables integrations to use template data loaded independently. For
@@ -376,7 +369,7 @@ export interface InitServerTemplateOptions extends GetServerTemplateOptions {
    * caching template data and then using this option to initialize the SDK with
    * that data.
    */
-  template?: ServerTemplateData,
+  template?: ServerTemplateData|string,
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/remote-config/remote-config-api.ts
```diff
@@ -399,6 +399,12 @@ export interface ServerTemplate {
    * project's {@link ServerTemplate}.
    */
   load(): Promise<void>;
+
+  /** 
+   * Convenient method that returns the JSON string of the cached template data
+   * @returns A JSON-string of this object.
```
erikeldridge, line 405: Nit: I think we can drop the "Convenient method that" since this is now the only way to get template data.

src/remote-config/remote-config.ts
```diff
@@ -198,7 +199,19 @@ export class RemoteConfig {
...
     if (options?.template) {
-      template.cache = options?.template;
+      // Check and instantiates the template via a json string
+      if (isString(options?.template)) {
+        try {
+          template.cache = new ServerTemplateDataImpl(JSON.parse(options?.template));
+        } catch (e) {
+          throw new FirebaseRemoteConfigError(
+            'invalid-argument',
+            `Failed to parse the JSON string: ${options?.template}. ` + e
```
erikeldridge, line 209: Looking at the other errors we throw, we don't include the raw error (`e`) in the message. We should check w Lahiru if this is an intentional pattern, and if so, follow it here.

Nit: If we do include the error, it might be more readable to differentiate it from the other error message, something like "... Raw error: ${e}".

erikeldridge (changes requested):

src/remote-config/remote-config.ts
trekforever: We don't include it for most errors, but we do have it for the createTemplateFromJSON:
https://github.com/firebase/firebase-admin-node/blob/master/src/remote-config/remote-config.ts#L163-L166
(most of this logic is consistent with that method).  But will check with Lahiru when I request his review on this

src/remote-config/remote-config-api.ts
```diff
@@ -376,7 +376,7 @@ export interface InitServerTemplateOptions extends GetServerTemplateOptions {
    * caching template data and then using this option to initialize the SDK with
    * that data.
    */
-  template?: ServerTemplateData,
+  template?: ServerTemplateData|string,
 }
 
 /**
```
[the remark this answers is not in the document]
trekforever: Discussed offline, we're going to punt on making the cache private for now pending amendment 4 in the API proposal

src/remote-config/remote-config-api.ts
trekforever: Discussed offline, we're going to remove this method for now pending amendment 4 in the proposal doc

src/remote-config/remote-config.ts
erikeldridge: Ah. I overlooked that. Makes sense. Shipit 🚢

erikeldridge (approved):

lahirumaramba (approved):
Thank you! LGTM!

PR #2522, opened 2024-04-05
Add #set and #toJSON to RC ServerTemplate

trekforever:
### Discussion

Right now, we can set a RC server template via the constructor. Customers, however, would need to update the template multiple times. The current API exposes `template.cache`, but that feels more like an internal detail than an API now.

* Mark `ServerTemplate.cache` as `private`, and should only be used with-in the class
* Add a new `ServerTemplate.set` to allow for setting the template or templateJSON
* Add a `ServerTemplate.toJSON` to return JSON representation (`ServerTemplateData`) of the template
* Update `InitServerTemplate` to make use of the new set method.

### Testing

Added new unit tests for the new `set` method. Update existing tests to make use of `toJSON()` instead of the cache.

### API Changes

* `ServerTemplate.cache` is now private and cannot be accessed outside of the class
* New  `ServerTemplate.set` to allow for setting the template or templateJSON
* New `ServerTemplate.toJSON` to return JSON representation (`ServerTemplateData`) of the template

### DIFF
--- a/src/remote-config/remote-config-api.ts
+++ b/src/remote-config/remote-config-api.ts
@@ -365,6 +365,13 @@ export interface GetServerTemplateOptions {
 }
 
 /**
+ * Represents the type of a Remote Config server template that can be set on
+ * {@link ServerTemplate}. This can either be a {@link ServerTemplateData} object
+ * or a template JSON string.
+ */
+export type ServerTemplateDataType = ServerTemplateData | string;
+
+/**
  * Represents optional arguments that can be used when instantiating
  * {@link ServerTemplate} synchonously.
  */
@@ -378,13 +385,6 @@ export interface InitServerTemplateOptions extends GetServerTemplateOptions {
    */
   template?: ServerTemplateDataType,
 }
-
-/**
- * Represents the type of a Remote Config server template that can be set on
- * {@link ServerTemplate}. This can either be a {@link ServerTemplateData} object
- * or a template JSON string.
- */
-export type ServerTemplateDataType = ServerTemplateData | string;
 
 /**
  * Represents a stateful abstraction for a Remote Config server template.
--- a/src/remote-config/remote-config-api.ts
+++ b/src/remote-config/remote-config-api.ts
@@ -385,12 +385,6 @@ export interface InitServerTemplateOptions extends GetServerTemplateOptions {
  * Represents a stateful abstraction for a Remote Config server template.
  */
 export interface ServerTemplate {
-
-  /**
-   * Cached {@link ServerTemplateData}.
-   */
-  cache: ServerTemplateData;
-
   /**
    * Evaluates the current template to produce a {@link ServerConfig}.
    */
@@ -401,6 +395,17 @@ export interface ServerTemplate {
    * project's {@link ServerTemplate}.
    */
   load(): Promise<void>;
+
+  /**
+   * Sets and caches a {@link ServerTemplateData} or a JSON string representing
+   * the server template
+   */
+  set(template: ServerTemplateData | string): void;
+
+  /**
+   * Returns a JSON representation of {@link ServerTemplateData}
+   */
+  toJSON(): ServerTemplateData;
 }
 
 /**
--- a/src/remote-config/remote-config-api.ts
+++ b/src/remote-config/remote-config-api.ts
@@ -375,11 +375,16 @@ export interface InitServerTemplateOptions extends GetServerTemplateOptions {
    * example, customers can reduce initialization latency by pre-fetching and
    * caching template data and then using this option to initialize the SDK with
    * that data.
-   * The template can be initialized with either a {@link ServerTemplateData}
-   * object or a JSON string.
    */
-  template?: ServerTemplateData|string,
+  template?: ServerTemplateDataType,
 }
+
+/**
+ * Represents the type of a Remote Config server template that can be set on
+ * {@link ServerTemplate}. This can either be a {@link ServerTemplateData} object
+ * or a template JSON string.
+ */
+export type ServerTemplateDataType = ServerTemplateData | string;
 
 /**
  * Represents a stateful abstraction for a Remote Config server template.
@@ -400,7 +405,7 @@ export interface ServerTemplate {
    * Sets and caches a {@link ServerTemplateData} or a JSON string representing
    * the server template
    */
-  set(template: ServerTemplateData | string): void;
+  set(template: ServerTemplateDataType): void;
 
   /**
    * Returns a JSON representation of {@link ServerTemplateData}
--- a/src/remote-config/remote-config.ts
+++ b/src/remote-config/remote-config.ts
@@ -40,6 +40,7 @@ import {
   DefaultConfig,
   GetServerTemplateOptions,
   InitServerTemplateOptions,
+  ServerTemplateDataType,
 } from './remote-config-api';
 import { isString } from 'lodash';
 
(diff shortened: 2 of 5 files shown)


src/remote-config/remote-config.ts
```diff
@@ -333,6 +322,25 @@ class ServerTemplateImpl implements ServerTemplate {
...
   }
 
+  /**
+   * Takes in either a {@link ServerTemplateData} or a JSON string
+   * representing the template, parses it, and caches it.
+   */
+  public set(template: ServerTemplateData | string): void {
+    if (isString(template)) {
+      try {
+        this.cache = new ServerTemplateDataImpl(JSON.parse(template));
```
erikeldridge, line 332: Since `ServerTemplateDataImpl` throws `FirebaseRemoteConfigError` itself, I think this should be outside the try block, so we're not catching our own error, eg:

```
let parsed;
try {
  parsed = JSON.parse(template);
} catch (e) {
  // Transforms JSON parse errors to Firebase error.
  throw new FirebaseRemoteConfigError(
  'invalid-argument',
  `Failed to parse the JSON string: ${template}. ` + e);
}
// Throws template parse errors.
this.cache = new ServerTemplateDataImpl(parsed);
```

test/unit/remote-config/remote-config.spec.ts
```diff
@@ -951,6 +954,82 @@ describe('RemoteConfig', () => {
...
+
+      it('should throw if template is an invalid JSON', () => {
+        const jsonString = '{invalidJson: null}';
+        const initializedTemplate = remoteConfig.initServerTemplate();
+        expect(() => initializedTemplate.set(jsonString))
+          .to.throw(/Failed to parse the JSON string: ([\D\w]*)\./);
+      });
+
+      INVALID_PARAMETERS.forEach((invalidParameter) => {
+        sourceTemplate.parameters = invalidParameter;
```
erikeldridge, line 1011: it's unclear what this is testing. All the values in `INVALID_PARAMETERS` would produce valid JSON. For example, '{"parameters":null}' is valid.

Since we're currently also catching errors from `ServerTemplateDataImpl`, and `parameters: null` is an invalid template, I get the impression this is testing that.

I put a comment above about moving `ServerTemplateDataImpl` outside the try/catch, to differentiate JSON and template parsing errors.

erikeldridge (changes requested):

src/remote-config/remote-config.ts
trekforever: Good idea, I needed to add an else condition to set `parsed = template`, but updated 👍

test/unit/remote-config/remote-config.spec.ts
```diff
@@ -951,6 +954,82 @@ describe('RemoteConfig', () => {
...
+            valueType: 'STRING'
+          }
+        };
+        const initializedTemplate = remoteConfig.initServerTemplate();
+        initializedTemplate.set(template);
+        const parsed = initializedTemplate.toJSON();
+        const expectedVersion = deepCopy(VERSION_INFO);
+        expectedVersion.updateTime = new Date(expectedVersion.updateTime).toUTCString();
+        template.version = expectedVersion as Version;
+        expect(parsed).deep.equals(deepCopy(template));
```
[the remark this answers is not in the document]
trekforever: Nope, not needed, removed

test/unit/remote-config/remote-config.spec.ts
```diff
@@ -951,6 +954,82 @@ describe('RemoteConfig', () => {
...
+            description: 'Type of dog breed',
+            valueType: 'STRING'
+          }
+        };
+        const initializedTemplate = remoteConfig.initServerTemplate();
+        initializedTemplate.set(template);
+        const parsed = initializedTemplate.toJSON();
+        const expectedVersion = deepCopy(VERSION_INFO);
+        expectedVersion.updateTime = new Date(expectedVersion.updateTime).toUTCString();
+        template.version = expectedVersion as Version;
```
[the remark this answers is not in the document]
trekforever: Updated

test/unit/remote-config/remote-config.spec.ts
```diff
@@ -951,6 +954,82 @@ describe('RemoteConfig', () => {
...
+              value: 'shiba'
+            },
+            description: 'Type of dog breed',
+            valueType: 'STRING'
+          }
+        };
+        const initializedTemplate = remoteConfig.initServerTemplate();
+        initializedTemplate.set(template);
+        const parsed = initializedTemplate.toJSON();
+        const expectedVersion = deepCopy(VERSION_INFO);
```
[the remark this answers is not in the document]
trekforever: I get an test failure if we don't set the version:

```
      -    "updateTime": "Mon, 15 Jun 2020 16:45:03 GMT"
      +    "updateTime": "2020-06-15T16:45:03.541527Z"
```

I'm assuming it's bc the `Date.toJSON` method natively converts it to an ISOString() when we call `initializedTemplate.toJSON()` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toJSON) whereas we need a UTCString

test/unit/remote-config/remote-config.spec.ts
trekforever: Moved `ServerTemplateDataImpl` to outside the try/catch and updated the tests here to differentiate JSON and template parsing errors

test/unit/remote-config/remote-config.spec.ts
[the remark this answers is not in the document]
erikeldridge: Makes sense. Thanks!

test/unit/remote-config/remote-config.spec.ts
```diff
@@ -951,6 +954,89 @@ describe('RemoteConfig', () => {
...
+          }
+        };
+        template.version = {
+          ...deepCopy(VERSION_INFO),
+          updateTime: new Date(VERSION_INFO.updateTime).toUTCString()
+        } as Version;
+        const initializedTemplate = remoteConfig.initServerTemplate();
+        initializedTemplate.set(template);
+        const parsed = initializedTemplate.toJSON();
+        expect(parsed).deep.equals(template);
```
erikeldridge, line 976: This is very readable 👍

test/unit/remote-config/remote-config.spec.ts
```diff
@@ -951,6 +954,89 @@ describe('RemoteConfig', () => {
...
+        const INVALID_CONDITIONS: any[] = [null, '', 'abc', 1, true, {}];
+  
+        let sourceTemplate = deepCopy(SERVER_REMOTE_CONFIG_RESPONSE);
+        const jsonString = '{invalidJson: null}';
+        it('should throw if template is an invalid JSON', () => {
+          expect(() => remoteConfig.initServerTemplate({ template: jsonString }))
+            .to.throw(/Failed to parse the JSON string: ([\D\w]*)\./);
+        });
+  
+        INVALID_PARAMETERS.forEach((invalidParameter) => {
```
erikeldridge, line 1012: Thinking: it feels like we're re-testing the cases covered by the tests on `ServerTemplateDataImpl`. Looking at the tests, I see us already doing this on the tests for `initServerTemplate` and `load`. Makes me think there's an opportunity for a later change to extract `ServerTemplateDataImpl` to the internal dir so we can test it directly.

src/remote-config/remote-config.ts
```diff
@@ -333,6 +323,28 @@ class ServerTemplateImpl implements ServerTemplate {
       });
   }
 
+  /**
+   * Takes in either a {@link ServerTemplateData} or a JSON string
```
erikeldridge, line 327: Nit: we can now just reference ServerTemplateDataType

erikeldridge (approved):

src/remote-config/remote-config.ts
trekforever: Done 😀

src/remote-config/remote-config-api.ts
```diff
@@ -401,6 +400,17 @@ export interface ServerTemplate {
    * project's {@link ServerTemplate}.
    */
   load(): Promise<void>;
+
+  /**
+   * Sets and caches a {@link ServerTemplateData} or a JSON string representing
+   * the server template
+   */
+  set(template: ServerTemplateDataType): void;
```
lahirumaramba, line 408: Just curious (because I think either is fine) so we need a separate type for this or should this just be `ServerTemplateData | string` here?

lahirumaramba (approved):
LGTM! Thanks!

src/remote-config/remote-config-api.ts
trekforever: I think either works, I originally had it as `ServerTemplateData | string` but we are defining it in quite a few places across multiple files, so I moved it to a type 😃

PR #2523, opened 2024-04-05
Update SSRC to use validator instead of lodash for non-dev environment

erikeldridge:
### Discussion

Our tests have used lodash for awhile. PR #2520 added a usage outside of tests. When I run a test server that depends on the integration branch, I get a `Error: Cannot find module 'lodash'` error. So, I think we need to move lodash from `devDependencies` to `dependencies`. 

Moving lodash to `dependencies` and updating the test server to depend on this branch fixes the issue.

Perhaps interesting: I wasn't able to repro this issue by making the test server depend on a local copy of the firebase-admin-node, presumably because all dependencies end up in `node_modules` so lodash was available.

**Note:** based on review discussion, this change ended up replacing lodash's `isString` with `validator.isString`. 

### Testing

Unit tested by running `npm test` on this branch.

Functionally tested by updating a test server to depend on this branch and observing it was able to start without error.

### API Changes

None

### DIFF
--- a/etc/firebase-admin.remote-config.api.md
+++ b/etc/firebase-admin.remote-config.api.md
@@ -45,7 +45,7 @@ export interface InAppDefaultValue {
 
 // @public
 export interface InitServerTemplateOptions extends GetServerTemplateOptions {
-    template?: ServerTemplateData | string;
+    template?: ServerTemplateDataType;
 }
 
 // @public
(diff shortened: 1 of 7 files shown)


lahirumaramba (changes requested):
Correct me if I missed anything but the only usage of `lodash` in #2520 is `isString`. Could we use the existing `isString()` from the `validator` util instead? https://github.com/firebase/firebase-admin-node/blob/34f0ac27e286972995ba92bb1cb381d40d9689ac/src/utils/validator.ts#L79

erikeldridge:
> Correct me if I missed anything but the only usage of `lodash` in #2520 is `isString`. Could we use the existing `isString()` from the `validator` util instead?

Thank you. Yes, we can use that. I've updated the PR with that change.

lahirumaramba (approved):
Thank you! LGTM!

PR #2529, opened 2024-04-09
feat(rc): Add server side Remote Config support

erikeldridge:
### Discussion

We've been merging SSRC changes into the `ssrc` integration branch. This PR merges `ssrc` into `master`.

### Testing

Unit tested by running `npm test`.

Functionally tested by installing the Admin SDK in a test server using this branch.

### API Changes

Incrementally reviewed via ssrc-prefixed feature branches.


src/remote-config/remote-config.ts
```diff
@@ -254,6 +293,200 @@ class RemoteConfigTemplateImpl implements RemoteConfigTemplate {
...
+ * Remote Config dataplane template data implementation.
+ */
+class ServerTemplateDataImpl implements ServerTemplateData {
+  public parameters: { [key: string]: RemoteConfigParameter };
+  public parameterGroups: { [key: string]: RemoteConfigParameterGroup };
+  public conditions: NamedCondition[];
+  public readonly etag: string;
+  public version?: Version;
+
+  constructor(template: ServerTemplateData) {
```
trekforever, line 453: nit: template can be readonly too?

trekforever (approved):

src/remote-config/condition-evaluator-internal.ts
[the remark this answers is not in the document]
erikeldridge: We can, but I think we can safely do that in a follow-up change, since it's internal.

I also just noticed that we use the `@internal` annotation on the `RemoteConfig` constructor, which makes me wonder if the convention should just be that. Anything with the annotation is internal, regardless of the file's name or location. We could then remove the "internal" directory and file name suffixes.

src/remote-config/remote-config.ts
erikeldridge: That'd make sense, but I'm unsure it's worth another PR. The current approach is comparable to the existing `RemoteConfigTemplateImpl` constructor.

Aside, this reminds me of java's `final`. Most things are readonly, but applying a modifier to each variable ends up being more trouble than it's worth. A single designation, like `const`, is better.

Issue #1377, opened 2021-07-15
[FR] Allow credential file type "external_account" when calling `admin.credential.applicationDefault()`


thomasmburke:
## **Is your feature request related to a problem? Please describe.**

Developers would like the ability to initialize a Firebase app instance with a credential file type "external_account". An example of when credential file of type "external_account" is used is when following the instructions listed in the GCP public docs for using [AWS Identify Federated Credentials](https://cloud.google.com/iam/docs/access-resources-aws). This credential file type is supported by other Google APIs (e.g. @google-cloud/storage, @google-cloud/firestore, etc...) and is also supported by the Firebase Admin SDK in other languages (e.g. [Java](https://github.com/firebase/firebase-admin-java))

## **Describe the solution you'd like**

Allow the initialization of a Firebase app instance using credential file type "external_account". This will likely require an update to the [`credentialFromFile` function](https://github.com/firebase/firebase-admin-node/blob/d961c3f705a8259762a796ac4f4d6a6dd0992eb1/src/credential/credential-internal.ts#L427-L448)

## **Describe alternatives you've considered**

Developers are currently limited to:
- Leveraging the Firebase Java Admin SDK instead
- Using Secrets Manager/environment variables to store a service account key outside of code
- Calling a corresponding RESTful endpoint from within node.js via something like `http.request`

## **Additional context**

When a developer has configured GOOGLE_APPLICATION_CREDENTIALS to point to a credential file type of "external_account" (e.g. a AWS Identity Federation Credentials Config file) the below are the differences between the implementations for the Firebase Java Admin SDK and the Firebase Node.js Admin SDK:

**Node.js Details:**
 Developer initializes Firebase app instance with code similar to the below:
```
const fbApp = admin.initializeApp({
        credential: admin.credential.applicationDefault(),
        projectId: 'GCP_PROJECT_ID',
        databaseURL: 'https://GCP_PROJECT_ID.firebaseio.com'
    });
```
Which in turn calls the [`credentialFromFile` function](https://github.com/firebase/firebase-admin-node/blob/d961c3f705a8259762a796ac4f4d6a6dd0992eb1/src/credential/credential-internal.ts#L427-L448) and returns the following error: 'Invalid contents in the credentials file'

**Java Details:**
Developer initializes Firebase app instance with code similar to the below:
```
        GoogleCredentials credentials = GoogleCredentials.getApplicationDefault();

        FirebaseOptions options = FirebaseOptions.builder()
[...]

PR #2553, opened 2024-05-13
fix: Make ADC + human account work with firebase-admin

foxrafa:
This PR is intended to fix a current bug in firebase-admin.

Google Cloud recommends utilizing [Application Default Credentials (ADC)] (https://cloud.google.com/docs/authentication/application-default-credentials), however this feature does not fully work with the firebase-admin-node library when you want to utilize a human account (person@domain.com) instead of a service account.

Some GCP APIs require you to specify the quota project associated to the request, however, the current implementation of the firebase-admin-node library does not provide the project ID when making requests to GCP APIs, even if you set them through `gcloud auth application-default set-quota-project YOUR_PROJECT`, because the code that handles the requests never adds the project ID to the request. And you get errors like this:
```
Your application is authenticating by using local Application Default Credentials. The identitytoolkit.googleapis.com API requires a quota project, which is not set by default. To learn how to set your quota project, see https://cloud.google.com/docs/authentication/adc-troubleshooting/user-creds .
```

Therefore, this pull request aims to fix this issue by checking if a `projectId` was provided when calling the `initializeApp()` function and adds it to the `x-goog-user-project` header of requests made to GCP APIs, just like the [documentation](https://cloud.google.com/docs/authentication/troubleshoot-adc#user-creds-client-based) states it should be set.

This implementation is similar to how the (working) google-auth-library library handles ADC authentication ([src/auth/authclient.ts:270](https://github.com/googleapis/google-auth-library-nodejs/blob/b2b9676f933c012fb2cd1789ad80b927af0de07c/src/auth/authclient.ts#L270)):
```
    /**
     * Append additional headers, e.g., x-goog-user-project, shared across the
     * classes inheriting AuthClient. This method should be used by any method
     * that overrides getRequestMetadataAsync(), which is a shared helper for
     * setting request information in both gRPC and HTTP API calls.
     *
     * @param headers object to append additional headers to.
     */
    addSharedMetadataHeaders(headers) {
        // quota_project_id, stored in application_default_credentials.json, is set in
        // the x-goog-user-project header, to indicate an alternate account for
        // billing and quota:
        if (!headers['x-goog-user-project'] && // don't override a value the user sets.
            this.quotaProjectId) {
            headers['x-goog-user-project'] = this.quotaProjectId;
        }
        return headers;
    }
```
This should also solve some problems described in #1377

### DIFF
--- a/src/utils/api-request.ts
+++ b/src/utils/api-request.ts
@@ -818,6 +818,11 @@ export class AuthorizedHttpClient extends HttpClient {
       const authHeader = 'Authorization';
       requestCopy.headers[authHeader] = `Bearer ${token}`;
 
+      // Fix issue where firebase-admin does not specify quota project that is necessary for use when utilizing human account with ADC
+      if (!requestCopy.headers['x-goog-user-project'] && this.app.options.projectId) {
+        requestCopy.headers['x-goog-user-project'] = this.app.options.projectId
+      }
+
       if (!requestCopy.httpAgent && this.app.options.httpAgent) {
         requestCopy.httpAgent = this.app.options.httpAgent;
       }
--- a/src/utils/api-request.ts
+++ b/src/utils/api-request.ts
@@ -818,7 +818,8 @@ export class AuthorizedHttpClient extends HttpClient {
       const authHeader = 'Authorization';
       requestCopy.headers[authHeader] = `Bearer ${token}`;
 
-      // Fix issue where firebase-admin does not specify quota project that is necessary for use when utilizing human account with ADC
+      // Fix issue where firebase-admin does not specify quota project that is
+      // necessary for use when utilizing human account with ADC (RSDF)
       if (!requestCopy.headers['x-goog-user-project'] && this.app.options.projectId) {
         requestCopy.headers['x-goog-user-project'] = this.app.options.projectId
       }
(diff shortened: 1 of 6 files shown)


lahirumaramba (approved):
LGTM! Thank you!

lahirumaramba:
Hi @foxrafa could you please take a look at the lint errors in the failing CIs? Thanks!

foxrafa:
@lahirumaramba fixed it.

foxrafa:
Oh, all of the test cases are not expecting the x-goog-user-project header. Can I change them @lahirumaramba ?

lahirumaramba:
> @lahirumaramba fixed it.

Thanks! We also need to update the unit tests :)

foxrafa:
> > @lahirumaramba fixed it.
> 
> Thanks! We also need to update the unit tests :)

I updated them now.

rogsilva:
When will a new version with this change be released?

foxrafa:
@lahirumaramba any ETA on when this fix will be generally available?

lahirumaramba:
Thanks for the updating the tests! This issue will also be addressed in the credentials migration work https://github.com/firebase/firebase-admin-node/pull/2466 (which I think is the proper way to address this issue). For now, I am okay with us merging this PR as a stopgap

lahirumaramba:
Thanks @foxrafa ! this change is now included in the `v12.2.0` release!

erlichmen:
This fixes so much pain, thank you for that.

Issue #2564, opened 2024-05-20
jwks-rsa requires upgrade to fix known vulnerability.


SmashingQuasar:
### [READ] Step 1: Are you in the right place?

  * For issues related to __the code in this repository__ file a Github issue.
  * If the issue pertains to Cloud Firestore, read the instructions in the "Firestore issue"
    template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Operating System version: Debian 11
  * Firebase SDK version: 12.1.0
  * Firebase Product: global
  * Node.js version: 20.x
  * NPM version: 10.2.4

#### Steps to reproduce:

Simply run `pnpm audit`, it will return that the dependency `jwks-rsa` embarks a vulnerable dependency.
By upgrading `jwks-rsa`, the warning disappears.

PR #2565, opened 2024-05-20
build(deps): Upgrade jwks-rsa

SmashingQuasar:
# Description

This PR resolves the issue #2564.

Changes:
- The dependency `jwks-rsa` has been updated.
 
Note: When the repository was first cloned, without touching anything, simply by running : `npm install && npm test:unit` there was a broken test.
This test is still broken with the exact same message but all other are passing.
I did not touch it as tests behave exactly the same as before the update of the dependency.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,47 +1,135 @@
 {
   "name": "firebase-admin",
   "version": "12.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 3,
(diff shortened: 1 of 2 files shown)


lahirumaramba:
Thanks for the contribution! To fix the CIs and unblock the release today I have addressed this in https://github.com/firebase/firebase-admin-node/pull/2570

Issue #2552, opened 2024-05-10
12.1.0 errors with `Python is not set from command line or npm configuration`


k2xl:
### [READ] Step 1: Are you in the right place?

  * For issues related to __the code in this repository__ file a Github issue.
  * If the issue pertains to Cloud Firestore, read the instructions in the "Firestore issue"
    template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Operating System version: Mac Sonoma 14.0 (23A344)
  * Firebase SDK version: 12.1.0
  * Firebase Product: admin
  * Node.js version: 20.6.1
  * NPM version: 9.8.1

### [REQUIRED] Step 3: Describe the problem

Getting error when building on docker on Mac
`FROM node:20-alpine AS base`
https://github.com/sspenst/thinky.gg/pull/1101/files

Works fine on 12.0.0.

Error is below:
```
21.06 npm ERR! gyp info using node@20.13.1 | linux | arm64
21.06 npm ERR! gyp ERR! find Python 
21.06 npm ERR! gyp ERR! find Python Python is not set from command line or npm configuration
21.06 npm ERR! gyp ERR! find Python Python is not set from environment variable PYTHON
21.06 npm ERR! gyp ERR! find Python checking if "python3" can be used
21.06 npm ERR! gyp ERR! find Python - executable path is ""
21.06 npm ERR! gyp ERR! find Python - "" could not be run
21.06 npm ERR! gyp ERR! find Python checking if "python" can be used
21.06 npm ERR! gyp ERR! find Python - executable path is ""
21.06 npm ERR! gyp ERR! find Python - "" could not be run
21.06 npm ERR! gyp ERR! find Python 
21.06 npm ERR! gyp ERR! find Python **********************************************************
21.06 npm ERR! gyp ERR! find Python You need to install the latest version of Python.
21.06 npm ERR! gyp ERR! find Python Node-gyp should be able to find and use Python. If not,
21.06 npm ERR! gyp ERR! find Python you can try one of the following options:
[...]

PR #2603, opened 2024-06-18
fix: Replace `farmhash` with `farmhash-modern`

lahirumaramba:
Fixes: #2552

The experimental flag `experimental-wasm-bigint` is required on Node.js 14
Note: Node 14 and 16 are deprecated


erikeldridge (approved):

oscar-b:
Just wanted to add that using `farmhash-modern` seem to work much better, I previously had issues with Next.js v14, Node.js v21 on an ARM Mac. Thanks!

PR #2614, opened 2024-06-26
Remove Remote Config dependency on Long JS

erikeldridge:
### Discussion

The conditional logic used by the Remote Config config fetch endpoint hashes randomization IDs into 64 bit numbers (via an algorithm called Farmhash). JS numbers are limited to 53 bits, so we used the long.js package to emulate the fetch endpoint logic in the Admin SDK.

https://github.com/firebase/firebase-admin-node/pull/2603 migrated the Admin SDK to use a Farmhash library that produces   BigInt, which supports 64 bit numbers natively, so we can remove the long.js dependency. Additionally, the SDK currently converts the BigInt to a string before converting the string to a longjs object, which is clunky. Long JS also broke the SDK's ES6 compatibility.

### Testing

Farmhash produces an unsigned 64 bit number, which we convert to a signed number. Longjs does this by default. For BigInt, we need to do this explicitly via `BigInt.asIntN(64, unsignedNumber)`. I've extracted the hashing logic to a method we can test explicitly.

### API Changes

None


lahirumaramba (approved):
Thank you! LGTM!

PR #2648, opened 2024-07-29
fix: `getToken()` returns existing promise to a token if one exists instead of a new token.

jonathanedey:
- Concurrent calls to getToken() don't utilize cache because of a race condition which causes multiple token refreshes.
- Fixing by returning the existing promise to a token if one is still pending instead of requesting a new one.


lahirumaramba (approved):
Thanks!

PR #2656, opened 2024-07-31
Ssrc targeting numeric version fixes

kjelko:
Address minor bug fixes in numeric version handling

### DIFF
--- a/src/remote-config/condition-evaluator-internal.ts
+++ b/src/remote-config/condition-evaluator-internal.ts
@@ -293,6 +293,7 @@ function compareNumbers(
   return predicateFn(actual < target ? -1 : actual > target ? 1 : 0);
 }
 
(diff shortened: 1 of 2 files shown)


src/remote-config/condition-evaluator-internal.ts
```diff
@@ -293,6 +293,9 @@ function compareNumbers(
   return predicateFn(actual < target ? -1 : actual > target ? 1 : 0);
 }
 
+// Max number of segments a numeric version can have. This is enforced by the server as well.
+const MAX_LENGTH = 5;
```
erikeldridge, line 297: Thinking: three segments deviates from the semver spec, but consistency w the backend is more important, since RC is cross-platform.

erikeldridge (approved):

PR #2659, opened 2024-08-01
[FDL deprecation #1] Added mobileLinksConfig field for project and tenant configs

Xiaoshouzi-gh:
1. Added mobileLinksConfig field for project and tenant creation and/or updating workflows.
This will be later used in Firebase Dynamic links deprecation work.

2. Fixed some legacy failing tests.

* Any API changes after the review will be addressed in a different PR.
* This PR will be merge to the main branch of FDL deprecation work called `fdl-deprecation`.  Later when launched, `fdl-deprecation` will be merged  into `master` branch.

### DIFF
--- a/test/unit/auth/project-config.spec.ts
+++ b/test/unit/auth/project-config.spec.ts
@@ -51,6 +51,18 @@ describe('ProjectConfig', () => {
         },
       ],
     },
+    recaptchaConfig: {
+      emailPasswordEnforcementState: 'AUDIT',
+      managedRules: [ {
+        endScore: 0.2,
+        action: 'BLOCK'
+      } ],
+      recaptchaKeys: [ {
+        type: 'WEB',
+        key: 'test-key-1' }
+      ],
+      useAccountDefender: true,
+    },
     passwordPolicyConfig: {
       passwordPolicyEnforcementState: 'ENFORCE',
       forceUpgradeOnSignin: true,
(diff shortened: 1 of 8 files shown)


lahirumaramba (approved):
Thanks! LGTM!

test/unit/auth/project-config.spec.ts
```diff
@@ -51,6 +51,18 @@ describe('ProjectConfig', () => {
         },
       ],
     },
+    recaptchaConfig: {
```
pragatimodi, line 54: Do we need this? 
`updateProjectConfigRequest3` already has the same update config for recaptcha and I believe we are testing that functionality?

test/unit/auth/project-config.spec.ts
```diff
@@ -490,6 +524,7 @@ describe('ProjectConfig', () => {
     it('should set readonly property multiFactorConfig', () => {
       const expectedMultiFactorConfig = {
         state: 'DISABLED',
+        factorIds: [],
```
pragatimodi, line 527: Is this required when state is `DISABLED`?

Xiaoshouzi-gh: yes.  I believe there are some issue with mocha concurrency.  If I try to ran the single test file, it fails.

test/unit/auth/project-config.spec.ts
Xiaoshouzi-gh: Yes, this is also to fix the unit test failure.

PR #2665, opened 2024-08-05
feat(rc): SSRC targeting

kjelko:
Adds custom targeting evaluation logic


src/remote-config/condition-evaluator-internal.ts
```diff
@@ -23,7 +23,9 @@ import {
   NamedCondition,
   OrCondition,
   PercentCondition,
-  PercentConditionOperator
+  PercentConditionOperator,
+  CustomSignalCondition,
+  CustomSignalOperator,
```
lahirumaramba, line 28: I think we need to export `UserProvidedSignals` and `PredefinedSignals` here

src/remote-config/remote-config-api.ts
```diff
@@ -105,6 +105,11 @@ export interface OneOfCondition {
    * Makes this condition a percent condition.
    */
   percent?: PercentCondition;
+
+  /**
+   * Makes this a custom signal condition.
+   */
```
lahirumaramba, line 111: If we haven't done so already, could we get a TW review on the docstrings here (and below), please?

lahirumaramba (changes requested):

src/remote-config/remote-config-api.ts
kjelko: added @jenh!

src/remote-config/remote-config-api.ts
```diff
@@ -213,6 +218,123 @@ export interface PercentCondition {
...
+  STRING_EXACTLY_MATCHES = 'STRING_EXACTLY_MATCHES',
+
+  /**
+   * The target regular expression matches a portion of AT LEAST ONE of the
+   * actual values (or the entire string). The regex conforms to RE2 format.
+   * See https://github.com/google/re2/wiki/Syntax
+   */
+  STRING_CONTAINS_REGEX = 'STRING_CONTAINS_REGEX',
+
```
jenh, line 283:
```suggestion

  /**
   * Semantic Versioning custom signal operators evaluate one semantic
   * versioning (semver) value. Values must conform to the
   * [Semantic Versioning 2.0.0 spec]{@link https://semver.org/spec/v2.0.0.html}.
   */
   
```

src/remote-config/remote-config-api.ts
```diff
@@ -213,6 +218,123 @@ export interface PercentCondition {
...
+  customSignalOperator?: CustomSignalOperator;
+
+  /**
+   * The key of the signal set in the EvaluationContext
+   */
+  customSignalKey?: string;
+
+  /**
+   * A list of at most 100 target custom signal values. For numeric operators,
+   * this will have exactly ONE target value.
```
jenh, line 333: Do we still need this if we document the format/limitations above? If so, recommend:

```
  /**
   * Custom signal value(s) to evaluate. String operators support up to 100 values.
   * Numeric and semantic versioning operators support one value.
   */

File: src/remote-config/remote-config-api.ts [hunk 5]
```diff
@@ -213,6 +218,123 @@ export interface PercentCondition {
...
+  /**
+   * The actual value is greater than the target value.
+   */
+  NUMERIC_GREATER_THAN = 'NUMERIC_GREATER_THAN',
+
+  /**
+   * The actual value is greater than or equal to the target value.
+   */
+  NUMERIC_GREATER_EQUAL = 'NUMERIC_GREATER_EQUAL',
+
```
jenh, line 260: Is there anything else re:accepted syntax or limitations we want to add for the string operators?

```suggestion

  /**
   * String custom signal operators can evaluate up to 100 values.
   */

```

jenh (changes requested):

src/remote-config/remote-config-api.ts
```diff
@@ -213,6 +218,123 @@ export interface PercentCondition {
...
+  /**
+   * AT LEAST ONE of the target values is a substring of the actual custom
+   * signal value. Eg: "abc" contains the string "a", "bc".
+   */
```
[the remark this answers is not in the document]
kjelko: It's the *target values* (ie the things developers set up in the console when creating this condition) that can be multiple - not the *actual value* that's passed in for evaluation in the SDK.

so if you call `evaluate({myCustomSignal: 'foo'})` it would match for a condition like  `app.customSignal.myCustomSignal.contains(['bar', 'foo'])` 

As such, I think I'd prefer to keep the existing wording, though I can remove the awkward capitalizations if that helps

src/remote-config/remote-config-api.ts
```diff
@@ -213,6 +218,123 @@ export interface PercentCondition {
...
+  /**
+   * The actual value exactly matches AT LEAST ONE of the target values.
+   */
+  STRING_EXACTLY_MATCHES = 'STRING_EXACTLY_MATCHES',
+
+  /**
+   * The target regular expression matches a portion of AT LEAST ONE of the
+   * actual values (or the entire string). The regex conforms to RE2 format.
+   * See https://github.com/google/re2/wiki/Syntax
+   */
```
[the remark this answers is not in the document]
kjelko: yeah, agree that the "portion of" thing is confusing here. your example of using $ would work as expected in that 2.0Beta5 would *not* match your regex condition

src/remote-config/remote-config-api.ts
```diff
@@ -213,6 +218,123 @@ export interface PercentCondition {
...
+/**
+ * Defines supported operators for custom signal conditions.
+ */
+export enum CustomSignalOperator {
+
+  /**
+   * A catchall error case.
+   */
+  UNKNOWN = 'UNKNOWN',
+
```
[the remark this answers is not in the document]
kjelko: Per offline discussion, these limitations are enforced on the management planes (when defining these conditions). They aren't as relevant here in the SDK where the value that's passed in is one single value. As such, we won't highlight too much here to prevent any confusion

src/remote-config/remote-config-api.ts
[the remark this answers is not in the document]
jenh: A-ha, thanks, that makes sense, and sorry for the mix-up! One tiny nit, can you update `Eg:` to `E.g.,`? You can even put it in parens, like:

* Matches if at least one of the target values is a substring of the actual custom
* signal value (e.g., "abc" contains the strings "a" and "bc").

jenh (approved):
Looks good to me with the Eg to (e.g., ...) nit. :D

lahirumaramba (approved):

Issue #2692, opened 2024-09-03
[FR] Allow initializing app manually with impersonated service account


swftvsn:
**Is your feature request related to a problem? Please describe.**
We have a need to instantiate multiple instances of firebase admin to access multiple projects at the same time. This works nicely in production using:

```
const serviceAccountObject = JSON.parse(serviceAccountString)
const app1 = initializeApp({
  projectId: projectId,
  credential: cert(serviceAccountObject),
  databaseURL: databaseUrl,
  storageBucket: storageBucket
}, 'myNamedApp1')
```

However, it seems to be impossible to supply the impersonated service account json from local ADC login (obtained using `gcloud auth application-default login --impersonate-service-account local-run@my-project.iam.gserviceaccount.com`) as `ImpersonatedServiceAccountCredential` is not exported. 

We rolled our own, but it fails as `isApplicationDefault` uses instanceof to do the checking.

**Describe the solution you'd like**
Add a way to initialize from ADC logins manually: add a method (like `refreshToken` and `cert`) to `src/app/credential-factory.ts` that takes `impersonatedServiceAccountPathOrObject` etc as parameter.

```ts
const globalImpersonatedServiceAccountCreds: { [key: string]: ImpersonatedServiceAccountCredential } = {};
export function impersonatedServiceAccount(impersonatedServiceAccountPathOrObject: string | object, httpAgent?: Agent): Credential {
  const stringifiedImpersonatedServiceAccount = JSON.stringify(impersonatedServiceAccountPathOrObject);
  if (!(stringifiedImpersonatedServiceAccount in globalImpersonatedServiceAccountCreds)) {
    globalImpersonatedServiceAccountCreds[stringifiedImpersonatedServiceAccount] = new ImpersonatedServiceAccountCredential(
      impersonatedServiceAccountPathOrObject, httpAgent);
  }
  return globalImpersonatedServiceAccountCreds[stringifiedImpersonatedServiceAccount];
}

swftvsn:
Added PR https://github.com/firebase/firebase-admin-node/pull/2695 to fix this.

swftvsn:
@dconeybe could someone please weigh in on this one if it is feasible or not? This is blocking us from using more secure way to develop software locally.

lahirumaramba:
Hey @swftvsn thanks for your contribution on this. We are migrating our internal credentials handling to `google-auth-library` in #2466, which should address this issue.

[...]

PR #2695, opened 2024-09-04
Allow initializing app manually with impersonated service account

swftvsn:
Fix for https://github.com/firebase/firebase-admin-node/issues/2692


lahirumaramba:
Thank you for the contribution! This should now be fixed in https://github.com/firebase/firebase-admin-node/pull/2466

PR #2701, opened 2024-09-11
feat: Add Data Connect API

lahirumaramba:
RELEASE NOTES: Added a new API to perform administrative queries and mutations on a Firebase Data Connect service.

go/admin-node-data-connect


src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -0,0 +1,233 @@
...
+    const message = error.message || `Unknown server error: ${response.text}`;
+    return new FirebaseDataConnectError(code, message);
+  }
+}
+
+interface ErrorResponse {
+  error?: Error;
+}
+
+interface Error {
```
hsubox76, line 189: Is Error not an existing native type that comes with TS? Maybe ServerError or ResponseError? I guess the second one looks kind of silly as a child of ErrorResponse.

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -0,0 +1,233 @@
...
+
+  private toFirebaseError(err: RequestResponseError): PrefixedFirebaseError {
+    if (err instanceof PrefixedFirebaseError) {
+      return err;
+    }
+
+    const response = err.response;
+    if (!response.isJson()) {
+      return new FirebaseDataConnectError(
+        'unknown-error',
```
hsubox76, line 171: Why not use the map here? `DATA_CONNECT_ERROR_CODE_MAPPING.UNKNOWN`?

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -0,0 +1,233 @@
...
+ * @internal
+ */
+export class DataConnectApiClient {
+  private readonly httpClient: HttpClient;
+  private projectId?: string;
+
+  constructor(private readonly connectorConfig: ConnectorConfig, private readonly app: App) {
+    if (!validator.isNonNullObject(app) || !('options' in app)) {
+      throw new FirebaseDataConnectError(
+        'invalid-argument',
```
hsubox76, line 53: Suggest using the map here: `DATA_CONNECT_ERROR_CODE_MAPPING.INVALID_ARGUMENT`? and in other locations, to reduce chance of typos.

src/data-connect/data-connect-api.ts
```diff
@@ -0,0 +1,29 @@
...
+export interface ConnectorConfig {
+	location: string;
+	serviceId: string;
+}
+
+export interface ExecuteGraphqlResponse<GraphqlResponse> {
+  data: GraphqlResponse;
+}
+
+export interface GraphqlOptions<Variables> {
```
hsubox76, line 27: Are there any restrictions at all about what Variables and GraphqlResponse (above) can be? Should Variables always be a map/object? If so maybe a little more restriction like:
```
export interface GraphqlOptions<Variables extends Record<string, unknown>>
```
to indicate Variables must be a map/object (or in TS, Record)?

I think GraphqlResponse also seems like it is supposed to be a map/object?

src/data-connect/data-connect.ts
```diff
@@ -0,0 +1,106 @@
...
+ * @param query - The GraphQL read-only query.
+ * @param options - Optional options object when creating a new App Check Token.
+ *
+ * @returns A promise that fulfills with a `Something`.
+ */
+  public executeGraphqlRead<GraphqlResponse, Variables>(
+    query: string,
+    options?: GraphqlOptions<Variables>,
+  ): Promise<ExecuteGraphqlResponse<GraphqlResponse>> {
+    return this.client.executeGraphqlRead(query, options);
```
hsubox76, line 104: Does this call, and the one above, not give any errors for not providing the generics' types? If not, that's fine, just seems odd.

src/data-connect/data-connect-api-client-internal.ts
lahirumaramba: Good call! Thanks!

src/data-connect/data-connect-api-client-internal.ts
lahirumaramba: Thanks! Fixed

src/data-connect/data-connect-api-client-internal.ts
lahirumaramba: I don't think there is a specific reason to call it `Error`. It's just how we have used it in all the previous service implementations. I will rename this to `ServerError`. Thanks!

src/data-connect/data-connect-api.ts
lahirumaramba: Good call. Yes, Variables is always a map/object. I will update this to make it more restrictive

src/data-connect/data-connect.ts
lahirumaramba: The Generics are optional in these

src/data-connect/data-connect-api.ts
lahirumaramba: Discussed offline. Making the types more restrictive throws for plain interface types, unless they are contain index signatures. Leaving it as is for now.

markarndt (approved):

src/data-connect/data-connect-api.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/**
+ * Interface representing GraphQL options.
+ */
+export interface GraphqlOptions<Variables> {
+  /**
+   * Values for GraphQL variables provided in this query or mutation.
+   */
+  variables?: Variables;
```
maneesht, line 50: I wonder if we should make this required, and then have an additional signature that has `GraphqlOptions<undefined>` and accepts no arguments?

maneesht: Like: https://github.com/firebase/firebase-js-sdk/blob/main/packages/data-connect/src/api/query.ts#L91-L120
Curious to hare @hsubox76's opinion

lahirumaramba: I am not sure I understood correctly... `GraphqlOptions` here is an interface and the link to js-sdk points to function signatures. What would be the use-case here for marking `variables` in `GraphqlOptions` required?

maneesht: Say a user defines `GraphqlOptions<number>`, based on the `executeGraphql` function, if the user uses that type, then they can optionally not provide a number. Do we want that? I feel like if a developer defines a variables type, then they want to enforce that that type is being passed in, and a way to do that is to make the Variables required in the type itself, and if developers don't want to pass something in, they can also completely leave out the Variables object.
Here's an example:
Currently:
```ts
executeGraphqlRead<number>(myQuery); // this is valid
```

We could change the signature so that the user has to pass in a variable if they pass in a type. WDYT?

maneesht (approved):

PR #2712, opened 2024-09-24
fix(auth): Add missing `Math.floor()` when setting `validDuration` in `createSessionCookie()`

jonathanedey:
- Add missing `Math.floor()` when setting `validDuration` in `createSessionCookie()`
- Modified test to use an `expiresIn` value set to the millisecond to verify this resolves the issue.


lahirumaramba (approved):
Thanks Jonathan! LGTM

PR #2726, opened 2024-10-08
fix(functions): Allow both partial and full resource names to enqueue tasks in emulator

blidd-google:
The docstring for `getFunctions.taskQueue()` advertises that task queues can be referenced by both fully and partially qualified resource names, but a user reported that fully qualified resource names was not working. This PR updates the resource resolution logic to properly resolve to the corresponding task queue in the new Cloud Tasks emulator.

Fixes https://github.com/firebase/firebase-admin-node/issues/2725.

### DIFF
--- a/src/functions/functions-api-client-internal.ts
+++ b/src/functions/functions-api-client-internal.ts
@@ -92,7 +92,7 @@ export class FunctionsApiClient {
     }
 
     try {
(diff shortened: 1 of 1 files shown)


lahirumaramba (approved):
LGTM!

Question: does this issue only apply to the emulator url?

blidd-google:
> LGTM!
> 
> Question: does this issue only apply to the emulator url?

Yes, it should only apply to the new Cloud Tasks emulator!

PR #2737, opened 2024-10-23
chore: Update auth.spec.ts

lahirumaramba:
- Remove an accidental `describe.only(...`  tag that went in #2625


Xiaoshouzi-gh (approved):
LGTM and thank you!

PR #2739, opened 2024-10-23
fix: Pin `@firebase/database-compat` to `v1.0.8`

lahirumaramba:
- Pin `@firebase/database-compat` to `v1.0.8`
- Pin `@firebase/database-types` to `v1.0.5`

Fix related to https://github.com/firebase/firebase-js-sdk/issues/8583

RELEASE NOTES: This version pins `@firebase/database-compat` to `v1.0.8` and `@firebase/database-types` to `v1.0.5` to maintain compatibility with Node.js 14 and 16. Node.js 14 and 16 support is deprecated and Node.js 18+ will be required in the next major version.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,11160 @@
 {
   "name": "firebase-admin",
   "version": "12.6.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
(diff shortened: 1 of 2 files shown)


jonathanedey (approved):
LGTM!

PR #2743, opened 2024-10-25
Request timeout const name correction

tauksun:
Request timeout value correction while sending messaging requests in a batch


lahirumaramba:
Hey @tauksun Thank you for your contribution. This change is related to https://github.com/firebase/firebase-admin-node/pull/1999
Instead of changing the timeout back to 10 seconds we should rename the constant to `FIFTEEN_SECONDS_IN_MILLIS`

lahirumaramba (approved):

PR #2745, opened 2024-10-28
feat(fcm): Add directBootOk field to AndroidConfig

lahirumaramba:
Added the `directBootOk` field to the `AndroidConfig` type.


jonathanedey (approved):
LGTM Thanks!

PR #2760, opened 2024-11-05
change(rc): Update Remote Config condition evaluation hashing

kjelko:
Update remote config condition evaluation hashing


src/remote-config/condition-evaluator-internal.ts
```diff
@@ -332,4 +322,4 @@ function compareSemanticVersions(
     if (version1[i] > version2[i]) return predicateFn(1);
   }
   return false;
-}
+}
```
lahirumaramba, line 325: nit: missing new line

lahirumaramba (approved):
LGTM! Thank you!

src/remote-config/condition-evaluator-internal.ts
```diff
@@ -173,18 +173,8 @@ export class ConditionEvaluator {
...
-    let hash64 = BigInt.asIntN(64, farmhash.fingerprint64(seededRandomizationId));
-
-    // Manually negate the hash if its value is less than 0, since Math.abs doesn't
-    // support BigInt.
-    if (hash64 < 0) {
-      hash64 = -hash64;
-    }
-
-    return hash64;
+    const hex = createHash('sha256').update(seededRandomizationId).digest('hex');
```
erikeldridge, line 176: Thinking (no action req'd): we had a 64 bits hash before, so we don't need 256 bits, but sha256 is broadly available and there's no noticeable performance difference, so sha256 is fine.

src/remote-config/condition-evaluator-internal.ts
```diff
@@ -173,18 +173,8 @@ export class ConditionEvaluator {
...
 
   static hashSeededRandomizationId(seededRandomizationId: string): bigint {
-    // For consistency with the Remote Config fetch endpoint's percent condition behavior
-    // we use Farmhash's fingerprint64 algorithm and interpret the resulting unsigned value
-    // as a signed value.
-    let hash64 = BigInt.asIntN(64, farmhash.fingerprint64(seededRandomizationId));
-
-    // Manually negate the hash if its value is less than 0, since Math.abs doesn't
-    // support BigInt.
-    if (hash64 < 0) {
```
erikeldridge, line 183: Thinking (no action req'd): we're now directly instantiating the bigint with the hash, rather than coercing it to behave like a 64bit signed Java long, so it will always be positive.

erikeldridge (approved):

PR #2761, opened 2024-11-05
fix: Set the Quota Project ID only for ADC human accounts

lahirumaramba:
- The quota project id should only be set for Application default credentials with human accounts.
- The SDK will use the quota project id found in credentials (through google-auth-client).
- `GOOGLE_CLOUD_QUOTA_PROJECT` overrides the project id found in credentials.

This PR fixes #2658

PR #2763, opened 2024-11-07
chore: Add `X-Goog-Api-Client` metric header to outgoing authorized http requests

jonathanedey:
Adds headers to all authorized http requests google backend endpoints.

Does not add headers to:
- Firestore
- Storage
- RTDB (database.compact)

### DIFF
--- a/src/utils/api-request.ts
+++ b/src/utils/api-request.ts
@@ -1112,12 +1112,6 @@ export class AuthorizedHttp2Client extends Http2Client {
       const authHeader = 'Authorization';
       requestCopy.headers[authHeader] = `Bearer ${token}`;
 
-      // Fix issue where firebase-admin does not specify quota project that is
-      // necessary for use when utilizing human account with ADC (RSDF)
-      if (!requestCopy.headers['x-goog-user-project'] && this.app.options.projectId) {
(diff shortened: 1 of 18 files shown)


src/utils/api-request.ts
```diff
@@ -1108,6 +1112,14 @@ export class AuthorizedHttp2Client extends Http2Client {
       const authHeader = 'Authorization';
       requestCopy.headers[authHeader] = `Bearer ${token}`;
 
+      // Fix issue where firebase-admin does not specify quota project that is
+      // necessary for use when utilizing human account with ADC (RSDF)
+      if (!requestCopy.headers['x-goog-user-project'] && this.app.options.projectId) {
+        requestCopy.headers['x-goog-user-project'] = this.app.options.projectId
+      }
```
lahirumaramba, line 1119: This is a bug in the current SDK and it should only be set for ADC human accounts. I have a fix coming up here https://github.com/firebase/firebase-admin-node/pull/2761

You can remove this part from this PR for now

lahirumaramba (approved):
LGTM! Left one comment

src/utils/api-request.ts
jonathanedey: Done!

Issue #2769, opened 2024-11-13
There is an error in the type defined in the ServiceAccount


k-redstone:
### Describe your environment

  * Operating System version:  window 11
  * Firebase SDK version: firebase-admin 13.0.0
  * Firebase Product: credential, auth
  * Node.js version: 20.12.0
  * NPM version: 10,5.0

### [REQUIRED] Step 3: Describe the problem

I followed the example provided for the admin.credential.cert() function, but encountered an issue.

I'm trying to initialize the app with admin.initializeApp() using Firebase Admin SDK version 13. However, there seems to be a discrepancy between the type defined in ServiceAccount and the key values actually used for authentication.

#### Steps to reproduce:

The following code causes the following error.

> The incoming JSON object does not contain a private_key field
> The incoming JSON object does not contain a client_email field

as-is
```typescript
import admin, { ServiceAccount } from "firebase-admin";

const serviceAccount: ServiceAccount = {
  projectId: process.env.NEXT_PUBLIC_FIREBASE_PROJECT_ID,
  privateKey: process.env.FIREBASE_PRIVATE_KEY?.replace(/\\n/g, "\n"),
  clientEmail: process.env.FIREBASE_CLIENT_EMAIL,
};

if (!admin.apps.length) {
  admin.initializeApp({
    credential: admin.credential.cert(serviceAccount),
    databaseURL: `https://${process.env.NEXT_PUBLIC_FIREBASE_PROJECT_ID}.firebaseio.com`,
  });
}

export default admin;
```

So, after modifying the code as follows, it worked correctly.
```typescript
import admin from 'firebase-admin'

const serviceAccount = {
  projectId: process.env.NEXT_PUBLIC_FIREBASE_PROJECT_ID,
  client_email: process.env.FIREBASE_CLIENT_EMAIL,
  private_key: process.env.FIREBASE_PRIVATE_KEY?.replace(/\\n/g, '\n'),
}

if (!admin.apps.length) {
  admin.initializeApp({
    credential: admin.credential.cert(serviceAccount),
    databaseURL: `https://${process.env.NEXT_PUBLIC_FIREBASE_PROJECT_ID}.firebaseio.com`,
  })
} else {
  admin.app()
}

export default admin
```

#### Relevant Code:

In my opinion, the transition to version 13.0 involved switching to use google-auth-library, which seems to be causing this issue.

maybe auth.fromJSON() cause error

src/app/credential-internal.ts  456 line
```typescript
[...]

Issue #2778, opened 2024-11-19
13.0.0 - Firebase Credential -The incoming JSON object does not contain a client_email field


mustafa-barakat:
### Describe your environment

  * Operating System version: Docker on Windows
  * Firebase SDK version: 13.0.0
  * Firebase Product: Firebase Admin
  * Node.js version: 18+
  * NPM version: 10.9

#### Steps to reproduce:

- Upgrade to firebase admin to 13.0.0
- When using initializeApp with credential and adding (projectId,clientEmail,privateKey)

This doesn't happen in previous versions

Error
message: 'Credential implementation provided to initializeApp() via the "credential" property failed to fetch a valid Google OAuth2 access token with the following error: "The incoming JSON object does not contain a client_email field"

#### Relevant Code:

```
import { initializeApp } from 'firebase-admin/app';
import { credential } from 'firebase-admin';

initializeApp({
      credential: credential.cert({
        projectId: 'PROJECTID',
        clientEmail:
          'EMAIL,
        privateKey:
          'PRIVKEY',
      }),
    });
```

ivstiv:
I am running into the exact same issue after the release of v13.

- Node 20
- Firebase Admin 13.0.0, 
- Operating System version: node:20-bullseye-slim (debian docker)

This is how I am initialising my app:
```ts
firebaseAdmin.initializeApp({
    credential: firebaseAdmin.credential.cert({
      projectId: "from-env",
      clientEmail: "from-env",
      privateKey: "from-env",
    }),
  },
  "label-here",
);
```

Probably related?: https://github.com/firebase/firebase-admin-node/pull/2466

mustafa-barakat:
> I am running into the exact same issue after the release of v13.
> 
> * Node 20
> * Firebase Admin 13.0.0,
> * Operating System version: node:20-bullseye-slim (debian docker)
> 
> This is how I am initialising my app:
> 
> ```ts
> import firebaseAdmin from "firebase-admin";
> 
> firebaseAdmin.initializeApp({
>     credential: firebaseAdmin.credential.cert({
>       projectId: "from-env",
>       clientEmail: "from-env",
>       privateKey: "from-env",
>     }),
>   },
>   "label-here",
> );
> ```
> 
> Probably related?: #2466

Yes, and for some reason tests were replaced

mustafa-barakat:
Related #2769

lahirumaramba:
[...]

PR #2779, opened 2024-11-19
fix: Fix the property names when initializing the SDK with the `ServiceAccount` type

lahirumaramba:
Fixes: #2769, #2778


jonathanedey (approved):
LGTM Thanks!

PR #2794, opened 2024-12-02
fix: equality comparison for semantic version when number of segments = max allowed segments

### LINKED ISSUE #2793: Semantic version equality comparison fails when number of segments is equal to the maximum number of segments allowed

  * For issues related to __the code in this repository__ file a Github issue.
  * If the issue pertains to Cloud Firestore, read the instructions in the "Firestore issue"
    template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

  * Operating System version: `macOS Sonoma 14.7.1`
  * Firebase SDK version: `13.0.1`
  * Firebase Product: `remoteconfig`
  * Node.js version: `v.23.1.0`
  * NPM version: `10.9.0`

#### Steps to reproduce:

1. Create the following condition using custom signals as (operator can be `>=`, `=` or `<=`) : 

image

2. When the value of the custom signal is set to `1.2.3.4.5`, the evaluation returns `false` instead of `true`

(This is seen when the semantic version has the [maximum number of segments](https://github.com/firebase/firebase-admin-node/blob/83585498d9bd48b9ba715a882a3b9c951ca8422d/src/remote-config/condition-evaluator-internal.ts#L293), here `5`)

Namyalg:

### Testing

  * Make sure all existing tests in the repository pass after your change [Verified all existing tests pass]
  * If you fixed a bug or added a feature, add a new test to cover your code [Unit tests added]

Fixes #2793

### DIFF
--- a/test/unit/remote-config/condition-evaluator.spec.ts
+++ b/test/unit/remote-config/condition-evaluator.spec.ts
@@ -1114,6 +1114,7 @@ describe('ConditionEvaluator', () => {
           { targets: ['5.12.3'], actual: '5.11.9', outcome: true },
           { targets: ['5.12.3'], actual: '5.12.3', outcome: true },
           { targets: ['5.12.3'], actual: '5.12.9', outcome: false },
+          { targets: ['5.6.7.8.9'], actual: '5.6.7.8.9', outcome: true },
           invalidNumericSignalTestCase,
         ];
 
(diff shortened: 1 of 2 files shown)


test/unit/remote-config/condition-evaluator.spec.ts
```diff
@@ -1127,6 +1128,8 @@ describe('ConditionEvaluator', () => {
           { targets: ['5.0'], actual: 5.0, outcome: true },
           { targets: ['5.12.3'], actual: '5.12.9', outcome: false },
           { targets: ['5.12.3'], actual: '5.12.3.0.0.0.0', outcome: false },
+          { targets: ['5.6.7.8.9'], actual: '5.6.7.8.9', outcome: true },
```
kjelko, line 1131: Consider adding a test with 5 segments that should not be equal, ie 5.6.7.8.9 and 4.5.6.7.8

kjelko (approved):
Thanks for taking this on!

PR #2801, opened 2024-12-05
fix(auth): Fix credentials issue in createToken() when using ADC

### LINKED ISSUE #2800: 13.x has broken auto finding service account for auth.createCustomToken

  * Operating System version: OSX 15.1.1
  * Firebase SDK version: 13.0.1
  * Firebase Product: auth (auth, database, storage, etc)
  * Node.js version: 20.11.1
  * NPM version: 10.9.1

#### Steps to reproduce:

What happened? How can we make the problem occur?

Prior to 13.x and the switch to google auth, you were able to run code like this

```typescript
import { initializeApp } from "firebase-admin/app"
import { getAuth } from "firebase-admin/auth"

initializeApp()
getAuth().createCustomToken("test")
```

and create a custom token fine.  I was able to do it locally, without error.

Now, we get errors locally like

> Failed to determine service account. Make sure to initialize the SDK with a service account credential. Alternatively specify a service account with iam.serviceAccounts.signBlob permission. Original error: Error: Error while making request: getaddrinfo ENOTFOUND metadata. Error code: ENOTFOUND

This is very similar to the hoops introduced in 13.x to get the projectId.

To get everything properly working with each other, you need to do something like unnecessarily create the google auth, wait for the projectId/for it to be "filled", and then set the service account that that had auto set in the `initializeApp` function,

```typescript
import { initializeApp } from "firebase-admin/app"
import { GoogleAuth } from "google-auth-library"

const gAuth = new GoogleAuth()
const projectId = await gAuth.getProjectId()
const jsonContent = gAuth.jsonContent as { client_email: string }
const app = initializeApp({ serviceAccountId: jsonContent.client_email })
```

All of the above seems like a big regression to many working features which needs to be fixed, or much of the documentation needs updating as running just `initializeApp` now does not set up your app fully like it did prior to 13.x, or plainly just a bug.

--- edit ---

My creds are loaded from my .env file, via

> GOOGLE_APPLICATION_CREDENTIALS=/Users/Shared/PATH_TO_SERVICE_FILE.json

The exact same service account and env vars are set, when testing against pre 13 and 13.x, and works fine with the appropriate code above, no issue with env vars, nor service account perms.

lahirumaramba:
fixes #2800

Instead of passing in an explicit service account name `IAMSigner` now takes in an app instance. This way we can use credentials (Google Auth) instance to get the service account name

### DIFF
--- a/src/utils/crypto-signer.ts
+++ b/src/utils/crypto-signer.ts
@@ -22,6 +22,7 @@ import { AuthorizedHttpClient, HttpRequestConfig, HttpClient, RequestResponseErr
 
 import { Algorithm } from 'jsonwebtoken';
 import { ErrorInfo } from '../utils/error';
+import * as utils from '../utils/index';
 import * as validator from '../utils/validator';
 
 const ALGORITHM_RS256: Algorithm = 'RS256' as const;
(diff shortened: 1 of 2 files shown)


jonathanedey (approved):
LGTM Thanks!

PR #2802, opened 2024-12-05
Use `crypto.randomUUID()` instead of `uuid` module

nikeee:
The `crypto` global is available in all supported Node.js versions as well as [in all major browsers](https://caniuse.com/mdn-api_crypto_randomuuid). The uuid module is rendered unnecessary for uuidv4.


nikeee:
The main motivation behind this PR was to reduce traffic in the realms of [e18e](https://e18e.dev). firebaes-admin has around 1466130 downloads per week. So requiring a 3.7 kB package results in 5.5GB/week, which is approx. 286 GB per year.

PR #2829, opened 2025-01-09
Adds support for exporting a serialized config response from the RC server side SDK.

kjelko:
Adds support for exporting a serialized config response from the RC server side SDK.


src/remote-config/remote-config.ts
```diff
@@ -613,3 +618,58 @@ class VersionImpl implements Version {
...
+    return this.response;
+  }
+
+  private processEtag(config: { [key: string]: string }, app: App): string {
+    const configJson = JSON.stringify(config);
+    let hash = 0;
+    for (let i = 0; i < configJson.length; i++) {
+      const char = configJson.charCodeAt(i);
+      hash = (hash << 5) - hash + char;
+      hash |= 0;
```
ashish-kothari, line 669: What is the significance of this operation?

src/remote-config/remote-config.ts
```diff
@@ -613,3 +618,58 @@ class VersionImpl implements Version {
...
+  public toJSON(): FetchResponseData {
+    return this.response;
+  }
+
+  private processEtag(config: { [key: string]: string }, app: App): string {
+    const configJson = JSON.stringify(config);
+    let hash = 0;
+    for (let i = 0; i < configJson.length; i++) {
+      const char = configJson.charCodeAt(i);
+      hash = (hash << 5) - hash + char;
```
ashish-kothari, line 668: This is quite cool. Should we add a comment describing the equation and also that we're trying to replicate the `String.hashcode()` impl here for context? I see that the java hashcode implementation mentions it.

ashish-kothari (approved):
LGTM with minor comments.

src/remote-config/remote-config.ts
```diff
@@ -613,3 +618,58 @@ class VersionImpl implements Version {
...
+        eTag: currentEtag,
+        config,
+      }
+    }
+  }
+
+  /** 
+   * @returns JSON representation of the fetch response that can be consumed
+   * by the RC client SDK.
+   */
```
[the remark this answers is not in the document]
kjelko: Hmm, I don't actually know - none of the other public methods are annotated with an `@public`...

src/remote-config/remote-config.ts
kjelko: Added a comment per below comment

src/remote-config/remote-config.ts
kjelko: Comment added

src/remote-config/remote-config.ts
```diff
@@ -613,3 +618,58 @@ class VersionImpl implements Version {
...
+/**
+ * Represents a fetch response that can be used to interact with RC's client SDK.
+ */
+export class RemoteConfigFetchResponse {
+  private response: FetchResponseData;
+
+  /**
+   * @param app - The app for this RemoteConfig service.
+   * @param serverConfig - The server config for which to generate a fetch response.
+   * @param requestEtag - A request eTag with which to compare the current response.
```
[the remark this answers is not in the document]
kjelko: That's correct - this would come in from a client request. it should be the etag of the most recent config that the client has. Since our internal servers do this, it makes sense to allow customer servers to do it as well

test/unit/remote-config/remote-config.spec.ts
```diff
@@ -1391,6 +1425,65 @@ describe('RemoteConfig', () => {
     });
   });
 
+  describe('RemoteConfigFetchResponse', () => {
+    it('should return a 200 response with no etag', () => {
```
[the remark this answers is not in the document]
kjelko: sure thing!

jenh (approved):

src/remote-config/remote-config-api.ts
```diff
@@ -727,6 +727,40 @@ export interface ServerConfig {
...
+   * implementation.
+   */
+  status: number;
+
+  /**
+   * Defines the ETag response header value.
+   * 
+   * This is consistent with Remote Config's server eTag implementation.
+   */
+  eTag: string;
```
lahirumaramba, line 758: nit: `eTag` is defined optional in the original API proposal (could be a typo in the proposal) just pointing out

lahirumaramba (approved):

src/remote-config/remote-config-api.ts
kjelko: Ah, hmm yeah it should be optional. It's true the implementation here doesn't ever set this as undefined - but that's just because we're offloading error handling to the developer. If they wanted to use this type to return an error response, they should be able to leave the eTag off.

Issue #2754, opened 2024-11-04
Authentication error in tasks firebase emulator on `enqueue`


jootten:
### Environment

  * Operating System version: Fedora Linux 40
  * Firebase SDK version: 6.0.1
  * Firebase Product: auth, tasks, emulator
  * Node.js version: 18
  * NPM version: 10.2.3
  * Firebase Tools: 13.23.1

### Problem

Authentication fails in the firebase tasks emulator on

```typescript
await queue.enqueue(...)  \\ full sample below
```

with the following error:
`Error scheduling task FirebaseAppError: Credential implementation provided to initializeApp() via the "credential" property failed to fetch a valid Google OAuth2 access token with the following error: "Error fetching access token: Error while making request: getaddrinfo ENOTFOUND metadata.google.internal. Error code: ENOTFOUND".`

We were able to reproduce the bug reported here https://github.com/firebase/firebase-tools/issues/7821 with the latest SDK and firebase tools versions.

**The error does not occur when application default credentials are available and is thus easily overlooked.**

#### Steps to reproduce:

1. Setup firebase project as in https://github.com/firebase/firebase-tools/issues/7821.
2. Start the emulator with the code below.
```bash
npm run build && npx firebase emulators:start --project='demo-project' --debug
```
3. Enqueue a task by executing the `testOnRequest` function (`http://127.0.0.1:5001/demo-project/us-central1/testOnRequest`).

[emulator-debug.log](https://github.com/user-attachments/files/17620329/debug.log)

#### Relevant Code:

Reproducible example that is even more minimal than in the mentioned issue:
```typescript
import { onRequest } from "firebase-functions/v2/https";
import { getFunctions } from "firebase-admin/functions";
import { onTaskDispatched } from "firebase-functions/v2/tasks";
import * as logger from "firebase-functions/logger";
import { initializeApp } from "firebase-admin/app";

// Initialize the Firebase app
initializeApp()

// The http function
export const testOnRequest = onRequest(async (request, response) => {
  const taskPayload = {
    foo: "bar",
  };

  const taskFunctionName = `testOnTaskDispatched`;
  const queue = getFunctions().taskQueue(taskFunctionName);

  try {
    await queue.enqueue(taskPayload);
    response.send("Success. Hello from HTTP onRequest!");
  } catch (error) {
    console.error("Error scheduling task", error);
    response.status(500).send("Error scheduling task");
    return;
  }
});

// The task function
[...]

PR #2857, opened 2025-02-24
fix(functions): Use emulated credentials when connecting to the emulator

lahirumaramba:
Fixes: #2754 

Added a new `FunctionsHttpClient` class that extends `AuthorizedHttpClient` to handle credentials when communicating with the Task Queue Emulator.


jonathanedey (approved):
LGTM, Thanks!

PR #2868, opened 2025-03-10
fix(fcm): Wrap HTTP/2 session errors in promise

jonathanedey:
RELEASE NOTE: Fixed a issue which caused session errors not to be thrown.

* Promise wrapped HTTP/2 session errors allowing them to be caught as a `FirebaseMessagingSessionError`
* Added `pendingBatchResponse` field to new `FirebaseMessagingSessionError` to allow access to the status of messages which were in flight at error time.

### DIFF
--- a/src/messaging/messaging.ts
+++ b/src/messaging/messaging.ts
@@ -206,10 +206,22 @@ export class Messaging {
         MessagingClientErrorCode.INVALID_ARGUMENT, 'dryRun must be a boolean');
     }
 
-    const http2SessionHandler = this.useLegacyTransport ? undefined : new Http2SessionHandler(`https://${FCM_SEND_HOST}`)
+    // const http2SessionHandler = this.useLegacyTransport ? undefined : new Http2SessionHandler(`https://${FCM_SEND_HOST}`)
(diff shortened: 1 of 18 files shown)


src/messaging/messaging.ts
```diff
@@ -206,48 +208,71 @@ export class Messaging {
...
-          successCount,
-          failureCount: responses.length - successCount,
-        };
+        if (http2SessionHandler) {
+          let sendResponsePromise: Promise<PromiseSettledResult<SendResponse>[]>;
+          return new Promise((resolve: (result: PromiseSettledResult<SendResponse>[]) => void, reject) => {
+            // Start session listeners
+            http2SessionHandler.invoke().catch((error) => {
+              const pendingBatchResponse = 
+                sendResponsePromise ? sendResponsePromise.then(this.parseSendResponses) : undefined;
```
lahirumaramba, line 221: Just thinking out loud, if a session error occurs before any of the requests have gone through would `pendingBatchResponse` be undefined?

lahirumaramba (approved):
Thanks Jonathan! Added one comment

src/messaging/messaging.ts
jonathanedey: Yes, this would be undefined.

Issue #2863, opened 2025-03-02
Messaging issue


hussenIbrahim:
Hello 
I have issue when sending notification via firebase admin, some functionality  doesn't  working properly 
[Original issue link](https://github.com/firebase/flutterfire/issues/17162)

hussenIbrahim:
Hello @chong-shao 
Many issues was reported 
[17165](https://github.com/firebase/flutterfire/issues/17165)
[17164](https://github.com/firebase/flutterfire/issues/17164)

AnthonyHaykal:
It's Happening with us as well! Just wanted to point out if we know at least what the problem could be?

AnthonyHaykal:
Thank you for letting me know!!

PR #2874, opened 2025-03-14
feat(fcm): Support `proxy` field in FCM `AndroidNotification`

jonathanedey:
Fixes: #2863

### DIFF
--- a/src/messaging/messaging-api.ts
+++ b/src/messaging/messaging-api.ts
@@ -608,6 +608,13 @@ export interface AndroidNotification {
    * displayed on the long-press menu each time a new notification arrives.
    */
   notificationCount?: number;
+
+  /**
+   * Sets if this notification should be proxied. Must be either `allow`,
+   * `deny` or `if_priority_lowered`. If unspecified, defaults to
+   * `if_priority_lowered`.
+   */
+  proxy?: ('allow' | 'deny' | 'if_priority_lowered');
 }
 
 /**
--- a/src/messaging/messaging-api.ts
+++ b/src/messaging/messaging-api.ts
@@ -610,8 +610,8 @@ export interface AndroidNotification {
   notificationCount?: number;
 
   /**
-   * Sets if this notification should be proxied. Must be either `allow`,
-   * `deny` or `if_priority_lowered`. If unspecified, defaults to
+   * Sets if this notification should attempt to be proxied. Must be either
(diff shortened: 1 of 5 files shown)


src/messaging/messaging-api.ts
```diff
@@ -608,6 +608,13 @@ export interface AndroidNotification {
    * displayed on the long-press menu each time a new notification arrives.
    */
   notificationCount?: number;
+
+  /**
+   * Sets if this notification should attempt to be proxied. Must be either
+   * `allow`, `deny` or `if_priority_lowered`. If unspecified, defaults to
+   * `if_priority_lowered`.
```
lahirumaramba, line 615: Let's double check this, because I think currently it is `undefined`

chong-shao (approved):

lahirumaramba (approved):
Thank you! LGTM!

src/messaging/messaging-api.ts
lahirumaramba: Resolved offline.

jonathanedey: @egilmorez, What are your thoughts on the wording of:

> If unspecified, defaults to `if_priority_lowered`

This could be read as the SDK defaults to `if_priority_lowered` where in reality we leave it undefined and leave the default mapping to the backend which eventually does the same. The consensus is that we are okay with this since we also use the same wording for the `AndroidNotification.visibility` but we could use your expertise on this.

jonathanedey: @OrlandriaH-G ^^

OrlandriaH-G: What do you think about: "If unspecified, it will remain undefined and no default value will be applied."

OrlandriaH-G, line 615: What do you think about: "If unspecified, it will remain undefined and no default value will be applied."

OrlandriaH-G (changes requested):

src/messaging/messaging-api.ts
OrlandriaH-G: Discussed offline and the new suggested text is: If unspecified, it remains undefined in the Admin SDK, and defers to the FCM backend's default mapping.

jonathanedey: Done!

OrlandriaH-G (approved):

PR #2890, opened 2025-04-01
Favor 'service' over 'serviceId' in FDC connectorConfig

inlined:
Baby first step towards reconciling client & server SDKs?


maneesht:
The issue is actually on the JS SDK side. I'll be sure to reconcile the differences there

Issue #2892, opened 2025-04-06
[FR] Messaging: Support for iOS 18.4 Declarative push messages


acrobid:
Declarative push messages are a new feature in iOS 18.4 that allow push messages to web apps without service worker handling. This greatly improves the deliverability of background notifications for web apps on iOS.

This requires the addition of an extra top level field in the Message interface as described [here](https://pr-preview.s3.amazonaws.com/w3c/push-api/pull/385.html#declarative-push-message).

Please allow the optional "web_push" (integer) field to the Message interface to enable Declarative push messages.

Currently, when adding this field, I receive the following validation error:

Error: Failed to send notification: Invalid JSON payload received. Unknown name "web_push" at 'message.webpush': Cannot find field.

More information on this feature [here](https://webkit.org/blog/16535/meet-declarative-web-push/#how-to-use-declarative-web-push).

acrobid:
@chong-shao 

@lahirumaramba 

Here is more background on why it is important, and how it doesn't break backwards compatibility:

https://developer.apple.com/videos/play/wwdc2025/235/?time=0

chong-shao:
Hello @acrobid ,  the API Admin SDK exposes needs to align with [FCM's REST API ](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages?_gl=1*ayt7y4*_up*MQ..*_ga*MTMwMjczMDQyNy4xNzUyMDk0OTIw*_ga_CW55HF8NVT*czE3NTIwOTQ5MTkkbzEkZzAkdDE3NTIwOTQ5MTkkajYwJGwwJGgw#webpushconfig) 

Could you file a feature request to FCM https://firebase.google.com/support/troubleshooter/report/features first, once that's accepted and implemented, we can start add the same feature in Admin SDK.

acrobid:
@chong-shao Thanks for the link. I just submitted a feature request using the form and provided a link to this issue, as well as some context. It is assigned as **Case 10366664: Support for iOS 18.4 Declarative push messages.**

For requests like this, from what I understand from @mbleigh, "we need to have an internal champion take it through a review process". @lahirumaramba  Is there anyone willing to ping that team to flag that request?

[...]

PR #2893, opened 2025-04-07
feat: add field to webpush options to allow for Declarative Web Push messages

acrobid:
See issue here:

https://github.com/firebase/firebase-admin-node/issues/2892

Spec proposal:

See https://pr-preview.s3.amazonaws.com/w3c/push-api/pull/385.html#declarative-push-message

### API Changes

This change allows for a new optional property in the webpush options, and will not affect existing projects.

### DIFF
--- a/src/messaging/messaging-api.ts
+++ b/src/messaging/messaging-api.ts
@@ -108,6 +108,13 @@ export interface WebpushConfig {
    * Options for features provided by the FCM SDK for Web.
    */
   fcmOptions?: WebpushFcmOptions;
+
+  /**
+   * Support for Declarative Web Push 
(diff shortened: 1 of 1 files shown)


lahirumaramba:
Hey @acrobid, thank you for your contribution we appreciate it! I don't think this feature is currently supported by the BE REST service so adding it to the SDK wouldn't address the issue.

Hey @chong-shao could you add more context here, please? Thank you!

chong-shao:
Yes, please refer to https://github.com/firebase/firebase-admin-node/issues/2892#issuecomment-3054019669 for context

acrobid:
My understand is that declarative web push has now landed in the specifications.

https://github.com/mdn/browser-compat-data/pull/28000

PR #2905, opened 2025-04-24
feat(fdc): Data Connect Bulk Import

lahirumaramba:
Add new API interfaces for `insert`, `insertMany`, `upsert`, `upsertMany` operations.

### DIFF
--- a/src/data-connect/data-connect-api-client-internal.ts
+++ b/src/data-connect/data-connect-api-client-internal.ts
@@ -204,7 +204,7 @@ export class DataConnectApiClient {
    * Handles nested objects, arrays, strings, numbers, and booleans.
    * Ensures strings are properly escaped.
    */
-  private objectToString(data: any): string {
+  private objectToString(data: unknown): string {
     if (typeof data === 'string') {
       const escapedString = data
         .replace(/\\/g, '\\\\') // Replace \ with \\
@@ -278,7 +278,7 @@ export class DataConnectApiClient {
     if (validator.isArray(data)) {
       throw new FirebaseDataConnectError(
         DATA_CONNECT_ERROR_CODE_MAPPING.INVALID_ARGUMENT,
-        '`data` must be an object, not an array, for single insert.');
+        '`data` must be an object, not an array, for single insert. For arrays, please use `insertMany` function.');
     }
     if (!validator.isNonNullObject(data)) {
       throw new FirebaseDataConnectError(
@@ -344,7 +344,7 @@ export class DataConnectApiClient {
     if (validator.isArray(data)) {
       throw new FirebaseDataConnectError(
         DATA_CONNECT_ERROR_CODE_MAPPING.INVALID_ARGUMENT,
-        '`data` must be an object, not an array, for single upsert.');
+        '`data` must be an object, not an array, for single upsert. For arrays, please use `upsertMany` function.');
     }
     if (!validator.isNonNullObject(data)) {
       throw new FirebaseDataConnectError(
--- a/src/data-connect/data-connect-api-client-internal.ts
+++ b/src/data-connect/data-connect-api-client-internal.ts
@@ -246,6 +246,14 @@ export class DataConnectApiClient {
     return String(data);
   }
 
+  private formatTableName(tableName: string): string {
+    // Format tableName: first character to lowercase
+    if (tableName && tableName.length > 0) {
+      return tableName.charAt(0).toLowerCase() + tableName.slice(1);
+    }
+    return tableName;
+  }
+
   /**
    * Insert a single row into the specified table.
    */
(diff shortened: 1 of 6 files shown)


src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -198,6 +198,203 @@ export class DataConnectApiClient {
     const message = error.message || `Unknown server error: ${response.text}`;
     return new FirebaseDataConnectError(code, message);
   }
+
+  /**
+   * Converts JSON data into a GraphQL literal string.
+   * Handles nested objects, arrays, strings, numbers, and booleans.
+   * Ensures strings are properly escaped.
+   */
+  private objectToString(data: any): string {
```
maneesht, line 207: Should this just be `unknown`?

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -198,6 +198,203 @@ export class DataConnectApiClient {
...
+      // Filter out properties where the value is undefined BEFORE mapping
+      const kvPairs = Object.entries(data)
+        .filter(([, val]) => val !== undefined)
+        .map(([key, val]) => {
+          // GraphQL object keys are typically unquoted.
+          return `${key}: ${this.objectToString(val)}`;
+        });
+  
+      if (kvPairs.length === 0) {
+        return '{}'; // Represent an object with no defined properties as {}
```
maneesht, line 231: Hmm, I'm not sure there's a case for when this would happen

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -198,6 +198,203 @@ export class DataConnectApiClient {
...
+  ): Promise<ExecuteGraphqlResponse<GraphQlResponse>> {
+    if (!validator.isNonEmptyString(tableName)) {
+      throw new FirebaseDataConnectError(
+        DATA_CONNECT_ERROR_CODE_MAPPING.INVALID_ARGUMENT,
+        '`tableName` must be a non-empty string.');
+    }
+    if (validator.isArray(data)) {
+      throw new FirebaseDataConnectError(
+        DATA_CONNECT_ERROR_CODE_MAPPING.INVALID_ARGUMENT,
+        '`data` must be an object, not an array, for single insert.');
```
maneesht, line 281: I wonder if we should help even more and say `for arrays, please use "insertMany"`

test/unit/data-connect/data-connect-api-client-internal.spec.ts
```diff
@@ -218,10 +224,382 @@ describe('DataConnectApiClient', () => {
...
+    director: {
+      name: undefined,
+      age: undefined
+    },
+    notes: undefined,
+    releaseYear: undefined,
+    extras: [1, undefined, 'hello', undefined, { a: 1, b: undefined }]
+  };
+
+  const tableNames = ['movie', 'Movie', 'MOVIE', 'toybox', 'toyBox', 'ToyBox', 'TOYBOX'];
```
maneesht, line 267: Can we also add a case for `mOvIE`?

lahirumaramba: Added `mOvIE` and `toyBOX`

src/data-connect/data-connect-api-client-internal.ts
lahirumaramba: Good call. Updated!

src/data-connect/data-connect-api-client-internal.ts
lahirumaramba: Good catch! Updated to `unknown`

src/data-connect/data-connect-api-client-internal.ts
lahirumaramba: I think this hits if you pass in an object with all `undefined` values
```
director: {
    name: undefined,
    age: undefined
}
```

markarndt (approved):

test/unit/data-connect/data-connect-api-client-internal.spec.ts
```diff
@@ -218,10 +224,382 @@ describe('DataConnectApiClient', () => {
...
+       ratings: null,
+       director: {},
+       extras: [1, null, "hello", null, { a: 1 }]
+      },
+      {
+       genre: "Action",
+       title: "Die Hard",
+       ratings: null,
+       director: {},
+       extras: [1, null, "hello", null, { a: 1 }]
```
maneesht, line 573: We shouldn't allow for sparsed arrays, can we catch this somehow?

lahirumaramba: Good catch! Made some changes to catch this and throw before we make the request to the server. Added new test cases to cover this change. Thanks!

lahirumaramba: Discussed offline and decided not to validate this on the SDK side. We will let the BE service handle this for now.

maneesht (approved):

PR #2909, opened 2025-05-02
fix: expose disableAuth property on Tenant

mnahkies:
adds the `disableAuth` property to the `Tenant` class, allowing users to distinguish between active and disabled tenants.

ref:
- https://cloud.google.com/identity-platform/docs/reference/rest/v2/projects.tenants#Tenant

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "firebase-admin",
-  "version": "13.2.0",
+  "version": "13.3.0",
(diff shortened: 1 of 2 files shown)


mnahkies:
@lahirumaramba is it possible to get a review of this? It's a fairly minor change, but its somewhat annoying not being able to filter out disabled tenants when scripting over multi-tenant auth projects (currently have to workaround by catching the exception when operations fail on the disabled ones).

PR #2934, opened 2025-06-15
feat: Add contentChanged in ApnsPayload 

woals4815:
### Purpose
Starting from iOS 26, it is now possible to reload a widget using the contentChanged value. To support this, we are adding the contentChanged value to the payload.
([iOS Widget documentation](https://developer.apple.com/documentation/widgetkit/updating-widgets-with-widgetkit-push-notifications#Send-the-WidgetKit-notification-payload-that-updates-your-widget))


woals4815:
@lahirumaramba is there any comment for this PR?

PR #2972, opened 2025-09-08
chore(fdc): update api version, integration tests, and CONTRIBUTING.md

stephenarosaj:
- Update API version to point from `v1alpha` to `v1` 
- Update Data Connect integration tests to make them only dependent on having an empty database (or a database which was previously empty, which the tests have been run against before)
- Update README to explain how to run Data Connect integration tests

### DIFF
--- a/test/integration/data-connect.spec.ts
+++ b/test/integration/data-connect.spec.ts
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2024 Google Inc.
+ * Copyright 2024 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2020 Google Inc.
+ * Copyright 2020 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/.github/actions/send-email/action.yml
+++ b/.github/actions/send-email/action.yml
@@ -1,4 +1,4 @@
-# Copyright 2021 Google Inc.
+# Copyright 2021 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/.github/actions/send-email/index.js
+++ b/.github/actions/send-email/index.js
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2021 Google Inc.
+ * Copyright 2021 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/.github/actions/send-tweet/action.yml
+++ b/.github/actions/send-tweet/action.yml
@@ -1,4 +1,4 @@
-# Copyright 2020 Google Inc.
+# Copyright 2020 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/.github/actions/send-tweet/index.js
+++ b/.github/actions/send-tweet/index.js
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2020 Google Inc.
+ * Copyright 2020 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/.github/scripts/publish_package.sh
+++ b/.github/scripts/publish_package.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright 2020 Google Inc.
+# Copyright 2020 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/.github/scripts/publish_preflight_check.sh
+++ b/.github/scripts/publish_preflight_check.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright 2020 Google Inc.
+# Copyright 2020 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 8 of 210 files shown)


CONTRIBUTING.md
```diff
@@ -245,6 +245,50 @@ to ensure that exported user records contain the password hashes of the user acc
...
+      4. Set your **Location** to `us-west2`
+      5. Set your **Cloud SQL instance ID** to `my-instance`
+      6. Set your **Database name** to `my-database`
+      7. Set your **Service ID** to `my-service`
+      8. Click **Submit**. This operation may take up to 10 minutes to complete - you may 
+         continue setting up while this completes.
+   2. Set up your Data Connect schema locally:
+      1. Run the following commands from the command line to setup your data connect app:
+         ```bash
+         $ mkdir fdc-integration-test
```
maneesht: Can we just check this folder in, actually?

test/integration/data-connect.spec.ts
```diff
@@ -17,186 +17,238 @@
...
+ *   id: String!
+ *   name: String!
+ *   address: String!
+ * }
+ */
 type User = {
-  uid?: string;
-  name?: string;
-  address?: string;
+  id: string;
```
maneesht, line 37: Any reason why this was updated to `uid`?

test/integration/data-connect.spec.ts
```diff
@@ -17,186 +17,238 @@
...
+    it('executeGraphql() successfully executes a GraphQL query', async () => {
+      const resp = await getDataConnect(connectorConfig)
+        .executeGraphql<ListUsersResponse, undefined>(queryListUsers);
       expect(resp.data.users).to.be.not.empty;
-      expect(resp.data.users.length).to.be.greaterThan(1);
-      expectedUsers.forEach((expectedUser) => {
-        expect(resp.data.users).to.deep.include(expectedUser);
+      expect(resp.data.users.length).to.greaterThan(1);
+      resp.data.users.forEach((user) => {
+        expect(expectedUserIds).to.include(user.id);
```
maneesht: Why don't we compare the full objects here? You can use `deep.equal` to compare the arrays I believe

test/integration/data-connect.spec.ts
```diff
@@ -258,74 +315,63 @@ describe('getDataConnect()', () => {
...
 
       it('executeGraphql() should throw for impersonated mutation with unauthenticated claims', async () => {
-        return getDataConnect(connectorConfig).executeGraphql(updateImpersonatedUser, optsUnauthorizedClaims)
+        return getDataConnect(connectorConfig).executeGraphql(updateFredrickUserImpersonated, optsUnauthorizedClaims)
           .should.eventually.be.rejected.and.has.property('code', 'data-connect/unauthenticated');
       });
 
       it('executeGraphql() should return null for an impersonated mutation with non-existing authenticated claims',
         async () => {
+          console.log('test')
```
maneesht: Remove console.log

test/integration/data-connect.spec.ts
```diff
@@ -258,74 +315,63 @@ describe('getDataConnect()', () => {
...
           const resp = await getDataConnect(connectorConfig).executeGraphql<UserUpdateResponse, undefined>(
-            updateImpersonatedUser, optsNonExistingClaims);
+            updateFredrickUserImpersonated, { ...optsNonExistingClaims });
```
maneesht: Why did this have to be spread?

test/integration/data-connect.spec.ts
stephenarosaj: there was a reason for it, i remember it was making my life difficult somehow... but that's lost on me now

CONTRIBUTING.md
stephenarosaj: Done.

test/integration/data-connect.spec.ts
stephenarosaj: Done.

test/integration/data-connect.spec.ts
stephenarosaj: We could check the full objects, but these objects are updated / edited by other test cases so I wanted to make the checks rely only on their IDs - the other tests which mutate the data check individual fields.

To make the tests even more air-tight, I could use different users / IDs for each test, rather than passing the same ID/object around to be used in different test scenarios. LMK WYT

test/integration/data-connect.spec.ts
stephenarosaj: doesn't have to - good catch. Done.

test/integration/dataconnect/dataconnect/.dataconnect/schema/main/implicit.gql
```diff
@@ -0,0 +1,6 @@
+extend type Email {
```
maneesht: Do we need these `.dataconnect` files?

test/integration/data-connect.spec.ts
maneesht: Ah, so you're worried that other tests may clobber the data, is that right?

test/integration/dataconnect/dataconnect/.dataconnect/schema/main/implicit.gql
stephenarosaj: Good catch - they're definitely not needed

test/integration/data-connect.spec.ts
stephenarosaj: Exactly. Right now, the order of operations seems to be the same every time, but since there's async stuff going on I still wanted to make it so I'm only checking data fields if I'm actually interested in what's there, and checking IDs if I'm only interested in the objects being present. 

BUT this actually made me go back and realize that despite being worried about order of operations / clobbering, future tests actually rely on previous tests having been completed successfully... just to be safe, I'm going to go back and explicitly set the state of the database in each test. And then at the end of the tests, I'll also add a "delete everything" mutation call to wipe things out

stephenarosaj: Just updated tests so they explicitly set the initial state of the database before each test, and clear the state after each test. This allowed me to update the check-expects so that each one is looking for full object equality rather than just IDs.

PR #2974, opened 2025-09-09
chore: update copyright headers from Google Inc. to Google LLC

lahirumaramba:


jonathanedey:
Do we need to rebuild these actions to get the headers here as well?
https://github.com/firebase/firebase-admin-node/blob/50047c595355aced9d6fefc43759e15b3e3409d5/.github/actions/send-email/dist/index.js#L28125

https://github.com/firebase/firebase-admin-node/blob/50047c595355aced9d6fefc43759e15b3e3409d5/.github/actions/send-tweet/dist/index.js#L1834

lahirumaramba:
Thanks! I think the build artifacts are probably fine for now

PR #2979, opened 2025-09-17
feat(dc): Add executeQuery and executeMutation APIs to Data Connect

stephenarosaj:
# API Changes
- Added `executeQuery()` and `executeMutation()` to `src/data-connect/data-connect.ts`. These allow users to call deployed operations with impersonated auth credentials.

# Testing
- New unit tests added which match the coverage of existing `executeGraphql*` APIs
- New integration tests added which more than match the coverage of existing `executeGraphql*` APIs
- **The tests pass locally, but a re-deploy of dataconnect to the CI/CD firebase project is required. Tagging @lahirumaramba**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
         "uuid": "^11.0.2"
       },
       "devDependencies": {
-        "@firebase/api-documenter": "^0.4.0",
+        "@firebase/api-documenter": "^0.5.0",
         "@firebase/app-compat": "^0.2.1",
         "@firebase/auth-compat": "^0.6.0",
         "@firebase/auth-types": "^0.13.0",
@@ -504,21 +504,21 @@
       }
     },
     "node_modules/@fastify/busboy": {
-      "version": "3.1.1",
-      "resolved": "https://registry.npmjs.org/@fastify/busboy/-/busboy-3.1.1.tgz",
-      "integrity": "sha512-5DGmA8FTdB2XbDeEwc/5ZXBl6UbBAyBOOLlPuBnZ/N1SwdH9Ii+cOX3tBROlDgcTXxjOYnLMVoKk9+FXAw0CJw==",
+      "version": "3.2.0",
+      "resolved": "https://registry.npmjs.org/@fastify/busboy/-/busboy-3.2.0.tgz",
+      "integrity": "sha512-m9FVDXU3GT2ITSe0UaMA5rU3QkfC/UXtCU8y0gSN/GugTqtVldOBWIB5V6V3sbmenVZUIpU6f+mPEO2+m5iTaA==",
       "license": "MIT"
     },
     "node_modules/@firebase/api-documenter": {
-      "version": "0.4.0",
-      "resolved": "https://registry.npmjs.org/@firebase/api-documenter/-/api-documenter-0.4.0.tgz",
-      "integrity": "sha512-UUPxxj1wIAGkXBCF9UL1dhfyzM4Lcd/gMSCoJqy8o75rLqQVGB2GSBaU7I4JmQsQhIIk/nKKLiy4HM/70kEfZA==",
+      "version": "0.5.0",
+      "resolved": "https://registry.npmjs.org/@firebase/api-documenter/-/api-documenter-0.5.0.tgz",
+      "integrity": "sha512-7fMC0sVInXsatjWMJG45W6ZQBtRcHh5D4oVpLXzKWBe/0TbRyBSE/ApecH9mn48XZGSs8/8Y0k7mKrLsXPLprA==",
       "dev": true,
       "license": "Apache-2.0",
       "dependencies": {
         "@microsoft/tsdoc": "0.12.24",
-        "@rushstack/node-core-library": "3.59.7",
-        "@rushstack/ts-command-line": "4.15.2",
+        "@rushstack/node-core-library": "3.66.1",
+        "@rushstack/ts-command-line": "4.23.3",
         "api-extractor-model-me": "0.1.1",
         "colors": "~1.4.0",
         "js-yaml": "4.1.0",
@@ -527,6 +527,9 @@
       },
       "bin": {
         "api-documenter-fire": "dist/start.js"
+      },
+      "engines": {
+        "node": ">=18.0.0"
       }
     },
     "node_modules/@firebase/app": {
@@ -1753,9 +1756,9 @@
       "optional": true
     },
     "node_modules/@rushstack/node-core-library": {
-      "version": "3.59.7",
-      "resolved": "https://registry.npmjs.org/@rushstack/node-core-library/-/node-core-library-3.59.7.tgz",
-      "integrity": "sha512-ln1Drq0h+Hwa1JVA65x5mlSgUrBa1uHL+V89FqVWQgXd1vVIMhrtqtWGQrhTnFHxru5ppX+FY39VWELF/FjQCw==",
+      "version": "3.66.1",
+      "resolved": "https://registry.npmjs.org/@rushstack/node-core-library/-/node-core-library-3.66.1.tgz",
+      "integrity": "sha512-ker69cVKAoar7MMtDFZC4CzcDxjwqIhFzqEnYI5NRN/8M3om6saWCVx/A7vL2t/jFCJsnzQplRDqA7c78pytng==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
@@ -1897,26 +1900,77 @@
       }
     },
     "node_modules/@rushstack/ts-command-line": {
-      "version": "4.15.2",
-      "resolved": "https://registry.npmjs.org/@rushstack/ts-command-line/-/ts-command-line-4.15.2.tgz",
-      "integrity": "sha512-5+C2uoJY8b+odcZD6coEe2XNC4ZjGB4vCMESbqW/8DHRWC/qIHfANdmN9F1wz/lAgxz72i7xRoVtPY2j7e4gpQ==",
+      "version": "4.23.3",
+      "resolved": "https://registry.npmjs.org/@rushstack/ts-command-line/-/ts-command-line-4.23.3.tgz",
+      "integrity": "sha512-HazKL8fv4HMQMzrKJCrOrhyBPPdzk7iajUXgsASwjQ8ROo1cmgyqxt/k9+SdmrNLGE1zATgRqMUH3s/6smbRMA==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
+        "@rushstack/terminal": "0.14.5",
         "@types/argparse": "1.0.38",
         "argparse": "~1.0.9",
-        "colors": "~1.2.1",
         "string-argv": "~0.3.1"
       }
     },
-    "node_modules/@rushstack/ts-command-line/node_modules/colors": {
-      "version": "1.2.5",
-      "resolved": "https://registry.npmjs.org/colors/-/colors-1.2.5.tgz",
-      "integrity": "sha512-erNRLao/Y3Fv54qUa0LBB+//Uf3YwMUmdJinN20yMXm9zdKKqH9wt7R9IIVZ+K7ShzfpLV/Zg8+VyrBJYB4lpg==",
+    "node_modules/@rushstack/ts-command-line/node_modules/@rushstack/node-core-library": {
+      "version": "5.10.2",
+      "resolved": "https://registry.npmjs.org/@rushstack/node-core-library/-/node-core-library-5.10.2.tgz",
+      "integrity": "sha512-xOF/2gVJZTfjTxbo4BDj9RtQq/HFnrrKdtem4JkyRLnwsRz2UDTg8gA1/et10fBx5RxmZD9bYVGST69W8ME5OQ==",
       "dev": true,
       "license": "MIT",
-      "engines": {
-        "node": ">=0.1.90"
+      "dependencies": {
+        "ajv": "~8.13.0",
+        "ajv-draft-04": "~1.0.0",
+        "ajv-formats": "~3.0.1",
+        "fs-extra": "~7.0.1",
+        "import-lazy": "~4.0.0",
+        "jju": "~1.4.0",
+        "resolve": "~1.22.1",
+        "semver": "~7.5.4"
+      },
+      "peerDependencies": {
+        "@types/node": "*"
+      },
+      "peerDependenciesMeta": {
+        "@types/node": {
+          "optional": true
+        }
+      }
+    },
+    "node_modules/@rushstack/ts-command-line/node_modules/@rushstack/terminal": {
+      "version": "0.14.5",
+      "resolved": "https://registry.npmjs.org/@rushstack/terminal/-/terminal-0.14.5.tgz",
+      "integrity": "sha512-TEOpNwwmsZVrkp0omnuTUTGZRJKTr6n6m4OITiNjkqzLAkcazVpwR1SOtBg6uzpkIBLgrcNHETqI8rbw3uiUfw==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@rushstack/node-core-library": "5.10.2",
+        "supports-color": "~8.1.1"
+      },
+      "peerDependencies": {
+        "@types/node": "*"
+      },
+      "peerDependenciesMeta": {
+        "@types/node": {
+          "optional": true
+        }
+      }
+    },
+    "node_modules/@rushstack/ts-command-line/node_modules/ajv": {
+      "version": "8.13.0",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-8.13.0.tgz",
+      "integrity": "sha512-PRA911Blj99jR5RMeTunVbNXMF6Lp4vZXnk5GQjcnUWUTsrXtekg/pnmFFI2u/I36Y/2bITGS30GZCXei6uNkA==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "fast-deep-equal": "^3.1.3",
+        "json-schema-traverse": "^1.0.0",
+        "require-from-string": "^2.0.2",
+        "uri-js": "^4.4.1"
+      },
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/epoberezkin"
       }
     },
     "node_modules/@sinonjs/commons": {
@@ -12001,9 +12055,9 @@
       }
     },
     "node_modules/validator": {
-      "version": "13.12.0",
-      "resolved": "https://registry.npmjs.org/validator/-/validator-13.12.0.tgz",
-      "integrity": "sha512-c1Q0mCiPlgdTVVVIJIrBuxNicYE+t/7oKeI9MWLj3fh/uq2Pxh/3eeWbVZ4OcGW1TUf53At0njHw5SMdA3tmMg==",
+      "version": "13.15.20",
+      "resolved": "https://registry.npmjs.org/validator/-/validator-13.15.20.tgz",
+      "integrity": "sha512-KxPOq3V2LmfQPP4eqf3Mq/zrT0Dqp2Vmx2Bn285LwVahLc+CsxOM0crBHczm8ijlcjZ0Q5Xd6LW3z3odTPnlrw==",
       "dev": true,
       "license": "MIT",
       "engines": {
--- a/package.json
+++ b/package.json
@@ -223,7 +223,7 @@
     "@google-cloud/storage": "^7.14.0"
   },
   "devDependencies": {
-    "@firebase/api-documenter": "^0.4.0",
+    "@firebase/api-documenter": "^0.5.0",
     "@firebase/app-compat": "^0.2.1",
     "@firebase/auth-compat": "^0.6.0",
     "@firebase/auth-types": "^0.13.0",


stephenarosaj:
## Note:
The integration tests have a huge diff because I moved the existing tests under a new parent `describe()`, and so their indentation level was increased. Existing tests were *not* changed in this PR.

stephenarosaj:
Forgot a few changes! Didn't know that you could close and re-open :P thanks Yuchen!

stephenarosaj:
wait - need to privatize the execute API, execution should come from operation refs (in a future PR)

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -27,17 +27,31 @@ import { ConnectorConfig, ExecuteGraphqlResponse, GraphqlOptions } from './data-
...
+// !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+// TODO: CHANGE THIS BACK TO PROD - AUTOPUSH IS ONLY USED FOR LOCAL TESTING BEFORE CHANGES PROPAGATE TO PROD
+// !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+/** The Firebase Data Connect backend service URL format. */
+const FIREBASE_DATA_CONNECT_SERVICES_URL_FORMAT =
+  'https://autopush-firebasedataconnect.sandbox.googleapis.com/{version}/projects/{projectId}/locations/{locationId}/services/{serviceId}:{endpointId}';
+
+/** The Firebase Data Connect backend connector URL format. */
+const FIREBASE_DATA_CONNECT_CONNECTORS_URL_FORMAT =
+  'https://autopush-firebasedataconnect.sandbox.googleapis.com/{version}/projects/{projectId}/locations/{locationId}/services/{serviceId}/connectors/{connectorId}:{endpointId}';
```
maneesht: Leaving a comment to block merging on these changes

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -138,28 +145,138 @@ export class DataConnectApiClient {
...
+    };
+    const url = await this.getUrl(
+      API_VERSION,
+      this.connectorConfig.location,
+      this.connectorConfig.serviceId,
+      endpoint,
+      this.connectorConfig.connector,
+    );
+    return this.makeGqlRequest<GraphqlResponse>(url, data)
+      .then((resp) => {
```
maneesht: Can we get rid of this line? And can we replace this block with a try/catch?

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -112,24 +135,8 @@ export class DataConnectApiClient {
...
-          throw new FirebaseDataConnectError(
-            DATA_CONNECT_ERROR_CODE_MAPPING.QUERY_ERROR, allMessages);
-        }
-        return Promise.resolve({
-          data: resp.data.data as GraphqlResponse,
-        });
-      })
+    const url = await this.getUrl(API_VERSION, this.connectorConfig.location, this.connectorConfig.serviceId, endpoint);
+    return this.makeGqlRequest<GraphqlResponse>(url, data)
       .then((resp) => {
```
maneesht: We can get rid of this

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -138,28 +145,138 @@ export class DataConnectApiClient {
...
+   * @param endpoint - The endpoint to call.
+   * @param options - The GraphQL options, including impersonation details.
+   * @returns A promise that fulfills with the GraphQL response.
+   */
+  private async executeOperationHelper<GraphqlResponse, Variables>(
+    endpoint: string,
+    options: GraphqlOptions<Variables>
+  ): Promise<ExecuteGraphqlResponse<GraphqlResponse>> {
+    if (
+      typeof options.operationName === 'undefined' ||
```
maneesht: Just as a note, this'll change when we move things like `name` and `options` out of the `GraphqlOptions` field

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -138,28 +145,138 @@ export class DataConnectApiClient {
...
+      version,
+      projectId,
+      locationId,
+      serviceId,
+      endpointId,
+      connectorId
+    };
+    let urlFormat: string;
+    if (useEmulator()) {
+      (urlParams as any).host = emulatorHost();
```
maneesht: Can we make this more strongly typed? I'd rather we add an empty host field in `urlParams` than cast to `any` and lose type-safety

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -138,28 +145,138 @@ export class DataConnectApiClient {
...
+      urlFormat = connectorId === undefined || connectorId === ''
```
maneesht: Can't this just be replaced by `!connectorId`?

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -138,28 +145,138 @@ export class DataConnectApiClient {
...
+        ? FIREBASE_DATA_CONNECT_EMULATOR_SERVICES_URL_FORMAT
+        : FIREBASE_DATA_CONNECT_EMULATOR_CONNECTORS_URL_FORMAT;
+    } else {
+      urlFormat = connectorId === undefined || connectorId === ''
```
maneesht: Same comment here, `!urlFormat`

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -138,28 +145,138 @@ export class DataConnectApiClient {
...
+        ? FIREBASE_DATA_CONNECT_SERVICES_URL_FORMAT
+        : FIREBASE_DATA_CONNECT_CONNECTORS_URL_FORMAT;
+    }
+    if (connectorId) {
+      (urlParams as any).connectorId = connectorId;
```
maneesht: Shouldn't be `any`

test/integration/dataconnect/dataconnect/my-connector/queries.gql
```diff
@@ -34,17 +34,18 @@ query ListUsersNoAccess @auth(level: NO_ACCESS) {
 		address
 	}
 }
-query ListUsersImpersonationAnon @auth(level: USER_ANON) {
+query ListUsersImpersonationAnon @auth(level: USER_ANON, insecureReason: "test") {
```
maneesht: Is the `insecureReason` needed?

test/integration/dataconnect/dataconnect/my-connector/mutations.gql
```diff
@@ -1,7 +1,7 @@
 mutation upsertFredUser @auth(level: NO_ACCESS) {
 	user_upsert(data: { id: "fred_id", address: "32 Elm St.", name: "Fred" })
 }
-mutation updateFredrickUserImpersonation @auth(level: USER) {
+mutation updateFredrickUserImpersonation @auth(level: USER, insecureReason: "test") {
```
maneesht: Note: These changes likely require a re-deploy on the CI project

test/unit/data-connect/data-connect-api-client-internal.spec.ts
```diff
@@ -230,6 +234,340 @@ describe('DataConnectApiClient', () => {
...
     });
   });
+
+  describe('executeQuery', () => {
+    const unauthenticatedOptions: GraphqlOptions<unknown> = 
+      { operationName: 'unauthenticatedQuery', impersonate: { unauthenticated: true } };
+    const authenticatedOptions: GraphqlOptions<unknown> = 
+      { operationName: 'authenticatedQuery', impersonate: { authClaims: { sub: 'authenticated-UUID' } } };
+
+    it('should reject when no operationName is provided', () => {
```
maneesht: Since these are all rejections, should they be nested in a separate `reject` describe, like on line 308?

test/unit/data-connect/data-connect-api-client-internal.spec.ts
```diff
@@ -230,6 +234,340 @@ describe('DataConnectApiClient', () => {
...
+            name: string;
+            address: string;
+          }
+        ];
+      }
+      it('for an unauthenticated request', () => {
+        const stub = sandbox
+          .stub(HttpClient.prototype, 'send')
+          .resolves(utils.responseFrom(TEST_RESPONSE, 200));
+        return apiClient.executeQuery<UsersResponse, unknown>(unauthenticatedOptions).then((resp) => {
```
maneesht: We can probably await this

test/unit/data-connect/data-connect-api-client-internal.spec.ts
```diff
@@ -230,6 +234,340 @@ describe('DataConnectApiClient', () => {
...
+              operationName: authenticatedOptions.operationName, 
+              extensions: { impersonate: authenticatedOptions.impersonate }
+            }
+          });
+        });
+      });
+    });
+
+    it('should use DATA_CONNECT_EMULATOR_HOST if set', () => {
+      process.env.DATA_CONNECT_EMULATOR_HOST = 'localhost:9399';
```
maneesht, line 369: Can we reset this after the test is complete?

test/unit/data-connect/data-connect-api-client-internal.spec.ts
```diff
@@ -230,6 +234,340 @@ describe('DataConnectApiClient', () => {
...
+            data: { 
+              operationName: unauthenticatedOptions.operationName, 
+              extensions: { impersonate: unauthenticatedOptions.impersonate }
+            }
+          });
+        });
+    });
+
+    it('should use DATA_CONNECT_EMULATOR_HOST if set', () => {
+      process.env.DATA_CONNECT_EMULATOR_HOST = 'localhost:9399';
```
maneesht, line 513: Reset after?

src/data-connect/data-connect-api-client-internal.ts
stephenarosaj: Resolved in previous change. Done.

test/integration/dataconnect/dataconnect/my-connector/mutations.gql
stephenarosaj: Yes, they deff do - I'll leave this comment open to make sure it gets re-deployed! Tagging @lahirumaramba

src/data-connect/data-connect-api-client-internal.ts
stephenarosaj: Done. Same as above.

src/data-connect/data-connect-api-client-internal.ts
stephenarosaj: Done. Let me know if this is what you were thinking.

test/unit/data-connect/data-connect-api-client-internal.spec.ts
stephenarosaj: Done. Also made this change to the existing executeGraphql* tests.

test/unit/data-connect/data-connect-api-client-internal.spec.ts
stephenarosaj: Done. Made similar changes to the existing unit tests.

src/data-connect/data-connect-api-client-internal.ts
stephenarosaj: Done. I've split `getUrl()` into the `getServicesUrl()` and `getConnectorsUrl()` methods. Type safety comes from the typed arguments to each method and typed interfaces used inside of these methods to build the URL param objects.

src/data-connect/data-connect-api-client-internal.ts
stephenarosaj: Taken care of now that `getUrl()` is split into `getConnectorsUrl()` and `getServicesUrl()`. Done.

src/data-connect/data-connect-api-client-internal.ts
stephenarosaj: Taken care of now that `getUrl()` is split into `getConnectorsUrl()` and `getServicesUrl()`. Done.

src/data-connect/data-connect-api-client-internal.ts
stephenarosaj: Taken care of now that `getUrl()` is split into `getConnectorsUrl()` and `getServicesUrl()`. Done.

test/integration/dataconnect/dataconnect/my-connector/queries.gql
stephenarosaj: Good catch - this one shouldn't need an `insecureReason`. Done.

test/unit/data-connect/data-connect-api-client-internal.spec.ts
stephenarosaj: Should be taken care of in the `afterEach` call around line 100. Done.

test/unit/data-connect/data-connect-api-client-internal.spec.ts
stephenarosaj: Should be taken care of in the `afterEach` call around line 100. Done.

src/data-connect/data-connect-api-client-internal.ts
stephenarosaj: Changes have pushed to prod, this code has been updated. Done.

stephenarosaj: Resolving because the changes have pushed to prod and the code has been updated accordingly. Done.

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "firebase-admin",
-  "version": "13.4.0",
+  "version": "13.5.0",
```
lahirumaramba, line 3: The version should only be bumped in a release PR. You can revert this for now and we will change it when we are ready to cut the next release

src/data-connect/data-connect.ts
```diff
@@ -157,4 +161,153 @@ export class DataConnect {
   ): Promise<ExecuteGraphqlResponse<GraphQlResponse>> {
     return this.client.upsertMany(tableName, variables);
   }
+
+  /** @internal */
+  public executeQuery<Data, Variables>(
```
lahirumaramba: If we are not exporting these as part of the public API, would it make sense to remove`executeQuery` and  `executeMutation` from`src/data-connect/data-connect.ts`?

etc/firebase-admin.data-connect.api.md
```diff
@@ -60,4 +65,60 @@ export interface ImpersonateUnauthenticated {
...
+    // (undocumented)
+    readonly name: string;
+    // (undocumented)
+    readonly options?: RefOptions | undefined;
+    // (undocumented)
+    readonly variables: Variables;
+}
+
+// @public (undocumented)
+export interface OperationResult<Data, Variables> {
```
lahirumaramba: Could you check why some of these new types are missing documentation, please?

package-lock.json
stephenarosaj: Reverted. Done.

src/data-connect/data-connect.ts
stephenarosaj: Discussed offline. It **_would_** make more sense, but the `OperationRef` classes needed to interface with the `DataConnectApiClient` class, which is internal and should not be exported. This class is a private field in the `DataConnect` class, so the `OperationRef` class would call this @internal `executeQuery()` method to access the private field.

However, after further design review, this design was abandoned in favor of an explicit public `executeQuery()` method on the `DataConnect` class - so this awkwardness is no longer an issue :)

etc/firebase-admin.data-connect.api.md
stephenarosaj: Removed after re-design. Done.

etc/firebase-admin.data-connect.api.md
```diff
@@ -60,4 +70,16 @@ export interface ImpersonateUnauthenticated {
     unauthenticated: true;
 }
 
+// @public
+export interface OperationOptions {
+    impersonate?: ImpersonateAuthenticated | ImpersonateUnauthenticated;
+}
+
+// @public
+export function validateAdminArgs<Variables extends object>(connectorConfig: ConnectorConfig, dcOrVarsOrOptions?: DataConnect | Variables | OperationOptions, varsOrOptions?: Variables | OperationOptions, options?: OperationOptions, hasVars?: boolean, validateVars?: boolean): {
```
maneesht: @lahirumaramba Any opinions on this? We only need this for the generated SDK, and we don't intend for developers to use this.

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -23,21 +23,32 @@ import {
...
-    'https://firebasedataconnect.googleapis.com/{version}/projects/{projectId}/locations/{locationId}/services/{serviceId}:{endpointId}';
+/** The Firebase Data Connect backend service URL format. */
+const FIREBASE_DATA_CONNECT_SERVICES_URL_FORMAT =
+  'https://firebasedataconnect.googleapis.com/{version}/projects/{projectId}/locations/{locationId}/services/{serviceId}:{endpointId}';
 
-/** Firebase Data Connect base URl format when using the Data Connect emultor. */
-const FIREBASE_DATA_CONNECT_EMULATOR_BASE_URL_FORMAT =
+/** The Firebase Data Connect backend connector URL format. */
+const FIREBASE_DATA_CONNECT_CONNECTORS_URL_FORMAT =
+  'https://firebasedataconnect.googleapis.com/{version}/projects/{projectId}/locations/{locationId}/services/{serviceId}/connectors/{connectorId}:{endpointId}';
```
maneesht: Nit: Maybe we should extract out the production URL, just in case we mis-spell it somewhere

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -129,52 +170,163 @@ export class DataConnectApiClient {
...
+
+    if (this.connectorConfig.connector === undefined || this.connectorConfig.connector === '') {
+      throw new FirebaseDataConnectError(
+        DATA_CONNECT_ERROR_CODE_MAPPING.INVALID_ARGUMENT,
+        `The 'connectorConfig.connector' field used to instantiate your Data Connect
+        instance must be a non-empty string (the connectorId) when calling executeQuery or executeMutation.`);
+    }
+
+    const data = {
+      ...(variables && { variables: variables }),
```
maneesht, line 259: nit: I don't think we need to check the `variables` nullity here.

src/data-connect/validate-admin-args.ts
```diff
@@ -0,0 +1,80 @@
...
+
+import { getDataConnect } from './index';
+import { DataConnect } from './data-connect';
+import { ConnectorConfig, OperationOptions } from './data-connect-api';
+import {
+  DATA_CONNECT_ERROR_CODE_MAPPING,
+  FirebaseDataConnectError,
+} from './data-connect-api-client-internal';
+
+/**
```
maneesht, line 26: Nit: Can we add example usage here?

test/integration/data-connect.spec.ts
```diff
@@ -204,223 +226,836 @@ describe('getDataConnect()', () => {
...
+        expect(jeffResponse.data.user_upsert.id).equals(jeffUser.id);
+
+        const emailResponse = await getDataConnect(connectorConfig).executeGraphql<EmailUpsertResponse, unknown>(
+          upsertFredEmail
+        );
+        //{ data: { email_upsert: { id: 'email_id' } } }
+        expect(emailResponse.data.email_upsert.id).to.not.be.empty;
+
+        const deleteResponse = await getDataConnect(connectorConfig).executeGraphql<DeleteResponse, unknown>(deleteAll);
+        expect(deleteResponse.data.email_deleteMany).to.be.greaterThan(0);
```
maneesht: Can this assertion be more exact?

test/integration/data-connect.spec.ts
```diff
@@ -204,223 +226,836 @@ describe('getDataConnect()', () => {
...
+        );
+        //{ data: { user_insert: { id: 'jeff_id' } } }
+        expect(jeffResponse.data.user_upsert.id).to.not.be.empty;
+        expect(jeffResponse.data.user_upsert.id).equals(jeffUser.id);
+
+        const emailResponse = await getDataConnect(connectorConfig).executeGraphql<EmailUpsertResponse, unknown>(
+          upsertFredEmail
+        );
+        //{ data: { email_upsert: { id: 'email_id' } } }
+        expect(emailResponse.data.email_upsert.id).to.not.be.empty;
```
maneesht: Can we also assert that the email matched what the upserted one was?

test/integration/data-connect.spec.ts
```diff
@@ -204,223 +226,836 @@ describe('getDataConnect()', () => {
...
-      expect(resp.data.emails).to.not.be.empty;
-      expect(resp.data.emails).to.deep.equal(initialState.emails);
-    });
+      it('executeGraphql() successfully executes a GraphQL query', async () => {
+        const resp = await getDataConnect(connectorConfig)
+          .executeGraphql<ListUsersResponse, unknown>(queryListUsers);
+        expect(resp.data.users).to.not.be.empty;
+        expect(resp.data.users.length).to.equal(initialState.users.length);
+        resp.data.users.forEach((user) => {
+          expect(initialState.users).to.deep.include(user);
```
maneesht: Nit: Can't this just be `expect(resp.data.emails).to.deep.equal(initialState.emails);`? It's cleaner

test/integration/data-connect.spec.ts
```diff
@@ -204,223 +226,836 @@ describe('getDataConnect()', () => {
...
+    describe('queryRef()', () => {
+      it("should fail when executing a query which doesn't exist", async () => {
+        return getDataConnect(connectorConfig).executeQuery(
+          'DOES_NOT_EXIST!!!',
+          undefined,
+          optsUnauthorizedClaims
+        ).should.eventually.be.rejected.and.have.property('code', 'data-connect/not-found');
+      })
+
+      it('should execut a query with variables', async () => {
```
maneesht: Nit: Spelling
```suggestion
      it('should execute a query with variables', async () => {
```

test/integration/data-connect.spec.ts
```diff
@@ -204,223 +226,836 @@ describe('getDataConnect()', () => {
...
+  });
 
-      it('executeGraphql() should return an empty list for an impersonated query with non-existing authenticated ' + 
-        'claims',
-      async () => {
-        const resp = await getDataConnect(connectorConfig).executeGraphql<ListUsersResponse, undefined>(
-          queryListUsersImpersonation, optsNonExistingClaims);
-        // Should find no data
-        expect(resp.data.users).to.be.empty;
+  describe('operation ref API', () => {
```
maneesht:
```suggestion
  describe('executeQuery|Mutation API', () => {
```

test/integration/data-connect.spec.ts
```diff
@@ -204,223 +226,836 @@ describe('getDataConnect()', () => {
...
+      describe('with unauthenticated impersonation', () => {
+        it('should successfully execute a query with @auth(level: PUBLIC)', async () => {
+          const resp = await getDataConnect(connectorConfig).executeQuery<ListUsersResponse, undefined>(
+            'ListUsersPublic',
+            undefined,
+            optsUnauthorizedClaims
+          );
+          expect(resp.data.users).to.not.be.empty;
+          expect(resp.data.users.length).to.equal(initialState.users.length);
+          resp.data.users.forEach((user) => {
```
maneesht:
```suggestion
          expect(resp.data.users).to.deep.equal(initialState.users);
```

test/integration/data-connect.spec.ts
```diff
@@ -204,223 +226,836 @@ describe('getDataConnect()', () => {
...
-        return getDataConnect(connectorConfig).executeGraphql(updateFredrickUserImpersonated, optsUnauthorizedClaims)
-          .should.eventually.be.rejected.and.has.property('code', 'data-connect/unauthenticated');
+      describe('with authenticated anonymous impersonation', () => {
+        it('should successfully execute a query with @auth(level: PUBLIC)', async () => {
+          const resp = await getDataConnect(connectorConfig).executeQuery<ListUsersResponse, undefined>(
+            'ListUsersPublic', undefined, optsAuthorizedFredAnonClaims
+          );
+          expect(resp.data.users).to.not.be.empty;
+          expect(resp.data.users.length).to.equal(initialState.users.length);
+          resp.data.users.forEach((user) => {
```
maneesht:
```suggestion
          expect(resp.data.users).to.deep.eq(initialState.users);
```

maneesht: Please apply the changes below as well

test/integration/data-connect.spec.ts
```diff
@@ -204,223 +226,836 @@ describe('getDataConnect()', () => {
...
+          'DOES_NOT_EXIST!!!', optsUnauthorizedClaims
+        ).should.eventually.be.rejected.and.have.property('code', 'data-connect/not-found');
+      })
+
+      describe('with unauthenticated impersonation', () => {
+        it('should successfully execute a mutation with @auth(level: PUBLIC)', async () => {
+          const resp = await getDataConnect(connectorConfig)
+            .executeMutation<InsertEmailResponse, InsertEmailVariables>(
+              'InsertEmailPublic',
+              { id: `email_id_${Math.random() * 1000}` },
```
maneesht: Nit: I would move this to a util function

test/integration/data-connect.spec.ts
```diff
@@ -204,223 +226,836 @@ describe('getDataConnect()', () => {
...
-      it('executeGraphql() should throw for an impersonated query with authenticated claims', async () => {
-        return await getDataConnect(connectorConfig).executeGraphql(queryListEmails, optsAuthorizedFredClaims)
-          .should.eventually.be.rejected.and.has.property('code', 'data-connect/permission-denied');
+    describe('mutationRef()', () => {
+      it("should fail when executing a mutation which doesn't exist", async () => {
+        return getDataConnect(connectorConfig).executeMutation(
+          'DOES_NOT_EXIST!!!', optsUnauthorizedClaims
+        ).should.eventually.be.rejected.and.have.property('code', 'data-connect/not-found');
+      })
+
```
maneesht, line 713: Looks like there's a missing test here for mutation with variables that we have for queries

test/integration/data-connect.spec.ts
```diff
@@ -204,223 +226,836 @@ describe('getDataConnect()', () => {
...
+        it("should fail to execute a mutation using the impersonated user's auth.uid", async () => {
+          return getDataConnect(connectorConfig).executeMutation<InsertEmailResponse, InsertEmailVariables>(
+            'InsertEmailImpersonation',
+            { id: `email_id_${Math.random() * 1000}` },
+          ).should.eventually.be.rejected.and.have.property('code', 'data-connect/query-error');
+        });
+      });
     });
   });
-});
```
maneesht, line 426: Newline please

src/data-connect/validate-admin-args.ts
```diff
@@ -0,0 +1,80 @@
...
+ * arguments passed in.
+ * @param connectorConfig - DataConnect connector config
+ * @param dcOrVarsOrOptions - the first argument provided to a generated admin function
+ * @param varsOrOptions - the second argument provided to a generated admin function
+ * @param options - the third argument provided to a generated admin function
+ * @param hasVars - boolean parameter indicating whether the operation has variables
+ * @param validateVars - boolean parameter indicating whether we should expect to find a value for realVars
+ * @returns parsed DataConnect, Variables, and Options for the operation
+ */
+export function validateAdminArgs<Variables extends object>(
```
maneesht: Can you write some tests for this?

src/data-connect/validate-admin-args.ts
```diff
@@ -0,0 +1,80 @@
...
+  varsOrOptions?: Variables | OperationOptions,
+  options?: OperationOptions,
+  hasVars?: boolean,
+  validateVars?: boolean
+): { dc: DataConnect; vars: Variables; options: OperationOptions; } {
+  let dcInstance: DataConnect;
+  let realVars: Variables;
+  let realOptions: OperationOptions;
+
+  if (dcOrVarsOrOptions && 'connectorConfig' in dcOrVarsOrOptions) {
```
maneesht, line 64: I wonder if we can just use `instanceof` to check whether the `dcOrVarsOrOptions` is a `DataConnect` object?

maneesht (changes requested):

src/data-connect/data-connect-api-client-internal.ts
stephenarosaj: I don't think so either, but currently, our `executeGraphqlHelper()` [implementation](https://github.com/firebase/firebase-admin-node/blob/8c9895e87485f5aa686f809b80fd0fb854d25035/src/data-connect/data-connect-api-client-internal.ts#L128C7-L128C67) uses this same pattern to avoid sending a `variables: undefined` field if variables are undefined. Those unit tests and the new unit tests expect this behavior, I'd prefer to not change it unless you feel there's a reason to include `variables: undefined`.

src/data-connect/validate-admin-args.ts
stephenarosaj: Added some examples, and a better explanation. Done.

Also, this implementation is ripped 1:1 from the previous work we did for the Generated React SDK's [`validateReactArgs()`](https://github.com/invertase/tanstack-query-firebase/blob/547d5b17032d565715fe293bc75e3f3d8cc854c2/packages/react/src/data-connect/validateReactArgs.ts#L32)

test/integration/data-connect.spec.ts
stephenarosaj: Done.

test/integration/data-connect.spec.ts
stephenarosaj: Done.

test/integration/data-connect.spec.ts
stephenarosaj: Done.

test/integration/data-connect.spec.ts
stephenarosaj: Done.

src/data-connect/validate-admin-args.ts
stephenarosaj: I think we can - DataConnect is a class, so it should be able to tell. But for some reason we opted not to do this for the Client SDK, nor the React SDK, where both DataConnect is also a class. 

There may have been some reason for this - I'm not sure. I think if we change the pattern here, we should do the same in the other SDKs. I'll look into it further after the admin work is done. Resolving for now.

test/integration/data-connect.spec.ts
stephenarosaj: Good catch - done.

src/data-connect/data-connect-api-client-internal.ts
stephenarosaj: Done.

src/data-connect/validate-admin-args.ts
stephenarosaj: Done.

test/unit/data-connect/validate-admin-args.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+    DATA_CONNECT_ERROR_CODE_MAPPING.INVALID_ARGUMENT,
+    'Variables required.'
+  );
+
+  const stubDcInstance = { connectorConfig: connectorConfig, source: 'STUB' } as unknown as DataConnect;
+  beforeEach(() => {
+    getDataConnectStub = sinon.stub(DataConnectService.prototype, 'getDataConnect').returns(stubDcInstance);
+
+    // initializing app required, "using" it is required for successful build/compile 
+    app = initializeApp();
```
maneesht: You might be able to get away with just calling `initializeApp` without setting the app to something

stephenarosaj: Was able to remove the `initializeApp` call entirely. Done.

test/unit/data-connect/validate-admin-args.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+      const { dc: dcInstance, vars: inputVars, options: inputOpts } = validateAdminArgs(
+        connectorConfig, providedDcInstance
+      );
+      expect(dcInstance).to.deep.equal(providedDcInstance);
+      expect(inputVars).to.be.undefined;
+      expect(inputOpts).to.be.undefined;
+    });
+
+    it('should successfully parse the provided options', () => {
+      const { dc: dcInstance, vars: inputVars, options: inputOpts } = validateAdminArgs<object>(
```
maneesht: Any reason why the `object` is needed?

test/unit/data-connect/validate-admin-args.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+        it('and the first argument is variables instance', () => {
+          expect(() => {
+            validateAdminArgs(connectorConfig, variables, undefined, undefined, true, false);
+          }).to.not.throw(invalidVariablesError);
+          
+          const { dc: dcInstance, vars: inputVars, options: inputOpts } = validateAdminArgs(
+            connectorConfig, variables, undefined, undefined, true, false
+          );
+          expect(dcInstance).to.deep.equal(stubDcInstance);
+          expect(inputVars).to.deep.equal(inputVars);
```
maneesht: Typo?
```suggestion
          expect(inputVars).to.deep.equal(variables);
```

test/unit/index.spec.ts
```diff
@@ -121,3 +121,4 @@ import './extensions/extensions-api-client-internal.spec';
 // Data Connect
 import './data-connect/index.spec';
 import './data-connect/data-connect-api-client-internal.spec';
+// import './data-connect/validate-admin-args.spec';
```
maneesht: Remove comment

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "firebase-admin",
-  "version": "13.5.0",
+  "version": "13.4.0",
```
lahirumaramba: You can revert this. We update the package.json only during a release PR (which we will handle separately from this PR)

src/data-connect/data-connect.ts
```diff
@@ -165,4 +167,76 @@ export class DataConnect {
...
+   * Optionally, you may provide auth impersonation.
+   *
+   * @param name - The name of the defined query to execute.
+   * @param options - The GraphQL options, must include operationName and impersonation details.
+   * @returns A promise that fulfills with the GraphQL response.
+   */
+  public executeQuery<Data>(
+    name: string,
+    options?: OperationOptions
+  ): Promise<ExecuteOperationResponse<Data>>;
```
lahirumaramba, line 184: Does this handle the method overload correctly? If I call `executeQuery('myQuery', { impersonate: ... })` would it take `options` as `variables`?

lahirumaramba: If that breaks the method overload we need to perform a runtime check to differentiate them.

```
public executeQuery<Data, Variables>(
  name: string,
  variablesOrOptions?: Variables | OperationOptions,
  options?: OperationOptions
): Promise<ExecuteOperationResponse<Data>> {
```

test/unit/index.spec.ts
lahirumaramba: We should uncomment this if we want to run the tests in `validate-admin-args.spec.ts`

etc/firebase-admin.data-connect.api.md
lahirumaramba: Good point, is there a way to not export this from the public API surface?

src/data-connect/index.ts
```diff
@@ -29,14 +29,19 @@ import { ConnectorConfig } from './data-connect-api';
...
   ImpersonateUnauthenticated,
-  AuthClaims
+  AuthClaims,
+  OperationOptions,
 } from './data-connect-api'
 export {
   DataConnect,
 } from './data-connect'
+export {
+  validateAdminArgs
```
lahirumaramba: If this function is only for internal usage let's not export it to the public API surface here

src/data-connect/data-connect-api.ts
```diff
@@ -43,7 +48,17 @@ export interface ExecuteGraphqlResponse<GraphqlResponse> {
 }
 
 /**
- * Interface representing GraphQL options.
+ * Interface representing ExecuteOperation response.
+ */
+export interface ExecuteOperationResponse<GraphqlResponse> {
```
lahirumaramba, line 53: Is this interface identical to `ExecuteGraphqlResponse`? If that is the case and if we don't see it changing in the future, it might be better to create a type alias.

`export type ExecuteOperationResponse<T> = ExecuteGraphqlResponse<T>;`

test/unit/data-connect/validate-admin-args.spec.ts
stephenarosaj: No, good catch - Done.

test/unit/data-connect/validate-admin-args.spec.ts
stephenarosaj: Good catch - Done.

package.json
stephenarosaj: Oops sorry! it keeps re-updating it as part of the install/build/apidocs script, good catch

test/unit/index.spec.ts
stephenarosaj: Done.

src/data-connect/data-connect.ts
stephenarosaj: @lahirumaramba i added some unit tests in [`test/unit/data-connect/data-connect.spec.ts`](https://github.com/firebase/firebase-admin-node/pull/2979/files#diff-2bbde60d5ed35190810009e83cbce341efc7ad930b25be895515a78962808a7c) - your suspicion is correct, without type parameters (for required or fully optional variables), the second parameter is always treated as variables. however, i believe this is WAI, and i'd like to know how you feel about keeping this behavior?

there are no `options` in our Client SDK, so we don't run into this issue there, but we do have `options` in our React / Angular SDKs, and they do have this behavior (see our [React SDK's README description](https://source.corp.google.com/piper///depot/google3/third_party/firebase/dataconnect/sdkgen/jsgen/react/README.md.go.tmpl;rcl=771239260;l=127)). our solution was to clearly document this and to always use type parameters in our generated code, because we didn't want to try and identify "these fields mean this must be an options object, and this one must be a variables object", since users can use any name for variable fields in their GQL files

also, while the IDE won't pick up on the error, the backend / emulators will reject requests which omit required variables or provide unrecognized variables for that specific operation. here's what shows up when testing out this scenario:
```
{
  errorInfo: {
    code: 'data-connect/query-error',
    message: '$id (String) is missing $impersonate is not expected (did you mean id?)'
  },
  codePrefix: 'data-connect'
}
```

lmk your thoughts :P

src/data-connect/data-connect-api.ts
stephenarosaj: Great callout - currently they're the same, but during a separate review it was brought up that they may indeed differ in the future, so it was defined on it's own to decouple the types. Resolving.

test/unit/index.spec.ts
stephenarosaj: Good catch. Done.

etc/firebase-admin.data-connect.api.md
stephenarosaj: I encountered some errors in simply marking it as `@internal` in [`validate-admin-args.ts`](https://github.com/firebase/firebase-admin-node/blob/e1d5c08389d09b5978eed6ba8e783de735b026a6/src/data-connect/validate-admin-args.ts#L27), but i was able to re-export it as an `@internal` const in [`index.ts`](https://github.com/firebase/firebase-admin-node/blob/e1d5c08389d09b5978eed6ba8e783de735b026a6/src/data-connect/index.ts#L91) and that worked. 

LMK what you think of this approach

src/data-connect/index.ts
stephenarosaj: Done. See above comment

package.json
lahirumaramba: The current version is `13.5.0` that's why this file still shows up in the PR :)

src/data-connect/data-connect.ts
lahirumaramba: Thanks for the additional context. I think it could create a confusion to the users if we break the method overload here. Does it make sense to try an approach similar to what we did in Firestore? https://github.com/firebase/firebase-admin-node/blob/8c9895e87485f5aa686f809b80fd0fb854d25035/src/firestore/index.ts#L154

src/data-connect/data-connect.ts
```diff
@@ -165,4 +167,76 @@ export class DataConnect {
...
+
+  public executeMutation<Data, Variables>(
+    name: string,
+    variables: Variables, 
+    options?: OperationOptions
+  ): Promise<ExecuteOperationResponse<Data>> {
+    return this.client.executeMutation<Data, Variables>(name, variables, options);
+  }
+}
```
lahirumaramba: nit: new line

src/data-connect/index.ts
```diff
@@ -80,3 +84,8 @@ export function getDataConnect(connectorConfig: ConnectorConfig, app?: App): Dat
   const dataConnectService = firebaseApp.getOrInitService('dataConnect', (app) => new DataConnectService(app));
   return dataConnectService.getDataConnect(connectorConfig);
 }
+
+/**
+ * @internal
+ */
+export const validateAdminArgs = _validateAdminArgs;
```
lahirumaramba: nit: new line

src/data-connect/validate-admin-args.ts
```diff
@@ -0,0 +1,91 @@
+/*!
+ * @license
+ * Copyright 2024 Google LLC
```
lahirumaramba: 2025

test/unit/data-connect/data-connect.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+            it('WITHOUT options provided', async () => {
+              await dc.executeMutation(operationName);
+              expect(executeMutationStub.calledOnceWithExactly(operationName, undefined, undefined)).to.be.true;
+            });
+          });
+        });
+      });
+    });
+  });
+});
```
lahirumaramba: nit: new line

test/unit/data-connect/validate-admin-args.spec.ts
```diff
@@ -0,0 +1,271 @@
+/*!
+ * @license
+ * Copyright 2024 Google LLC
```
lahirumaramba: 2025

lahirumaramba:
```suggestion
 * Copyright 2025 Google LLC
```

test/unit/index.spec.ts
```diff
@@ -121,3 +121,5 @@ import './extensions/extensions-api-client-internal.spec';
 // Data Connect
 import './data-connect/index.spec';
 import './data-connect/data-connect-api-client-internal.spec';
+import './data-connect/data-connect.spec';
+import './data-connect/validate-admin-args.spec';
```
lahirumaramba: nit: new line

etc/firebase-admin.data-connect.api.md
```diff
@@ -60,4 +70,11 @@ export interface ImpersonateUnauthenticated {
     unauthenticated: true;
 }
 
+// @public
+export interface OperationOptions {
+    impersonate?: ImpersonateAuthenticated | ImpersonateUnauthenticated;
+}
+
+// (No @packageDocumentation comment for this package)
```
lahirumaramba: Do you know where this is coming from?

lahirumaramba (changes requested):

test/integration/data-connect.spec.ts
stephenarosaj: Done.

test/integration/data-connect.spec.ts
stephenarosaj: Done.

test/integration/data-connect.spec.ts
stephenarosaj: Done.

test/integration/data-connect.spec.ts
stephenarosaj: Done.

test/integration/data-connect.spec.ts
stephenarosaj: Done.

package.json
stephenarosaj: Ah I see - thanks!

src/data-connect/data-connect.ts
stephenarosaj: Done.

src/data-connect/index.ts
stephenarosaj: Done.

src/data-connect/validate-admin-args.ts
stephenarosaj: Done.

test/unit/data-connect/data-connect.spec.ts
stephenarosaj: Done.

test/unit/data-connect/validate-admin-args.spec.ts
stephenarosaj: Done.

stephenarosaj: Done.

test/unit/index.spec.ts
stephenarosaj: Done.

etc/firebase-admin.data-connect.api.md
stephenarosaj: I believe it was caused by the licensing comment being the first comment rather than the `@packageDocumentation` comment ([source](https://api-extractor.com/pages/tsdoc/tag_packagedocumentation/#:~:text=it%20should%20be%20the%20first%20/**%20comment%20encountered%20in%20that%20file.): "it should be the first /** comment encountered in that file") - fixed.

src/data-connect/data-connect.ts
stephenarosaj: Discussed via go/fdc:admin-core-variadic-args - we've decided to allow this behavior to align with Client SDK

kevinthecheung (approved):

src/data-connect/data-connect.ts
```diff
@@ -165,4 +167,76 @@ export class DataConnect {
   ): Promise<ExecuteGraphqlResponse<GraphQlResponse>> {
     return this.client.upsertMany(tableName, variables);
   }
+
+  /**
+   * Executes a GraphQL query. The query must be defined in your Data Connect GraphQL files.
+   * Optionally, you may provide auth impersonation.
```
kevinthecheung:
```suggestion
   * Optionally, you can provide auth impersonation details. If you don't
   * specify a value for this option, the query will run with admin privileges
   * and will ignore all auth directives.
```

Here and throughout, rephrasing as needed

stephenarosaj: Done. Thanks!

PR #2985, opened 2025-10-01
chore(dc): Implement gen tracking

maneesht:
Allows DataConnect to track whether users are using the generated or core SDKs.

From the generated SDKs, we just need to run:
```ts
dataConnect.useGen(true);
```


src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -38,9 +38,16 @@ const FIREBASE_DATA_CONNECT_EMULATOR_BASE_URL_FORMAT =
...
-const DATA_CONNECT_CONFIG_HEADERS = {
-  'X-Firebase-Client': `fire-admin-node/${utils.getSdkVersion()}`
-};
+function getHeaders(isUsingGen: boolean): { [key: string]: string } {
+  const headerValue = {
+    'X-Firebase-Client': `fire-admin-node/${utils.getSdkVersion()}`,
+    'X-Goog-Api-Client': utils.getMetricsHeader(),
+  };
+  if (isUsingGen) {
+    headerValue['X-Goog-Api-Client'] += ' js/gen';
```
stephenarosaj: should this be something like `nodeadmin/gen` instead of `js/gen` - wdyt ?

stephenarosaj: also, what do you think of adding `js/` or `nodeadmin/` or whatever in the case that our core SDK is used without gen? this is what we do in the client SDK, but then again we already have so many requests that have come in with nothing added onto them...

maneesht: Updated

lahirumaramba (approved):
LGTM! Thanks!

PR #2994, opened 2025-10-23
feat(fcm): Add bandwidthConstrainedOk and restrictedSatelliteOk

lahirumaramba:
Adds two new optional boolean properties, `bandwidthConstrainedOk` and `restrictedSatelliteOk`, to the `AndroidConfig` type in the messaging API.

This change includes:
- Updating the `AndroidConfig` interface in `src/messaging/messaging-api.ts`.
- Adding the corresponding property mappings in `src/messaging/messaging-internal.ts`.
- Updating unit tests in `test/unit/messaging/messaging.spec.ts` to cover the new properties.

### DIFF
--- a/src/messaging/messaging-api.ts
+++ b/src/messaging/messaging-api.ts
@@ -430,14 +430,14 @@ export interface AndroidConfig {
   directBootOk?: boolean;
 
   /**
-   * A boolean indicating whether the message is allowed to be delivered in
-   * a low-bandwidth environment.
+   * A boolean indicating whether messages will be allowed to be delivered to  
+   * the app while the device is on a bandwidth constrained network.
    */
   bandwidthConstrainedOk?: boolean;
 
   /**
-   * A boolean indicating whether the message is allowed to be delivered
-   * via satellite.
+   * A boolean indicating whether messages will be allowed to be delivered to  
+   * the app while the device is on a restricted satellite network.
    */
   restrictedSatelliteOk?: boolean;
 }


src/messaging/messaging-api.ts
```diff
@@ -428,6 +428,18 @@ export interface AndroidConfig {
...
+   * A boolean indicating whether the message is allowed to be delivered in
+   * a low-bandwidth environment.
+   */
+  bandwidthConstrainedOk?: boolean;
+
+  /**
+   * A boolean indicating whether the message is allowed to be delivered
+   * via satellite.
+   */
+  restrictedSatelliteOk?: boolean;
```
jonathanedey, line 442: It would be nice to have the docs here match the other SDKs as well. But I'll hold off on changing it until it's TW approved in one of the other SDKs.

```node
   /**
   * A boolean indicating whether messages will be allowed to be delivered to
   * the app while the device is on a bandwidth constrained network.
   */
  bandwidthConstrainedOk?: boolean;

  /**
   * A boolean indicating whether messages will be allowed to be delivered to
   * the app while the device is on a restricted satellite network.
   */
  restrictedSatelliteOk?: boolean;
```

OrlandriaH-G: This looks good to me! I believe I approved this in one of the other SDKs as well

lahirumaramba: Thanks! Fixed.

PR #3009, opened 2025-11-07
fix: some validateAdminArgs unit tests wouldn't actually fail under incorrect behavior 

stephenarosaj:
some of the `validateAdminArgs()` unit tests used a pattern of nested check-expects, and the outer expect would just be checking that a specific error wasn't thrown, which would swallow up any failures from inner expects - it would always pass the outer expect, meaning that even if the inner expects should fail the test, the test would pass.

this is shown on lines [224](https://github.com/firebase/firebase-admin-node/blob/2e2b36a84ba28679bcb7aecdacabfec0bded2d48/test/unit/data-connect/validate-admin-args.spec.ts#L224) of `validate-admin-args.spec.ts`, where variables are provided to `validateAdminArgs()` but on line [227](https://github.com/firebase/firebase-admin-node/blob/2e2b36a84ba28679bcb7aecdacabfec0bded2d48/test/unit/data-connect/validate-admin-args.spec.ts#L227) we check that they are `undefined`. if you bring this incorrect inner expect out of the outer expect, the test fails - because we are expecting `undefined` when in reality we passed in variables and are correctly returned variables. to be clear, this is an incorrectly written test, but the behavior of the function is correct - the bad test wasn't caught because of this nested check-expects pattern.

changing the testing pattern here to flatten the expects to the same level solves the problem!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -708,7 +708,6 @@
       "version": "0.7.0",
       "resolved": "https://registry.npmjs.org/@firebase/component/-/component-0.7.0.tgz",
       "integrity": "sha512-wR9En2A+WESUHexjmRHkqtaVH94WLNKt6rmeqZhSLBybg4Wyf0Umk04SZsS6sBq4102ZsDBFwoqMqJYj2IoDSg==",
-      "dev": true,
       "dependencies": {
         "@firebase/util": "1.13.0",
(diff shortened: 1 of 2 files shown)


lahirumaramba (approved):
Thanks!

PR #3016, opened 2025-11-16
refactor: remove uuid dependency and use native crypto.randomUUID

LeonardoTonizza:
## Description

This PR removes the external `uuid` dependency and replaces it with Node.js native `crypto.randomUUID()` for UUID generation.

## Motivation

Since PR #2756 updated the minimum Node.js version requirement to v18, and `crypto.randomUUID()` has been available since Node.js v14.17.0, there is no longer a need for the external `uuid` package dependency.

## Changes

- Removed `uuid` package from `package.json`
- Replaced `import { v4 as uuid } from 'uuid'` with `import { randomUUID } from 'crypto'`
- Updated all `uuidv4()` calls to `randomUUID()`
- Updated tests to use the native implementation

## Testing

- [x] Existing unit tests pass

## References

- Node.js crypto.randomUUID() documentation: https://nodejs.org/api/crypto.html#cryptorandomuuidoptions
- PR #2756

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,8 +18,7 @@
         "google-auth-library": "^9.14.2",
         "jsonwebtoken": "^9.0.0",
         "jwks-rsa": "^3.1.0",
-        "node-forge": "^1.3.1",
-        "uuid": "^11.0.2"
+        "node-forge": "^1.3.1"
       },
       "devDependencies": {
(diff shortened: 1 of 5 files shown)


lahirumaramba:
Thank you for your contribution. This should be addressed in #3130 and #3051

PR #3027, opened 2025-11-28
feat(appcheck): Add support for minting limited-use tokens and custom JTI

lahirumaramba:
This PR adds support for the `limitedUse` and `jti` parameters in the `createToken()` API of the App Check module. This allows users to generate tokens with replay protection, suitable for one-time use or session-based contexts.

RELEASE_NOTE: Added support for creating limited use App Check tokens with custom `jti` in the `createToken()` API.

Internal: go/firebase-admin-node-limited-use-fac-tokens

### DIFF
--- a/src/app-check/app-check-api-client-internal.ts
+++ b/src/app-check/app-check-api-client-internal.ts
@@ -58,7 +58,11 @@ export class AppCheckApiClient {
    * @param appId - The mobile App ID.
    * @returns A promise that fulfills with a `AppCheckToken`.
    */
-  public exchangeToken(customToken: string, appId: string): Promise<AppCheckToken> {
+  public exchangeToken(
+    customToken: string,
+    appId: string,
+    limitedUse?: boolean
+  ): Promise<AppCheckToken> {
     if (!validator.isNonEmptyString(appId)) {
       throw new FirebaseAppCheckError(
         'invalid-argument',
@@ -75,7 +79,10 @@ export class AppCheckApiClient {
           method: 'POST',
           url,
           headers: FIREBASE_APP_CHECK_CONFIG_HEADERS,
-          data: { customToken }
+          data: {
+            customToken,
+            limitedUse,
+          }
         };
         return this.httpClient.send(request);
       })
--- a/src/app-check/app-check-api-client-internal.ts
+++ b/src/app-check/app-check-api-client-internal.ts
@@ -73,6 +73,11 @@ export class AppCheckApiClient {
         'invalid-argument',
         '`customToken` must be a non-empty string.');
     }
+    if (typeof limitedUse !== 'undefined' && !validator.isBoolean(limitedUse)) {
+      throw new FirebaseAppCheckError(
+        'invalid-argument',
+        '`limitedUse` must be a boolean value.');
+    }
     return this.getUrl(appId)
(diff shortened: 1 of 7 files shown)


jonathanedey (approved):
LGTM, Thanks!

test/unit/app-check/app-check-api-client-internal.spec.ts
```diff
@@ -210,11 +210,80 @@ describe('AppCheckApiClient', () => {
...
+          expect(resp.token).to.deep.equal(TEST_RESPONSE.token);
+          expect(resp.ttlMillis).to.deep.equal(3000);
+          expect(stub).to.have.been.calledOnce.and.calledWith({
+            method: 'POST',
+            url: `https://firebaseappcheck.googleapis.com/v1/projects/test-project/apps/${APP_ID}:exchangeCustomToken`,
+            headers: EXPECTED_HEADERS,
+            data: {
+              customToken: TEST_TOKEN_TO_EXCHANGE,
+              limitedUse: true,
+              jti: undefined,
```
weixifan: Just to confirm, will the App Check backend see a JSON object with this `jti: undefined` field being omitted in this case? Or will the backend see a literal string `undefined` being the value of the `jti` field?

src/app-check/app-check-api-client-internal.ts
```diff
@@ -69,13 +73,34 @@ export class AppCheckApiClient {
...
+        'invalid-argument',
+        '`limitedUse` must be a boolean value.');
+    }
+    if (typeof options?.jti !== 'undefined') {
+      if (!validator.isString(options.jti)) {
+        throw new FirebaseAppCheckError(
+          'invalid-argument',
+          '`jti` must be a string value.');
+      }
+      if (options.jti !== '' && !options.limitedUse) {
```
weixifan: I'm not familiar with JavaScript, but would it be possible to add a test case for leaving `jti` as undefined, while setting `limitedUse` to `false`? Ideally this should be an allowed combination, but I wonder if the `!== ''` would classify this as rejected. Similarly, it should also be allowed to leave both `jti` and `limitedUse` be undefined as an explicitly supplied empty `options` tuple `{}`.

weixifan (approved):
Thank you for this PR Lahiru!!

test/unit/app-check/app-check-api-client-internal.spec.ts
lahirumaramba: Good question, I think normally the REST endpoint should consider `undefined` JSON values as empty and omit. We could also omit them here on the SDK side if we want.

src/app-check/app-check-api-client-internal.ts
lahirumaramba: Good point. The current implementation should handle this. Added tests to cover this case. This could be an issue if the `jti` is empty string but we also have a check for that. :)

test/unit/app-check/app-check-api-client-internal.spec.ts
lahirumaramba: Updated the logic to omit undefined values

kevinthecheung (approved):

lahirumaramba:
@gemini-code-assist review

PR #3030, opened 2025-12-09
feat(auth): Add support for FPNV

boikoa-gl:

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/etc/firebase-admin.api.md
+++ b/etc/firebase-admin.api.md
@@ -27,6 +27,8 @@ export namespace app {
         // (undocumented)
         firestore(): firestore.Firestore;
         // (undocumented)
+        fpnv(): fpnv.Fpnv;
+        // (undocumented)
         installations(): installations.Installations;
         // @deprecated (undocumented)
         instanceId(): instanceId.InstanceId;
@@ -289,6 +291,17 @@ export namespace firestore {
     import UnionToIntersection = _firestore.UnionToIntersection;
     import ReadOnlyTransactionOptions = _firestore.ReadOnlyTransactionOptions;
     import setLogFunction = _firestore.setLogFunction;
+}
+
+// @public
+export function fpnv(app?: App): fpnv.Fpnv;
+
+// @public (undocumented)
+export namespace fpnv {
+    // Warning: (ae-forgotten-export) The symbol "Fpnv" needs to be exported by the entry point default-namespace.d.ts
+    export type Fpnv = Fpnv;
+    // Warning: (ae-forgotten-export) The symbol "FpnvToken" needs to be exported by the entry point default-namespace.d.ts
+    export type FpnvToken = FpnvToken;
 }
 
 // @public
--- a/src/app/firebase-namespace.ts
+++ b/src/app/firebase-namespace.ts
@@ -19,7 +19,7 @@ import { App as AppCore } from './core';
 import { AppStore, defaultAppStore } from './lifecycle';
 import {
   app, appCheck, auth, messaging, machineLearning, storage, firestore, database,
-  instanceId, installations, projectManagement, securityRules , remoteConfig, AppOptions,
+  instanceId, installations, projectManagement, securityRules , remoteConfig, fpnv, AppOptions,
 } from '../firebase-namespace-api';
 import { cert, refreshToken, applicationDefault } from './credential-factory';
 import { getSdkVersion } from '../utils/index';
@@ -37,6 +37,7 @@ import ProjectManagement = projectManagement.ProjectManagement;
 import RemoteConfig = remoteConfig.RemoteConfig;
 import SecurityRules = securityRules.SecurityRules;
 import Storage = storage.Storage;
+import Fpnv = fpnv.Fpnv;
 
 export interface FirebaseServiceNamespace <T> {
   (app?: App): T;
@@ -279,6 +280,18 @@ export class FirebaseNamespace {
     return Object.assign(fn, { AppCheck: appCheck });
   }
 
+  /**
+   * Gets the `Fpnv` service namespace. The returned namespace can be used to get the
+   * `Fpnv` service for the default app or an explicitly specified app.
+   */
+  get fpnv(): FirebaseServiceNamespace<Fpnv> {
+    const fn: FirebaseServiceNamespace<Fpnv> = (app?: App) => {
+      return this.ensureApp(app).fpnv();
+    };
+    const fpnv = require('../fpnv/fpnv').Fpnv;
+    return Object.assign(fn, { Fpnv: fpnv });
+  }
+
   // TODO: Change the return types to app.App in the following methods.
 
   /**
@@ -394,6 +407,11 @@ function extendApp(app: AppCore): App {
 
   result.remoteConfig = () => {
     const fn = require('../remote-config/index').getRemoteConfig;
+    return fn(app);
+  };
+
+  result.fpnv = () => {
+    const fn = require('../fpnv/index').getFirebasePnv;
     return fn(app);
   };
 
--- a/src/firebase-namespace-api.ts
+++ b/src/firebase-namespace-api.ts
@@ -26,6 +26,7 @@ import { projectManagement } from './project-management/project-management-names
 import { remoteConfig } from './remote-config/remote-config-namespace';
 import { securityRules } from './security-rules/security-rules-namespace';
 import { storage } from './storage/storage-namespace';
+import { fpnv } from './fpnv/fpnv-namespace';
 
 import { App as AppCore, AppOptions } from './app/index';
 
@@ -56,6 +57,7 @@ export namespace app {
     remoteConfig(): remoteConfig.RemoteConfig;
     securityRules(): securityRules.SecurityRules;
     storage(): storage.Storage;
+    fpnv(): fpnv.Fpnv;
 
     /**
      * Renders this local `FirebaseApp` unusable and frees the resources of
@@ -91,6 +93,7 @@ export { projectManagement } from './project-management/project-management-names
 export { remoteConfig } from './remote-config/remote-config-namespace';
 export { securityRules } from './security-rules/security-rules-namespace';
 export { storage } from './storage/storage-namespace';
+export { fpnv } from './fpnv/fpnv-namespace';
 
 // Declare other top-level members of the admin namespace below. Unfortunately, there's no
 // compile-time mechanism to ensure that the FirebaseNamespace class actually provides these
--- a/src/fpnv/fpnv-namespace.ts
+++ b/src/fpnv/fpnv-namespace.ts
@@ -1,0 +1,63 @@
+/*!
+ * @license
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { App } from '../app';
+import {
+  FpnvToken as TFpnvToken,
+} from './fpnv-api';
+import { Fpnv as TFpnv } from './fpnv';
+
+/**
+ * Gets the {@link firebase-admin.fpnv#Fpnv} service for the default app or a given app.
+ *
+ * `admin.fpnv()` can be called with no arguments to access the default
+ * app's `Fpnv` service or as `admin.fpnv(app)` to access the
+ * `Fpnv` service associated with a specific app.
+ *
+ * @example
+ * ```javascript
+ * // Get the `Fpnv` service for the default app
+ * var defaultFpnv = admin.fpnv();
+ * ```
+ *
+ * @example
+ * ```javascript
+ * // Get the `Fpnv` service for a given app
+ * var otherFpnv = admin.fpnv(otherApp);
+ * ```
+ *
+ * @param app - Optional app for which to return the `Fpnv` service.
+ *   If not provided, the default `Fpnv` service is returned.
+ *
+ * @returns The default `Fpnv` service if no
+ *   app is provided, or the `Fpnv` service associated with the provided
+ *   app.
+ */
+export declare function fpnv(app?: App): fpnv.Fpnv;
+
+// eslint-disable-next-line @typescript-eslint/no-namespace
+export namespace fpnv {
+  /**
+   * Type alias to {@link firebase-admin.fpnv#fpnv}.
+   */
+  export type Fpnv = TFpnv;
+
+  /**
+   * Type alias to {@link firebase-admin.fpnv#FpnvToken}.
+   */
+  export type FpnvToken = TFpnvToken
+}


src/fpnv/base-fpnv.ts
```diff
@@ -0,0 +1,23 @@
+import { App } from '../app';
```
lahirumaramba: Please add the license headers

src/fpnv/fpnv-api.ts
```diff
@@ -0,0 +1,39 @@
+/*!
```
lahirumaramba, line 1: Thanks for adding this file, but the namespaces are now deprecated. We don't support namespaces for new features. Please remove the namespaces support from this PR. (You can check `src/data-connect` for an example of a namespace-less feature).

src/fpnv/fpnv.ts
```diff
@@ -0,0 +1,41 @@
+/*!
+ * @license
+ * Copyright 2017 Google LLC
```
lahirumaramba:
```suggestion
 * Copyright 2025 Google LLC
```

src/fpnv/base-fpnv.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import {
+  FirebasePhoneNumberTokenVerifier,
+  FpnvToken,
+  createFPNTVerifier,
+} from './token-verifier';
+
+
+export abstract class BaseFpnv {
```
lahirumaramba: You can probably move this class to `src/fpnv/fpnv.ts`

src/fpnv/index.ts
```diff
@@ -0,0 +1,74 @@
+/*!
+ * Copyright 2020 Google LLC
```
lahirumaramba:
```suggestion
 * Copyright 2025 Google LLC
```

src/fpnv/index.ts
```diff
@@ -0,0 +1,74 @@
...
+ * ```javascript
+ * // Get the Fpnv service for the default app
+ * const defaultFpnv = getFirebasePnv();
+ * ```
+ *
+ * @example
+ * ```javascript
+ * // Get the Fpnv service for a given app
+ * const otherFpnv = getFirebasePnv(otherApp);
+ * ```
```
lahirumaramba, line 54: Let's get a tech writer review on the new docstrings

src/fpnv/token-verifier.ts
```diff
@@ -0,0 +1,246 @@
+import { App } from '../app';
+import { FpnvErrorCode, ErrorInfo, FirebasePnvError } from '../utils/error';
+import * as util from '../utils/index';
+import * as validator from '../utils/validator';
+import {
+  DecodedToken, decodeJwt, JwtError, JwtErrorCode,
+  PublicKeySignatureVerifier, ALGORITHM_ES256, SignatureVerifier,
+} from '../utils/jwt';
+
+export interface FpnvToken {
```
lahirumaramba: Please extract the public API surface (including any public types) to a new file named `fpnv-api.ts`. See `src/data-connect/data-connect-api.ts` for example. Add documentation for each new type.

src/fpnv/token-verifier.ts
```diff
@@ -0,0 +1,246 @@
+import { App } from '../app';
```
lahirumaramba, line 18: missing license header

lahirumaramba (changes requested):

src/utils/error.ts
```diff
@@ -1145,3 +1145,35 @@ const TOPIC_MGT_SERVER_TO_CLIENT_CODE: ServerToClientCode = {
...
+}
+
+export class FpnvErrorCode {
+  static INVALID_TOKEN = 'invalid_token';
+  static EXPIRED_TOKEN = 'expired_token';
+  static COMMON_ISSUE = {
+    code: "code",
+    message: "message"
+  };
+  // TODO: need to update codes properly
```
lahirumaramba: I understand that this is still WIP, as a note: please try not to commit any placeholder code. You can add them when you are ready and when necessary (instead of adding a placeholder for later).

src/fpnv/base-fpnv.ts
```diff
@@ -0,0 +1,23 @@
...
+  protected readonly fPNTVerifier: FirebasePhoneNumberTokenVerifier;
```
sam-gc: `fpnvVerifier`

src/fpnv/fpnv.ts
```diff
@@ -0,0 +1,41 @@
...
+  constructor(app: App) {
+    super(app);
+
+    this.app_ = app;
+  }
+
+  /**
+     * Returns the app associated with this Fpnv instance.
+     *
+     * @returns The app associated with this Fpnv instance.
```
sam-gc: Formatting (indentation) on the TSdoc

src/fpnv/index.ts
sam-gc: Should we do that as a separate pass later?

src/fpnv/fpnv.ts
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+import { App } from '../app';
+import { BaseFpnv } from './base-fpnv';
+
+/**
+ * Fpnv service bound to the provided app.
+ */
+export class Fpnv extends BaseFpnv {
```
sam-gc: I'm not sure there's much value in splitting `Fpnv` from `BaseFpnv` ... it's a common Java idiom but I'm not sure it's that useful, especially in this case. You can't use FPNV without an app so I think it's fine to make just a single `Fpnv` class that requires an app

src/fpnv/token-verifier.ts
```diff
@@ -0,0 +1,246 @@
...
+export const PN_TOKEN_INFO: FirebasePhoneNumberTokenInfo = {
+  url: 'https://firebase.google.com/docs/phone-number-verification',
+  verifyApiName: 'verifyToken()',
+  jwtName: 'Firebase Phone Verification token',
+  shortName: 'FPNV token',
+  typ: 'JWT',
+  expiredErrorCode: FpnvErrorCode.COMMON_ISSUE,
+};
+
+export interface FirebasePhoneNumberTokenInfo {
```
sam-gc: I'm not sure this token info class is actually necessary? I'm guessing this is related to something that Auth does, which manages multiple types of tokens. For FPNV we're just managing a single type of token. I think we should probably just hard-code some of the values as constants and call it a day, rather than constructing an object to pass in.

src/fpnv/token-verifier.ts
```diff
@@ -0,0 +1,246 @@
...
+    return decodedIdToken;
+  }
+
+  private async ensureProjectId(): Promise<string> {
+    const projectId = await util.findProjectId(this.app);
+    if (!validator.isNonEmptyString(projectId)) {
+      throw new FirebasePnvError(
+        FpnvErrorCode.COMMON_ISSUE,
+        'Must initialize app with a cert credential or set your Firebase project ID as the ' +
+                `GOOGLE_CLOUD_PROJECT environment variable to call ${this.tokenInfo.verifyApiName}.`);
```
sam-gc: For example, this should probably just be "... environment variable to call verifyToken()"

We can worry about abstracting this later *if* we support different types of tokens (which is unlikely at this point)

src/fpnv/token-verifier.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    let errorMessage: string | undefined;
+
+    // JWT Header
+    if (typeof header.kid === 'undefined') {
+      errorMessage = `${this.tokenInfo.jwtName} has no "kid" claim.`;
+      errorMessage += verifyJwtTokenDocsMessage;
+    } else if (header.alg !== ALGORITHM_ES256) {
+      errorMessage = `${this.tokenInfo.jwtName} has incorrect algorithm. Expected "` + ALGORITHM_ES256 + '" but got ' +
+                '"' + header.alg + '".' + verifyJwtTokenDocsMessage;
```
sam-gc: Let's use template strings consistently for these messages (you may still need to use `+` in order to split onto multiple lines)

src/fpnv/token-verifier.ts
boikoa-gl: it will 
```
export const PN_TOKEN_INFO: FirebasePhoneNumberTokenInfo = {
  url: 'https://firebase.google.com/docs/phone-number-verification',
  verifyApiName: 'verifyToken()',<-------------------------> this.tokenInfo.verifyApiName
  jwtName: 'Firebase Phone Verification token',
  shortName: 'FPNV token',
  typ: 'JWT',
  expiredErrorCode: FpnvErrorCode.EXPIRED_TOKEN,
};
```

src/fpnv/token-verifier.ts
boikoa-gl: Sure, we can remove or change this at all

src/fpnv/fpnv.ts
```diff
@@ -0,0 +1,51 @@
...
+} from './token-verifier';
+
+/**
+ * Fpnv service bound to the provided app.
+ */
+export class Fpnv  {
+  private readonly app_: App;
+
+  protected readonly fpnvVerifier: FirebasePhoneNumberTokenVerifier;
+
```
lahirumaramba, line 29:
```suggestion
  /**
 * @param app - The app for this `Fpnv` service.
 * @constructor
 * @internal
 */
```

src/fpnv/fpnv.ts
```diff
@@ -0,0 +1,51 @@
...
+import {
+  FirebasePhoneNumberTokenVerifier,
+  createFPNTVerifier,
+} from './token-verifier';
+
+/**
+ * Fpnv service bound to the provided app.
+ */
+export class Fpnv  {
+  private readonly app_: App;
```
lahirumaramba: nit: you can simplify this by using constructor parameter properties.

```js
constructor(readonly app: App) { ...

// this.app
```

src/fpnv/fpnv.ts
```diff
@@ -0,0 +1,51 @@
...
+  constructor(app: App) {
+
+    this.app_ = app;
+    this.fpnvVerifier = createFPNTVerifier(app);
+  }
+
+  /**
+   * Returns the app associated with this Auth instance.
+   *
+   * @returns The app associated with this Auth instance.
```
lahirumaramba: Did you mean Fpnv instance here?

src/fpnv/fpnv.ts
```diff
@@ -0,0 +1,51 @@
...
+   */
+  get app(): App {
```
lahirumaramba, line 51: Is this used for anything? If not, we can probably remove this getter

src/fpnv/fpnv.ts
```diff
@@ -0,0 +1,51 @@
...
+    return this.app_;
+  }
+
+  public async verifyToken(idToken: string): Promise<FpnvToken> {
```
lahirumaramba: Is this really an `idToken`?

src/fpnv/index.ts
lahirumaramba: Sounds good!

src/fpnv/token-verifier.ts
```diff
@@ -0,0 +1,249 @@
...
+import * as util from '../utils/index';
+import * as validator from '../utils/validator';
+import {
+  DecodedToken, decodeJwt, JwtError, JwtErrorCode,
+  PublicKeySignatureVerifier, ALGORITHM_ES256, SignatureVerifier,
+} from '../utils/jwt';
+
+const CLIENT_CERT_URL = 'https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com';
+
+
```
lahirumaramba: nit: extra new line

src/fpnv/token-verifier.ts
```diff
@@ -0,0 +1,249 @@
...
+                    `the entire string JWT which represents ${this.shortNameArticle} ` +
+                    `${this.tokenInfo.shortName}.` + verifyJwtTokenDocsMessage;
+        throw new FirebasePnvError(FpnvErrorCode.INVALID_ARGUMENT,
+          errorMessage);
+      }
+      throw new FirebasePnvError(FpnvErrorCode.INVALID_ARGUMENT, err.message);
+    }
+  }
+
+
```
lahirumaramba: nit: extra new line

src/fpnv/token-verifier.ts
```diff
@@ -0,0 +1,249 @@
...
+      errorMessage += verifyJwtTokenDocsMessage;
+    } else if (header.alg !== ALGORITHM_ES256) {
+      errorMessage = `${this.tokenInfo.jwtName} has incorrect algorithm. Expected ` + 
+      `"${ALGORITHM_ES256}" but got "${header.alg}". ${verifyJwtTokenDocsMessage}`;
+    } else if (header.typ !== this.tokenInfo.typ) {
+      errorMessage = `${this.tokenInfo.jwtName} has incorrect typ. Expected "${this.tokenInfo.typ}" but got ` +
+                '"' + header.typ + '".' + verifyJwtTokenDocsMessage;
+    }
+    // FPNV Token
+    else if (!((Array.isArray(payload.aud) ? payload.aud : []).some(item => item === this.issuer + projectId))) {
```
lahirumaramba: Can this be simplified with something like:

```ts
const scopedProjectId = `${this.issuer}${projectId}`;

} else if (!validator.isNonEmptyArray(payload.aud) || !payload.aud.includes(scopedProjectId)) {
....
```

src/utils/error.ts
```diff
@@ -1145,3 +1145,45 @@ const TOPIC_MGT_SERVER_TO_CLIENT_CODE: ServerToClientCode = {
...
+
+    /* tslint:disable:max-line-length */
+    // Set the prototype explicitly. See the following link for more details:
+    // https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work
+    /* tslint:enable:max-line-length */
+    (this as any).__proto__ = FirebasePnvError.prototype;
+  }
+}
+
+export class FpnvErrorCode {
```
lahirumaramba: Please use a type here to keep things consistent with the rest of the codebase. We have some work planned for error handling revamp so having it consistent here will reduce conflicts. Thanks!

For example see: https://github.com/firebase/firebase-admin-node/blob/179dab70b7ccc89185fc90e5c7a48ffecef60102/src/app-check/app-check-api-client-internal.ts#L228

https://github.com/firebase/firebase-admin-node/blob/179dab70b7ccc89185fc90e5c7a48ffecef60102/src/data-connect/data-connect-api-client-internal.ts#L657

src/fpnv/token-verifier.ts
```diff
@@ -0,0 +1,249 @@
...
+      return new FirebasePnvError(FpnvErrorCode.INVALID_ARGUMENT, errorMessage);
+    }
+    return new FirebasePnvError(FpnvErrorCode.INVALID_ARGUMENT, error.message);
+  }
+
+}
+
+export function createFPNTVerifier(app: App): FirebasePhoneNumberTokenVerifier {
+  return new FirebasePhoneNumberTokenVerifier(
+    CLIENT_CERT_URL,
```
lahirumaramba: Do we expect these to change in the future? If not, I don't think we would need a factory here. `FirebasePhoneNumberTokenVerifier` already has access to these consts. If you have used the auth token verification code as an example for this, we use a factory method there because we support two different token types (custom token and session tokens)

src/fpnv/fpnv.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export class Fpnv  {
+  private readonly app_: App;
+
+  protected readonly fpnvVerifier: FirebasePhoneNumberTokenVerifier;
+
+  constructor(app: App) {
+
+    this.app_ = app;
+    this.fpnvVerifier = createFPNTVerifier(app);
```
lahirumaramba: Can this be simplified to a `new FirebasePhoneNumberTokenVerifier(app)`?

etc/firebase-admin.fpnv.api.md
```diff
@@ -0,0 +1,64 @@
...
+    // (undocumented)
+    exp: number;
+    // (undocumented)
+    getPhoneNumber(): string;
+    // (undocumented)
+    iat: number;
+    // (undocumented)
+    iss: string;
+    // (undocumented)
+    sub: string;
```
lahirumaramba, line 26: Please add documentation to these types. For example see: https://github.com/firebase/firebase-admin-node/blob/179dab70b7ccc89185fc90e5c7a48ffecef60102/src/app-check/app-check-api.ts#L87

src/fpnv/fpnv.ts
lahirumaramba:
```suggestion
  public async verifyToken(fpnvJwt: string): Promise<FpnvToken> {
```

src/fpnv/fpnv.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * Returns the app associated with this Auth instance.
+   *
+   * @returns The app associated with this Auth instance.
+   */
+  get app(): App {
+    return this.app_;
+  }
+
```
lahirumaramba, line 54: Missing documentation here

src/fpnv/fpnv.ts
boikoa-gl: Tbh, i don't know about this. But,  what i noticed is that every module has such a method,

test/unit/fpnv/fpnv-api.spec.ts
```diff
@@ -0,0 +1,101 @@
+/*!
```
sam-gc: Let's just remove these tests and this file overall. All these tests do is test the mocked out object which doesn't provide any extra assurance. The parent file, `fpnv-api.ts` is just a bunch of interfaces and we can trust the TS compiler with type checks and so forth.

boikoa-gl: @sam-gc
Done

PR #3034, opened 2025-12-15
fix: Invalid provider UID handling in federated user lookup

demolaf:
Align federated UID validation with the rest of Auth requests. When [getAccountInfoByFederatedUid](https://github.com/firebase/firebase-admin-node/blob/255851bfd80684e5b75e1fa5e362ac8e19c6cb2c/src/auth/auth-api-request.ts#L1154) receives an empty provider UID, it now returns auth/invalid-provider-uid (previously incorrectly auth/invalid-provider-id), matching [validateProviderUserInfo](https://github.com/firebase/firebase-admin-node/blob/255851bfd80684e5b75e1fa5e362ac8e19c6cb2c/src/auth/auth-api-request.ts#L313). 

Unit test updated accordingly.

### DIFF
--- a/src/auth/auth-api-request.ts
+++ b/src/auth/auth-api-request.ts
@@ -1152,8 +1152,12 @@ export abstract class AbstractAuthRequestHandler {
   }
 
   public getAccountInfoByFederatedUid(providerId: string, rawId: string): Promise<object> {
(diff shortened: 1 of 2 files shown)


demolaf:
A similar PR was created here #2305

PR #3055, opened 2026-01-11
fix(data-connect): Include connector in DataConnect cache key

### LINKED ISSUE #3054: [Data Connect] getDataConnect() ignores connector field in cache key

Yes, this is a bug in the `firebase-admin/data-connect` module code.

  * Operating System version: macOS / Linux
  * Firebase SDK version: firebase-admin@13.x (all versions with Data Connect support)
  * Firebase Product: Data Connect
  * Node.js version: 24.x
  * NPM version: 11.x

#### Summary

The `DataConnectService.getDataConnect()` method caches `DataConnect` instances using only `location` and `serviceId` as the cache key, ignoring the `connector` field from `ConnectorConfig`.

This causes incorrect behavior when using multiple connectors with the same Data Connect service.

#### Steps to reproduce:

1. Create a Data Connect service with multiple connectors (e.g., `public`, `user`, `admin`)
2. Call `getDataConnect()` with different connector configs that share the same `location` and `serviceId`
3. Observe that both calls return the same `DataConnect` instance

#### Expected behavior:

Each unique combination of `location`, `serviceId`, and `connector` should return a distinct `DataConnect` instance.

#### Actual behavior:

The second call returns the cached instance from the first call, ignoring the different `connector` value.

#### Relevant Code:

```typescript
import { getDataConnect, ConnectorConfig } from 'firebase-admin/data-connect';

const publicConfig: ConnectorConfig = {
  location: 'us-west2',
  serviceId: 'my-service',
  connector: 'public',
};

const userConfig: ConnectorConfig = {
  location: 'us-west2',
  serviceId: 'my-service',
  connector: 'user',  // Different connector!
};

const publicDc = getDataConnect(publicConfig);
const userDc = getDataConnect(userConfig);

console.log(publicDc === userDc);  // true (BUG! Should be false)
console.log(userDc.connectorConfig.connector);  // 'public' (BUG! Should be 'user')
```

#### Impact:

When using multiple connectors (a common pattern for role-based access control), operations intended for one connector are incorrectly routed to another, causing "operation not found" errors.

#### Root cause:

In `src/data-connect/data-connect.ts`, line 39:

```typescript
const id = `${connectorConfig.location}-${connectorConfig.serviceId}`;
```

The cache key should also include the `connector` field:

```typescript
const id = `${connectorConfig.location}-${connectorConfig.serviceId}-${connectorConfig.connector ?? ''}`;
```

I have a fix ready with tests: #3055

itok01:
The `DataConnectService.getDataConnect()` method was caching `DataConnect` instances using only `location` and `serviceId` as the cache key, ignoring the `connector` field from `ConnectorConfig`.

This caused incorrect behavior when calling `getDataConnect()` with different connector configs that share the same `location` and `serviceId`:

```typescript
const dc1 = getDataConnect({ location: 'us-west2', serviceId: 'svc', connector: 'public' });
const dc2 = getDataConnect({ location: 'us-west2', serviceId: 'svc', connector: 'user' });
// dc1 === dc2 was true (BUG!)
```

Fixed by including `connector` in the cache key. Added 5 new unit tests for cache key behavior.

Fixes #3054.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,10 +15,18 @@
 name: Release Candidate
 
 on:
-  # Only run the workflow when a PR is updated or when a developer explicitly requests
-  # a build by sending a 'firebase_build' event.
+  # Run the workflow when:
+  # 1. A PR is created or updated (staging checks).
+  # 2. A commit is pushed to main (release publication).
+  # 3. A developer explicitly requests a build via 'firebase_build' event.
   pull_request:
-    types: [opened, synchronize, closed]
+    types: [opened, synchronize]
+
+  push:
+    branches:
+      - main
+    paths:
+      - 'package.json'
 
   repository_dispatch:
     types:
@@ -26,23 +34,19 @@ on:
 
 jobs:
   stage_release:
-    # To publish a release, merge the release PR with the label 'release:publish'.
+    # To publish a release, merge a PR with the title prefix '[chore] Release ' to main
+    # and ensure the squashed commit message also has the prefix.
     # To stage a release without publishing it, send a 'firebase_build' event or apply
     # the 'release:stage' label to a PR.
     if: github.event.action == 'firebase_build' ||
       contains(github.event.pull_request.labels.*.name, 'release:stage') ||
-      (github.event.pull_request.merged &&
-        contains(github.event.pull_request.labels.*.name, 'release:publish'))
+      (github.event_name == 'push' && startsWith(github.event.head_commit.message, '[chore] Release '))
 
     runs-on: ubuntu-latest
 
-    # When manually triggering the build, the requester can specify a target branch or a tag
-    # via the 'ref' client parameter.
     steps:
     - name: Checkout source for staging
       uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
-      with:
-        ref: ${{ github.event.client_payload.ref || github.ref }}
 
     - name: Set up Node.js
       uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
@@ -90,16 +94,13 @@ jobs:
   publish_release:
     needs: stage_release
 
-    # Check whether the release should be published. We publish only when the trigger PR is
-    #   1. merged
+    # Check whether the release should be published. We publish only when the trigger is
+    #   1. a push (merge)
     #   2. to the main branch
-    #   3. with the label 'release:publish', and
-    #   4. the title prefix '[chore] Release '.
-    if: |
-      github.event.pull_request.merged &&
+    #   3. and the commit message has the title prefix '[chore] Release '.
+    if: github.event_name == 'push' &&
       github.ref == 'refs/heads/main' &&
-      contains(github.event.pull_request.labels.*.name, 'release:publish') &&
-      startsWith(github.event.pull_request.title, '[chore] Release ')
+      startsWith(github.event.head_commit.message, '[chore] Release ')
 
     runs-on: ubuntu-latest
     environment: Release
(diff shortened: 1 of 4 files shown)


src/data-connect/data-connect.ts
```diff
@@ -36,7 +36,7 @@ export class DataConnectService {
   }
 
   getDataConnect(connectorConfig: ConnectorConfig): DataConnect {
-    const id = `${connectorConfig.location}-${connectorConfig.serviceId}`;
+    const id = `${connectorConfig.location}-${connectorConfig.serviceId}-${connectorConfig.connector ?? ''}`;
```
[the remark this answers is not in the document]
itok01: @gemini-code-assist Thank you for catching this! You're right that hyphen-separated keys could collide. I've updated the implementation to use `JSON.stringify` as suggested, and added the test case for ambiguous keys.

test/unit/data-connect/data-connect.spec.ts
```diff
@@ -224,3 +224,112 @@ describe('DataConnect', () => {
...
+      };
+
+      const dc1 = getDataConnect(configWithConnector);
+      const dc2 = getDataConnect(configWithoutConnector);
+
+      expect(dc1).to.not.equal(dc2);
+      expect(dc1.connectorConfig.connector).to.equal('my-connector');
+      expect(dc2.connectorConfig.connector).to.be.undefined;
+    });
+  });
```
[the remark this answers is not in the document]
itok01: @gemini-code-assist Good suggestion! I've added the test case for cache key collisions. All 172 tests pass now.

src/data-connect/data-connect.ts
```diff
@@ -36,7 +36,7 @@ export class DataConnectService {
   }
 
   getDataConnect(connectorConfig: ConnectorConfig): DataConnect {
-    const id = `${connectorConfig.location}-${connectorConfig.serviceId}`;
+    const id = JSON.stringify([connectorConfig.location, connectorConfig.serviceId, connectorConfig.connector ?? '']);
```
stephenarosaj: This successfully makes the cache keys unique per connector. `JSON.stringify()` is also what we use in Data Connect's Client JS SDK ([link](https://github.com/firebase/firebase-js-sdk/blob/5579b387ab697347ecd4a29ffed47c4cf808d544/packages/data-connect/src/api/DataConnect.ts#L286)), so that's definitely the approach we want.

I feel that instead of using an empty string when the `connector` field is `undefined`, we should keep it `undefined` and just call `JSON.stringify(connectorConfig)`, so that we're aligned with the Client JS SDK. This also makes it so that if a user explicitly sets their `connector` field to be an empty string, there's no collision with a config that has the field unset.

I also think that in both the Client JS SDK and here, we should be ordering our keys before we stringify. That way, two `ConnectorConfig` objects defined as `{ serviceId: "s", location: "l" }` and `{ location: "l", serviceId: "s" }` return the same `DataConnect` instance. Something like:

```typescript
const orderedConfig = Object.keys(connectorConfig)
	.sort()
	.reduce((obj, key) => {
		obj[key] = connectorConfig[key as keyof ConnectorConfig];
		return obj;
	}, {} as any);
const id = JSON.stringify(orderedConfig);
```

test/unit/data-connect/data-connect.spec.ts
```diff
@@ -224,3 +224,128 @@ describe('DataConnect', () => {
     });
   });
 });
+
+describe('getDataConnect() caching', () => {
```
stephenarosaj: could you rename this section just `'getDataConnect()'`?

test/unit/data-connect/data-connect.spec.ts
```diff
@@ -224,3 +224,128 @@ describe('DataConnect', () => {
...
+      const config: ConnectorConfig = {
+        location: 'us-west2',
+        serviceId: 'my-service',
+        connector: 'my-connector',
+      };
+
+      const dc1 = getDataConnect(config);
+      const dc2 = getDataConnect(config);
+
+      expect(dc1).to.equal(dc2);
```
stephenarosaj: we should probably use `.to.deep.equal` when checking `DataConnect` instances

test/unit/data-connect/data-connect.spec.ts
```diff
@@ -224,3 +224,128 @@ describe('DataConnect', () => {
...
+      const config2: ConnectorConfig = {
+        location: 'us-west2',
+        serviceId: 'my-service',
+        connector: 'connector-b',
+      };
+
+      const dc1 = getDataConnect(config1);
+      const dc2 = getDataConnect(config2);
+
+      expect(dc1).to.not.equal(dc2);
```
stephenarosaj: same as above - `.to.not.deep.equal`

test/unit/data-connect/data-connect.spec.ts
```diff
@@ -224,3 +224,128 @@ describe('DataConnect', () => {
...
+        connector: 'my-connector',
+      };
+
+      const dc1 = getDataConnect(config1);
+      const dc2 = getDataConnect(config2);
+
+      expect(dc1).to.not.equal(dc2);
+    });
+
+    it('should handle connector being undefined vs defined', () => {
```
stephenarosaj: could you make this more descriptive? something like `"should consider configs with connector undefined and defined as different"`

stephenarosaj: let's add a test here to make sure the difference between empty `connector: ""` and no defined value for `connector` is picked up on by the code

test/unit/data-connect/data-connect.spec.ts
```diff
@@ -224,3 +224,128 @@ describe('DataConnect', () => {
...
+
+      const dc1 = getDataConnect(configWithConnector);
+      const dc2 = getDataConnect(configWithoutConnector);
+
+      expect(dc1).to.not.equal(dc2);
+      expect(dc1.connectorConfig.connector).to.equal('my-connector');
+      expect(dc2.connectorConfig.connector).to.be.undefined;
+    });
+
+    it('should not have cache collisions with ambiguous keys', () => {
```
stephenarosaj, line 354: this is a good test. could you also add one for different ordering of `ConnectorConfig` keys?

itok01:
Thank you for the thorough review @stephenarosaj! I've addressed all your feedback:

**Cache key generation:**
- Updated to use `Object.keys().sort().reduce()` before `JSON.stringify()` to ensure consistent keys regardless of property order
- Kept `connector` as `undefined` instead of converting to empty string, aligning with the Client JS SDK approach

**Test updates:**
- Renamed describe to `'getDataConnect()'`
- Changed to `.to.deep.equal` / `.to.not.deep.equal` for DataConnect comparisons
- Renamed test to be more descriptive
- Added test for empty string vs undefined connector
- Added test for different ConnectorConfig key orderings

All 165 Data Connect tests pass. Let me know if there's anything else!

Issue #3060, opened 2026-02-03
UPDATE `fast-xml-parser` to fix `CVE-2026-25128`


thienscmon:
Audit Report: 
```
fast-xml-parser  4.3.6 - 5.3.3
Severity: high
fast-xml-parser has RangeError DoS Numeric Entities Bug - https://github.com/advisories/GHSA-37qj-frw5-hhjh
fix available via `npm audit fix --force`
Will install firebase-admin@12.7.0, which is a breaking change
node_modules/fast-xml-parser
  @google-cloud/storage  >=7.12.1
  Depends on vulnerable versions of fast-xml-parser
  node_modules/@google-cloud/storage
    firebase-admin  >=13.0.0
    Depends on vulnerable versions of @google-cloud/storage
    node_modules/firebase-admin
```

lahirumaramba:
Hey @thienscmon, thanks for submitting this issue. Unfortunately, we can't simply just upgrade the cloud storage dependency as the latest version of `@google-cloud/storage` bumped the typescript version, which we consider as a breaking change. We are currently looking into this.

One option is to go ahead and include the typescript upgrade in a minor version bump of Firebase Admin Node.js, which I think is a reasonable step to address a vulnerability. We will use this issue for any updates.

mnahkies:
FYI there's now a critical level CVE against the same library https://github.com/advisories/GHSA-m7jm-9gc2-mpf2 - looks like https://github.com/firebase/firebase-admin-node/pull/3071 would also resolve this, so would be great to see it merged/released

lahirumaramba:
Hey folks, this issue should now be resolved in `v13.7.0`. We are aware that this release includes TypeScript updates, which could potentially cause issues. Please open a new issue if you encounter any problems.

We also want to use this release to assess whether we can include TypeScript updates in future minor version bumps. I would like to hear your thoughts on this.

PR #3071, opened 2026-02-17
fix: upgrade @google-cloud/storage@7.19.0

lahirumaramba:
## Description
This PR bumps the optional dependency `@google-cloud/storage` to address a known security vulnerability in its underlying dependency, `fast-xml-parser`. As part of this update, TypeScript has been bumped in a minor version bump (breaking change exception) to support newer compilation requirements.

### Changes Included:
* **Security Fix**: Upgrades the optional `@google-cloud/storage` dependency from `^7.14.0` to `^7.19.0` to resolve a vulnerability in `fast-xml-parser`.  Fixes: #3060 
* **TypeScript Update:** 
  * Bumps the dev dependency `typescript` from `5.5.4` to `^5.7.3`. (this will result in future typescript upgrades in minor version bumps).
  * Refactors the `options.headers` non-null assertion (`!`) to an explicit type assertion (`as http.OutgoingHttpHeaders`) in `src/utils/api-request.ts` to satisfy stricter type-checking requirements introduced by the newer TypeScript version.
## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Security Fix / Dependency Update
- [x] (❗ exception) Breaking change (fix or feature that would cause existing functionality to not work as expected)
## Testing
- [x] All existing automated tests pass
- [x] Project successfully builds without TypeScript errors


src/utils/api-request.ts
```diff
@@ -828,7 +828,7 @@ class AsyncHttp2Call extends AsyncRequestCall {
     try {
       this.http2ConfigImpl = http2ConfigImpl;
       this.options = this.http2ConfigImpl.buildRequestOptions();
-      this.entity = this.http2ConfigImpl.buildEntity(this.options.headers!);
+      this.entity = this.http2ConfigImpl.buildEntity(this.options.headers as http.OutgoingHttpHeaders);
```
jonathanedey: I think this one should be `as http2.OutgoingHttpHeaders` but I also don't remember if both map to the same interface.

lahirumaramba: Replaced this with a typeguard instead

lahirumaramba:
Adding these notes on the `eslint` changes for future reference:

1. `@typescript-eslint/no-require-imports: 0`
Why it triggered: The repo still supports namespaces (deprecated) and uses require()

2. `@typescript-eslint/no-unused-expressions: 0`
This rule flags code like `expect(foo).to.be.true;`. In the test files (and possibly some source files using similar patterns), these look like unused property accesses to the linter, even though they trigger assertions in libraries like Chai. Disabling this globally is safe if the project mostly uses it for assertions. However, it does risk masking real bugs (e.g., typing `a == b` when you meant `a = b`).
In the future, we could enable this rule but exclude test files, or configure it to allow tagged template literals and specific patterns.
3. `@typescript-eslint/no-unused-vars: ['error', { 'varsIgnorePattern': '^_', ... }]`
The previous setting was just `['error']`. I relaxed it to allow variables starting with `_` to be unused.
It allows us to deliberately ignore variables (e.g., in function arguments) by prefixing them with an underscore, which is better than disabling the rule entirely for a line.

PR #3080, opened 2026-02-25
fix(auth): capture FIREBASE_AUTH_EMULATOR_HOST at init time

### LINKED ISSUE #2948: FIREBASE_AUTH_EMULATOR_HOST affects all initailized apps, unlike other emulator env var
When initializing Firebase Admin to use the emulator, it's required to set the env var `FIREBASE_AUTH_EMULATOR_HOST` prior to calling `initializeApp`.  This works fine, however, the way this value is used internally has a significant problem: The env var is consulted every time an API call is made. This makes it impossible to have two initialized apps in one process (for example, one pointing to the cloud project, and the other to the emulator.

The relevant code in this repo:

https://github.com/firebase/firebase-admin-node/blob/26b884f70300ef8e3213dfbbd72351ce7f68c695/src/auth/auth-api-request.ts#L2313-L2323

`useEmulator` is used whenever an API call is made, which affects all applications.  So, one application cannot simply set the env var while another unsets it.  What's especially confusing about this behavior is that the other env vars don't work like this at all.  I can set the var only when needed prior to init, then unset it later to initialize another app.

For example, this works for Firestore and Storage, but not for Auth:

```
// Initializing an app for remote cloud project access

export const remoteApp = initializeApp(options, "remote")
export const remoteAuth = getAuth(remoteApp)
export const remoteFirestore = getFirestore(remoteApp)
export const remoteStorage = getStorage(remoteApp)

// Initializing a second app for local emulator access

process.env.FIREBASE_AUTH_EMULATOR_HOST="127.0.0.1:9099"
process.env.FIRESTORE_EMULATOR_HOST = "localhost:8080"
process.env.FIREBASE_STORAGE_EMULATOR_HOST = "127.0.0.1:9199"
export const localApp = initializeApp(options, "local")
export const localAuth = getAuth(localApp)
export const localFirestore = getFirestore(localApp)
export const localStorage = getStorage(localApp)
delete process.env.FIREBASE_AUTH_EMULATOR_HOST
delete process.env.FIRESTORE_EMULATOR_HOST
delete process.env.FIREBASE_STORAGE_EMULATOR_HOST
```

Deleting `FIREBASE_AUTH_EMULATOR_HOST` here causes the app for the local emulator to instead use the remote service.  However, *not* deleting it causes the remote app to use the local emulator.  Again, Firestore and Storage don't have this problem.

As of now, I can't find a way to use emulator auth and remote auth in the same project.  It would be great if the admin SDK would take the value of `FIREBASE_AUTH_EMULATOR_HOST` just once at the time of init, store it, and use it later so that var can be deleted afterward.  Or even better, `initializeApp` could directly take emulator configs as input instead of relying solely on env vars that affect the entire process globally.

See also this comment in another issue: https://github.com/firebase/firebase-admin-node/issues/776#issuecomment-1781154086

daytonmux:
## Summary

Fixes #2948.

`FIREBASE_AUTH_EMULATOR_HOST` was read from `process.env` on every API call, which meant it affected all initialized apps globally. This made it impossible to have one app pointing to the emulator and another to the cloud project in the same process — unlike Firestore, Database, and Storage which already capture their emulator env vars at init time.

This change captures the value of `FIREBASE_AUTH_EMULATOR_HOST` once when the `Auth` service is initialized and stores it per-instance, so each app remembers its own emulator configuration independently.

### What changed

- **`auth-api-request.ts`**: `AuthResourceUrlBuilder`, `TenantAwareAuthResourceUrlBuilder`, and `AuthHttpClient` now receive the emulator host/flag as constructor parameters instead of reading from `process.env`. `AbstractAuthRequestHandler` captures `emulatorHost()` once at construction and threads it through to all URL builders and the HTTP client. `TenantAwareAuthRequestHandler` accepts an optional emulator host override so lazily-created tenant handlers inherit the frozen value.
- **`base-auth.ts`**: `BaseAuth` derives its `emulatorMode` from the request handler captured value instead of calling `useEmulator()` on every `verifyIdToken`/`verifySessionCookie`/`_verifyAuthBlockingToken` call. `createFirebaseTokenGenerator` accepts an optional `isEmulator` override for the same reason.
- **`tenant-manager.ts`**: `TenantManager` captures the emulator host at construction and passes it to `TenantAwareAuth` instances created lazily by `authForTenant()`, so tenant auth inherits the original emulator state.

## Test plan

- [x] All 6197 existing unit tests pass (`npm test`)
- [x] Lint passes
- [x] New tests in `auth-api-request.spec.ts`:
  - Emulator mode persists after env var is deleted post-init
  - Non-emulator mode persists when env var is set after init
- [x] New tests in `auth.spec.ts`:
  - `verifyIdToken()` retains emulator behavior after env var deletion
  - `verifySessionCookie()` retains emulator behavior after env var deletion
  - `authForTenant()` does not use emulator when env var set after `Auth` init
  - `authForTenant()` uses emulator when env var set before `Auth` init (and deleted after)

### DIFF
--- a/src/auth/auth-api-request.ts
+++ b/src/auth/auth-api-request.ts
@@ -127,14 +127,15 @@ class AuthResourceUrlBuilder {
   /**
    * The resource URL builder constructor.
    *
-   * @param projectId - The resource project ID.
+   * @param app - The app for this URL builder.
    * @param version - The endpoint API version.
+   * @param emulatorHost - Optional emulator host captured at init time.
    * @constructor
    */
-  constructor(protected app: App, protected version: string = 'v1') {
-    if (useEmulator()) {
+  constructor(protected app: App, protected version: string = 'v1', emHost?: string) {
+    if (emHost) {
       this.urlFormat = utils.formatString(FIREBASE_AUTH_EMULATOR_BASE_URL_FORMAT, {
-        host: emulatorHost()
+        host: emHost
       });
     } else {
       this.urlFormat = FIREBASE_AUTH_BASE_URL_FORMAT;
@@ -191,16 +192,17 @@ class TenantAwareAuthResourceUrlBuilder extends AuthResourceUrlBuilder {
   /**
    * The tenant aware resource URL builder constructor.
    *
-   * @param projectId - The resource project ID.
+   * @param app - The app for this URL builder.
    * @param version - The endpoint API version.
    * @param tenantId - The tenant ID.
+   * @param emHost - Optional emulator host captured at init time.
    * @constructor
    */
-  constructor(protected app: App, protected version: string, protected tenantId: string) {
-    super(app, version);
-    if (useEmulator()) {
+  constructor(protected app: App, protected version: string, protected tenantId: string, emHost?: string) {
+    super(app, version, emHost);
+    if (emHost) {
       this.urlFormat = utils.formatString(FIREBASE_AUTH_EMULATOR_TENANT_URL_FORMAT, {
-        host: emulatorHost()
+        host: emHost
       });
     } else {
       this.urlFormat = FIREBASE_AUTH_TENANT_URL_FORMAT;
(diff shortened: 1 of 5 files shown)


lahirumaramba (approved):

PR #3102, opened 2026-03-30
feat(core): Error Handling Revamp

jonathanedey:
This PR makes changes the usage of firebase errors and includes the following changes:
* Exposes all service-specific error classes, error code types and base `FirebaseError` class.
* Exposes the associated error codes types for those errors.
* Exposes `ErrorInfo` and `HttpResponse` classes for better error handing.
* Adds 2 new fields to ErrorInfo FirebaseErrors: `cause` and `HttpResponse`.
* Converts `AuthClientErrorCode`, `MessagingClientErrorCode`, `InstallationsClientErrorCode`, `InstanceIdClientErrorCode` from classes to constants.


src/utils/error.ts
lahirumaramba:
```suggestion
   * @param [serverError] The error's raw server response.
```

etc/firebase-admin.app-check.api.md
```diff
@@ -39,6 +42,14 @@ export interface DecodedAppCheckToken {
     sub: string;
 }
 
+// Warning: (ae-forgotten-export) The symbol "PrefixedFirebaseError" needs to be exported by the entry point index.d.ts
+//
```
lahirumaramba, line 46: Is this something we can address (for both `PrefixedFirebaseError` and `ErrorInfo`? I think it is okay to ignore this but I am just double checking :)

lahirumaramba (approved):
LGTM! Thanks! (added couple comments)

etc/firebase-admin.app-check.api.md
jonathanedey: For `ErrorInfo` it's the same scenario we have for `App`. Where even though we export it commonly, api documenter wants it to be exported again from other paths where it is used in the definition. So i don't think we have good options other than exposing it from multiple paths.

For `PrefixedFirebaseError` this is technically internal logic and not exposed at all even though it's a middle step between `FirebaseError` and the service specific errors. Not sure if there is a way to exclude it from the docs.

PR #3105, opened 2026-04-08
chore(deps): update @google-cloud/firestore to ^8.0.0

ouzkilic:
## Summary

Update `@google-cloud/firestore` optional dependency range from `^7.11.0` to
`^8.0.0` to eliminate the vulnerable `@tootallnate/once@2.0.0` transitive dependency.

## Problem

`@tootallnate/once` versions prior to 3.0.1 are vulnerable to Incorrect Control
Flow Scoping (CWE-705). When AbortSignal is used, Promises remain permanently
pending after abort, causing stalled requests and degraded availability.

The vulnerability chain through `@google-cloud/firestore@^7.11.0`:
google-gax@4 → retry-request@7 → teeny-request@9 → http-proxy-agent@5 → @tootallnate/once@2.0.0

## Fix

`@google-cloud/firestore@8.x` uses an updated dependency chain that no longer
includes `@tootallnate/once`:
google-gax@5 → retry-request@8 → teeny-request@10 → http-proxy-agent@7 (no @tootallnate/once)

Additionally, replaced the internal import of `DEFAULT_DATABASE_ID` from
`@google-cloud/firestore/build/src/path` with a local constant `'(default)'`,
as v8 no longer ships `.d.ts` for that internal module.

## Note on @google-cloud/storage

`@google-cloud/storage@7.19.0` (latest) still pulls in the same vulnerable chain
via `teeny-request@9`. This needs to be addressed separately upstream in the
googleapis/google-cloud-node repository.

## Testing

- [x] Existing test suite passes (6187 tests)
- [ ] `npm ls @tootallnate/once` shows reduced/no results after update

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "firebase-admin",
-  "version": "13.7.0",
+  "version": "13.8.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "firebase-admin",
-      "version": "13.7.0",
+      "version": "13.8.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@fastify/busboy": "^3.0.0",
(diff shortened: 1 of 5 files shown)


lahirumaramba:
Thanks for your contribution. This is a breaking dependency upgrade and should be included in a major version release. We will look into this.

lahirumaramba:
Thank you again for your contribution! This is now addressed in #3137

PR #3124, opened 2026-04-25
fix: Replace deprecated url.parse() with WHATWG URL (DEP0169)

### LINKED ISSUE #3118: Migrate `url.parse()` → WHATWG URL API to eliminate DEP0169 deprecation warnings in serverless logs

Yes — this is a bug report for code in `firebase-admin-node` itself. The issue is a deprecation warning emitted by internal utilities (`api-request.js`, `validator.js`) and inherited from transitive dependencies (`http-proxy-agent`, `teeny-request`, `@firebase/database-compat`, `faye-websocket`). It is not specific to Firestore.

  * **Operating System version:** Vercel serverless (Amazon Linux 2023 container) in production; reproducible locally on Windows 11, macOS, and Ubuntu.
  * **Firebase SDK version:** `firebase-admin@^12.1.0` (currently resolving to `12.7.x` per lockfile).
  * **Firebase Product:** multiple — `api-request.js` is shared across Auth, Firestore, Realtime Database, Storage. The warning fires regardless of which product is invoked.
  * **Node.js version:** 24.x (Vercel Fluid Compute runtime). Also reproducible on 20.x and 22.x.
  * **NPM version:** 10.x (bundled with Node 24).

On every cold start and on many subsequent operations, Node emits:

*****************
Node.js deprecated `url.parse()` because it has edge cases around non-standard URL structures (e.g., `//evil.com`, null-byte injection, embedded credentials in the authority) that can lead to security misinterpretations. The replacement is the WHATWG URL API (`new URL(...)` / `URL.canParse(...)`).

A grep of `node_modules/firebase-admin/` confirms direct `url.parse()` callers in:
  * `firebase-admin/lib/utils/api-request.js`
  * `firebase-admin/lib/utils/validator.js`
  * `@firebase/database-compat/dist/index.standalone.js`

Plus transitive callers `firebase-admin` pulls in:
  * `http-proxy-agent`
  * `@google-cloud/storage` → `teeny-request` → `agent-base`
  * `faye-websocket`

#### Why this matters for production users

  * **Log noise at scale:** the warning fires per cold start and on many REST-ish operations. On serverless platforms that aggregate logs (Vercel, Cloud Functions, Lambda) it drowns legitimate warnings and makes real issues (auth failures, quota errors, timeouts) harder to spot.
  * **Pressure to suppress warnings globally:** teams are tempted to use `NODE_OPTIONS=--no-deprecation` or `process.emit` overrides to silence the noise. Both patterns hide **unrelated legitimate warnings** and normalize an anti-pattern where future deprecations (potentially real security ones) get masked. Our team explicitly refused to do this, but the upstream noise incentivizes the bad pattern across the ecosystem.
  * **Signal integrity:** deprecation warnings are often the first signal of library drift in long-running production systems. Normalizing their suppression is itself a security-relevant concern.

#### Steps to reproduce:

  1. Install `firebase-admin@^12.1.0` in a Node 20+ environment.
  2. Initialize the admin SDK with service-account credentials and perform any single Firestore (or Auth, or Storage) operation.
  3. Observe the `DEP0169` warning on stderr per cold start and on subsequent REST-backed operations.

#### Relevant Code:

Minimal repro:

```js
// repro.mjs — run with: node --trace-deprecation repro.mjs
import admin from 'firebase-admin';

admin.initializeApp({
  credential: admin.credential.cert({
    projectId: process.env.FIREBASE_PROJECT_ID,
    clientEmail: process.env.FIREBASE_CLIENT_EMAIL,
    privateKey: process.env.FIREBASE_PRIVATE_KEY.replace(/\\n/g, '\n'),
  }),
});

const db = admin.firestore();
await db.collection('any').limit(1).get();
// → (node:*) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized...

```

// Before (legacy url.parse — deprecated)
const parsed = url.parse(input);
const hostname = parsed.hostname;
const pathname = parsed.pathname;

// After (WHATWG URL API)
if (!URL.canParse(input)) throw new Error(`Invalid URL: ${input}`);
const parsed = new URL(input);
const hostname = parsed.hostname;
const pathname = parsed.pathname;

---
URL.canParse() is available in Node 18.17+, well within the supported matrix for firebase-admin@12.

For the transitive chain (http-proxy-agent, teeny-request, agent-base, faye-websocket), the fix likely belongs in those repos, but pinning to versions that have already migrated (where available) would propagate the improvement here.

I'm happy to open a PR migrating the two direct firebase-admin callers (lib/utils/api-request.js, lib/utils/validator.js) if that would be useful — please let me know.

Related:

Node.js DEP0169: https://nodejs.org/api/deprecations.html#DEP0169
WHATWG URL API: https://nodejs.org/api/url.html#the-whatwg-url-api

kyungseopk1m:
Closes #3118.

`url.parse()` is deprecated (DEP0169) and emits warnings on every cold start in serverless runtimes. This swaps the two call sites in `BaseRequestConfigImpl.buildUrl()` for the WHATWG `URL` constructor, matching the pattern @lahirumaramba used for `validator.isURL()` in #3061.

### Changes

- `BaseRequestConfigImpl.buildUrl()` returns `URL` instead of `url.UrlWithStringQuery`. The method is `protected` and only used by two subclasses inside the same file, so this is not a public surface change.
- Both branches of `buildUrl()` now reuse a single `new URL(fullUrl)` instance instead of allocating one and re-parsing it via `url.parse()`.
- `parsed.path` is replaced with `${parsed.pathname}${parsed.search}` so the request path on the wire stays byte-for-byte identical.
- `parsed.port` is now `string` (always) rather than `string | null`. The existing `if (!port)` falsy check still picks up the empty case.

### Behavior

For valid inputs the request URL is byte-for-byte identical. For malformed inputs the WHATWG parser is stricter than `url.parse()`: depending on the input shape it may either canonicalize the input differently (`urlWithProtocol()` always prepends `https://`, so e.g. `//evil.com` becomes `https://evil.com/`) or throw a `TypeError`. SDK-internal callers normalize their URLs before invoking, so well-formed usage is unaffected.

### Out of scope

The transitive callers listed in #3118 (`http-proxy-agent`, `teeny-request`/`agent-base`, `faye-websocket`, `@firebase/database-compat`) live in separate repositories and need to be addressed there.

### Verification

- `npx tsc --noEmit -p .` — clean
- `npx eslint src/utils/api-request.ts` — 0 warnings
- `npx mocha test/unit/utils/api-request.spec.ts --require ts-node/register` — 199 passing
- `npx mocha test/unit/index.spec.ts --require ts-node/register` — 6231 passing

No spec changes: existing api-request tests already cover explicit non-default ports (`mockHost:8080`), POST with entity body, GET path matching, and nock interceptor URL parsing — all of which exercise the changed code paths.

### DIFF
--- a/src/utils/api-request.ts
+++ b/src/utils/api-request.ts
@@ -23,7 +23,6 @@ import * as validator from './validator';
 import http = require('http');
 import https = require('https');
 import http2 = require('http2');
-import url = require('url');
 import { EventEmitter } from 'events';
 import { Readable } from 'stream';
 import * as zlibmod from 'zlib';
@@ -991,23 +990,23 @@ class BaseRequestConfigImpl implements BaseRequestConfig {
     return data;
   }
 
-  protected buildUrl(): url.UrlWithStringQuery {
+  protected buildUrl(): URL {
     const fullUrl: string = this.urlWithProtocol();
+    const parsedUrl = new URL(fullUrl);
     if (!this.hasEntity() || this.isEntityEnclosingRequest()) {
-      return url.parse(fullUrl);
+      return parsedUrl;
     }
     if (!validator.isObject(this.data)) {
       throw new Error(`${this.method} requests cannot have a body`);
     }
-    // Parse URL and append data to query string.
-    const parsedUrl = new url.URL(fullUrl);
-    const dataObj = this.data as { [key: string]: string; };
+    // Append data to query string.
+    const dataObj = this.data as {[key: string]: string};
     for (const key in dataObj) {
       if (Object.prototype.hasOwnProperty.call(dataObj, key)) {
         parsedUrl.searchParams.append(key, dataObj[key]);
       }
     }
-    return url.parse(parsedUrl.toString());
+    return parsedUrl;
   }
 
   protected urlWithProtocol(): string {
@@ -1044,7 +1043,7 @@ class HttpRequestConfigImpl extends BaseRequestConfigImpl implements HttpRequest
   public buildRequestOptions(): https.RequestOptions {
     const parsed = this.buildUrl();
     const protocol = parsed.protocol;
-    let port: string | null = parsed.port;
+    let port: string = parsed.port;
     if (!port) {
       const isHttps = protocol === 'https:';
       port = isHttps ? '443' : '80';
@@ -1054,7 +1053,7 @@ class HttpRequestConfigImpl extends BaseRequestConfigImpl implements HttpRequest
       protocol,
       hostname: parsed.hostname,
       port,
-      path: parsed.path,
+      path: `${parsed.pathname}${parsed.search}`,
       method: this.method,
       agent: this.httpAgent,
       headers: Object.assign({}, this.headers),
@@ -1082,7 +1081,7 @@ class Http2RequestConfigImpl extends BaseRequestConfigImpl implements Http2Reque
 
     return {
       protocol,
-      path: parsed.path,
+      path: `${parsed.pathname}${parsed.search}`,
       method: this.method,
       headers: Object.assign({}, this.headers),
     };


lahirumaramba (approved):

PR #3126, opened 2026-04-29
feat(fcm): Improve HTTP/2 session error handling.

jonathanedey:
This PR refactors how FCM http2 session errors are handled making use of the new `cause` attribute on `FirebaseErrors` to allow failing of request at the stream level with context on session error cause.
* Defer session error reporting by instead of failing fast then waiting for the `BatchResponse` attached to the error to resolve, we now let streams fail naturally and attach session errors as `cause` to failed messages in `BatchResponse`.
* Add `AggregateError` support when multiple session errors are captured during a batch.
* Simplify `Http2SessionHandler` by removing the promise-based failure tracking (`invoke()`), since we no longer need to catch rejections in `sendEach()`.
* Remove internal `FirebaseMessagingSessionError`


lahirumaramba (approved):
Looks great! Thank you!

PR #3128, opened 2026-05-06
chore: Deprecate support for Node.js 20

lahirumaramba:
Node.js 20 has reached EOL. Deprecate support for Node.js 20


jonathanedey (approved):
LGTM, Thanks!

PR #3138, opened 2026-05-12
change: Drop support for Node.js 18 and 20 (v14)

lahirumaramba:
Drop support for Node.js 18 and 20
Add Node.js 26 to CIs
Upgrade firebase-tools to v15


jonathanedey (approved):
LGTM, Thanks!

PR #3142, opened 2026-05-13
fix(auth): skip project config name assertion when using the Auth emulator

### LINKED ISSUE #2461: Got 'INTERNAL ASSERT FAILED: Unable to get project config' error when use Emulator

  * For issues related to __the code in this repository__ file a Github issue.
  * If the issue pertains to Cloud Firestore, read the instructions in the "Firestore issue"
    template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

  * Operating System version: macOS 12.4
  * Firebase SDK version: v11.11.1
  * Firebase Product: auth (auth, database, storage, etc)
  * Node.js version: v18.18.2
  * NPM version: v.9.8.1

#### Steps to reproduce:

What happened? How can we make the problem occur?
This could be a description, log/console output, etc.

#### Relevant Code:

export FIREBASE_AUTH_EMULATOR_HOST="127.0.0.1:9099"

```js
import { initializeApp } from "firebase-admin/app";
import { getAuth } from "firebase-admin/auth";

const app = initializeApp({
  projectId: 'xxxxx',
});

const run = async () => {
    const projectConfigManager = getAuth(app).projectConfigManager();
    const projectConfig = await projectConfigManager.getProjectConfig();
};

run();
```

Got error when getProjectConfig with local emulator

```console
FirebaseAuthError: INTERNAL ASSERT FAILED: Unable to get project config
    at /xxx/node_modules/firebase-admin/lib/auth/auth-api-request.js:1658:15
    at /xxx/node_modules/firebase-admin/lib/auth/auth-api-request.js:1615:13
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async run {
  errorInfo: {
    code: 'auth/internal-error',
    message: 'INTERNAL ASSERT FAILED: Unable to get project config'
  },
  codePrefix: 'auth'
}
```

It seems to be missing the config name field in the response ?

kyungseopk1m:
Closes #2461.

Thanks @nerder and @mexican-jack for the same-symptom reports on `updateProjectConfig` and password policy updates — both go through the endpoints touched here.

## Summary

`getProjectConfig()` and `updateProjectConfig()` throw `INTERNAL ASSERT FAILED: Unable to get/update project config` when the Auth emulator is running. The emulator doesn't populate the resource `name` field on `/config` responses, and both endpoint validators require `name` to be a non-empty string.

This PR skips that assertion when `useEmulator()` is true so the call succeeds against the emulator. Production behavior is unchanged.

## Behavior change

- Emulator (`FIREBASE_AUTH_EMULATOR_HOST` set): `getProjectConfig()` / `updateProjectConfig()` resolve with whatever the emulator returns.
- Production: identical to before. A backend response missing `name` still throws `INTERNAL ASSERT FAILED`.

The Identity Toolkit `projects.getConfig` / `projects.updateConfig` REST contract treats `name` as an output-only resource name that production always populates — the emulator just doesn't implement it.

## Scope

The guard is limited to `GET_PROJECT_CONFIG` and `UPDATE_PROJECT_CONFIG` because those are the two endpoints with reproductions in the issue and its comments. Other validators in the same file (`GET_TENANT`, `UPDATE_TENANT`, OIDC/SAML config) follow the same pattern but have no emulator reports — leaving them out keeps the diff narrow.

Password policy updates (raised by @mexican-jack) go through `updateProjectConfig` internally, so this PR covers that case as well.

## Implementation notes

The guard uses the existing `useEmulator()` helper at `src/auth/auth-api-request.ts:2321` — the same one `AuthResourceUrlBuilder` (line 134), `TenantAwareAuthResourceUrlBuilder` (line 200), and `AuthHttpClient` (line 232) use to branch on the emulator. Same dynamic-read pattern as the rest of `main`.

## Relationship to #3080

#3080 is open in the same area and captures the emulator host at init time instead of reading it on each call. This PR uses the dynamic `useEmulator()` read because that's what `main` does today. If #3080 lands first, I'll rebase and switch the validator guard to the captured value so the emulator semantics stay aligned.

## Test coverage

`test/unit/auth/auth-api-request.spec.ts` adds a `getProjectConfig` and an `updateProjectConfig` describe block under the `supportsTenantManagement` branch. Each has three cases:

- normal production response with `name` populated — strict `callParams` match (URL, method, body, headers, timeout) for regression coverage
- production response missing `name` — still throws (covers the existing assertion)
- response missing `name` with `FIREBASE_AUTH_EMULATOR_HOST` set — resolves, with a strict match against the emulator URL `http://localhost:9099/identitytoolkit.googleapis.com/v2/projects/project_id/config[?updateMask=...]` and the `Bearer owner` emulator headers

`afterEach` clears `FIREBASE_AUTH_EMULATOR_HOST` so emulator state doesn't bleed into other specs.

## Verification

- `npm run lint` — 0 errors.
- `npx mocha test/unit/auth/auth-api-request.spec.ts --require ts-node/register` — 746 passing, 0 failing.
- `npm run test:unit` — 6246 passing, 0 failing.

Diff: 2 files changed, +136 insertions, -1 deletion (10 LOC in `src`, the rest in test).

### DIFF
--- a/src/auth/auth-api-request.ts
+++ b/src/auth/auth-api-request.ts
@@ -2035,6 +2035,11 @@ export abstract class AbstractAuthRequestHandler {
 /** Instantiates the getConfig endpoint settings. */
 const GET_PROJECT_CONFIG = new ApiSettings('/config', 'GET')
   .setResponseValidator((response: RequestResponse) => {
+    // The Auth emulator does not populate the resource `name` field on the
+    // project config response, so skip the assertion when running against it.
+    if (useEmulator()) {
+      return;
+    }
     const data = response.data;
     // Response should always contain at least the config name.
     if (!validator.isNonEmptyString(data?.name)) {
@@ -2049,6 +2054,11 @@ const GET_PROJECT_CONFIG = new ApiSettings('/config', 'GET')
 /** Instantiates the updateConfig endpoint settings. */
 const UPDATE_PROJECT_CONFIG = new ApiSettings('/config?updateMask={updateMask}', 'PATCH')
   .setResponseValidator((response: RequestResponse) => {
+    // The Auth emulator does not populate the resource `name` field on the
+    // project config response, so skip the assertion when running against it.
+    if (useEmulator()) {
+      return;
+    }
     const data = response.data;
     // Response should always contain at least the config name.
     if (!validator.isNonEmptyString(data?.name)) {
--- a/test/unit/auth/auth-api-request.spec.ts
+++ b/test/unit/auth/auth-api-request.spec.ts
@@ -46,7 +46,7 @@ import { getMetricsHeader, getSdkVersion } from '../../../src/utils/index';
 import {
   UserImportRecord, OIDCAuthProviderConfig, SAMLAuthProviderConfig, OIDCUpdateAuthProviderRequest,
   SAMLUpdateAuthProviderRequest, UserIdentifier, UpdateRequest, UpdateMultiFactorInfoRequest,
-  CreateTenantRequest, UpdateTenantRequest,
+  CreateTenantRequest, UpdateTenantRequest, UpdateProjectConfigRequest,
 } from '../../../src/auth/index';
 
 chai.should();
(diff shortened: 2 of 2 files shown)


lahirumaramba (approved):

PR #3144, opened 2026-05-13
chore: Add node: prefix to crypto imports

lahirumaramba:
Prefix native `crypto` module imports with `node:` across source files and tests. This explicitly distinguishes the built-in Node.js module from third-party packages, aligning with modern ESM and cross-runtime compatibility standards (e.g., Cloudflare Workers, Deno, Bun).


jonathanedey (approved):
LGTM, Thanks!

PR #3145, opened 2026-05-13
feat(fcm): Enable `fid` and deprecate `token` for Send API

yvonnep165:
This change deprecates `TokenMessage` interface, transitioning support to `FidMessage` interface. A new `FidMulticastMessage` interface containing only fids: string[] is added. An optional `fids` is added to the legacy `MulticastMessage` interface, which is deprecated, to support mixed transition batches.

The multicast helper function `sendEachForMulticast` is updated to support both target types concurrently up to a combined total of 500. And function overloads are implemented on `sendEachForMulticast` to support both tokens and fids payloads. 

Note for API doc: Hyper-reference link {@link Messaging.sendEachForMulticast}  doesn't work in api extraction since there are distinct method declarations due to function overloads. (throwing an `ae-unresolved-link` error) . So I change it to plain comment text.


src/messaging/messaging.ts
```diff
@@ -274,50 +275,82 @@ export class Messaging {
...
 
   /**
-   * Sends the given multicast message to all the FCM registration tokens
+   * Sends the given multicast message to all the FCM registration tokens or fids
    * specified in it.
    *
    * This method uses the {@link Messaging.sendEach} API under the hood to send the given
    * message to all the target recipients. The responses list obtained from the
-   * return value corresponds to the order of tokens in the `MulticastMessage`.
+   * return value corresponds to the order of tokens/fids in the `MulticastMessage`.
```
Doris-Ge, line 283: nit: Maybe clarify that `tokens` take precedence if both are provided.

Doris-Ge (approved):

src/messaging/messaging.ts
yvonnep165: Ok, added!

src/messaging/messaging-api.ts
```diff
@@ -40,18 +53,38 @@ export interface ConditionMessage extends BaseMessage {
 
 /**
  * Payload for the {@link Messaging.send} operation. The payload contains all the fields
- * in the BaseMessage type, and exactly one of token, topic or condition.
+ * in the BaseMessage type, and exactly one of fid, token, topic or condition.
```
lahirumaramba: Should we mention that token is deprecated here as well?

`in the BaseMessage type, and exactly one of fid, token (deprecated, use fid instead), topic or condition.`

lahirumaramba (approved):
LGTM! Thank you!

src/messaging/messaging-api.ts
yvonnep165: Good catch! Added, Thanks!

OrlandriaH-G (approved):

src/messaging/messaging.ts
```diff
@@ -274,50 +275,83 @@ export class Messaging {
...
+
+    if ('tokens' in copy && !validator.isArray(copy.tokens)) {
+      throw new FirebaseMessagingError(
+        MessagingClientErrorCode.INVALID_ARGUMENT, 'tokens must be a valid array');
+    }
+    if ('fids' in copy && !validator.isArray(copy.fids)) {
       throw new FirebaseMessagingError(
-        MessagingClientErrorCode.INVALID_ARGUMENT, 'tokens must be a non-empty array');
+        MessagingClientErrorCode.INVALID_ARGUMENT, 'fids must be a valid array');
     }
```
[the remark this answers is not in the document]
yvonnep165: Resolved

lahirumaramba (approved):
LGTM! Thanks!

PR #3147, opened 2026-05-14
ci: run tests on Node 26

gameroman:


lahirumaramba:
Hey @gameroman, thank you for your contribution! This change was addressed in https://github.com/firebase/firebase-admin-node/pull/3138 and will be included in the upcoming major release `v14`

PR #3153, opened 2026-05-19
chore: upgrade nock to v14 and fix api-request unit tests

lahirumaramba:
Upgrades `nock` to `v14.0.15` and updates `HttpClient` unit tests to align with `nock` v14's spec-compliant response handling and socket interception refactors.

- Bumps `nock` from `^13.0.0` to `^14.0.15` in `package.json`.
- Updates `mockRequestWithError` to normalize mock error inputs into real `Error` objects, preventing `nock` v14 from silently dropping request error events.
- In `nock` v13, a bug existed where `nock` incorrectly returned a response body for `HEAD` requests if `respData` was supplied to `.reply()`, forcing the old test to assert that `resp.data` was parsed. In `nock` v14, `nock` correctly conforms to the HTTP specification and ignores any reply body for `HEAD` requests, resulting in an empty raw response (`""`).
- Removed the obsolete mock response body (`respData`) from the `nock` reply. The test now correctly asserts that `resp.text === ''` and that accessing `resp.data` throws a missing data error.
- Refactors socket timeout tests to directly reply with an `ETIMEDOUT` error using `replyWithError(err)` instead of relying on mock socket delay events, which were not reliably triggering under `nock` v14.


jonathanedey (approved):
LGTM, Thanks!

PR #3159, opened 2026-05-22
fix: Refactor error logic to remove `PrefixedFirebaseError` as a intermediate class

jonathanedey:
This PR:
* Removes the internal `PrefixedFirebaseError` to directly extend service error classes from the exposed `FirebaseError`
* Moves `hasCode()` helper function to `FirebaseError` and encapsulate prefix logic in the service error classes
* Replaces instances of `FirebaseError` being used with a manual service error prefix.

This PR also creates specific error classes for Database, Firestore, Storage, and Machine Learning, replacing legacy instances where generic `FirebaseError` objects were thrown with hardcoded string prefixes.


lahirumaramba (approved):
LGTM! Thanks!

egilmorez (approved):
Thanks for the offline clarifications Jonathan!  LGTM

PR #3163, opened 2026-05-27
fix: resolve failing getApplicationDefault unit tests in gcloud env

lahirumaramba:
This PR resolves the unit test failures in `getApplicationDefault` and `isApplicationDefault` that occur on local machines where `gcloud` application default credentials exist. It also fixes a potential environment pollution issue in the test suite and cleans up unused ESLint directives.

### Root Cause of Failures
In a prior migration to `google-auth-library`, `getApplicationDefault()` was refactored to return an instance of `ApplicationDefaultCredential` (which dynamically delegates credential resolution under the hood) instead of manually parsing credential files and returning specific wrappers like `RefreshTokenCredential`.

However, two unit tests still asserted that the return value of `getApplicationDefault()` was directly an `instanceof RefreshTokenCredential`. On environments without local `gcloud` credentials (like CI), these tests were silently skipped. In environments where the developer has authenticated via `gcloud`, these tests executed and failed.

### Changes Implemented
1. **Modernized Spec Assertions:**
   - Updated the spec to check that `getApplicationDefault()` returns `ApplicationDefaultCredential`.
   - Resolved the underlying authenticating client using `(c as any).googleAuth.getClient()` and verified it resolves to an instance of `UserRefreshClient` (from `google-auth-library`), preserving the exact test coverage and intent.
   - Injected a mock `process.env.GOOGLE_CLOUD_PROJECT` inside these asynchronous test cases to prevent `google-auth-library` from hanging during GCE metadata service auto-discovery on local environments.

2. **Isolated test environment variables (`process.env`):**
   - Fixed the `beforeEach` and `afterEach` environment setup/teardown in `credential-internal.spec.ts` to shallow-copy and properly restore environment variables, preventing modified environment variables from polluting other test suites.

3. **ESLint directive cleanups:**
   - Removed unused `eslint-disable-next-line @typescript-eslint/no-var-requires` comments in `crypto-signer` files to resolve linting warnings.

### DIFF
--- a/src/utils/crypto-signer.ts
+++ b/src/utils/crypto-signer.ts
@@ -79,7 +79,7 @@ export class ServiceAccountSigner implements CryptoSigner {
    * @inheritDoc
    */
   public sign(buffer: Buffer): Promise<Buffer> {
-    const crypto = require('node:crypto'); // eslint-disable-line @typescript-eslint/no-var-requires
+    const crypto = require('node:crypto');
     const sign = crypto.createSign('RSA-SHA256');
     sign.update(buffer);
     return Promise.resolve(sign.sign(this.credential.privateKey));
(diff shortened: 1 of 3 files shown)


jonathanedey (approved):
LGTM, Thanks!

PR #2930, opened 2025-05-27
Corrected types for CreateRequest

clintonb:
Hey there! So you want to contribute to a Firebase SDK? 
Before you file this pull request, please read these guidelines:

### Discussion
https://github.com/firebase/firebase-admin-node/issues/2929

### Testing

  * Make sure all existing tests in the repository pass after your change.
  * If you fixed a bug or added a feature, add a new test to cover your code.

### API Changes

  * At this time we cannot accept changes that affect the public API.  If you'd like to help 
    us make Firebase APIs better, please propose your change in an issue so that we 
    can discuss it together.

Issue #3152, opened 2026-05-18
`auth.createUser({ displayName: null })` type-checks but throws `auth/invalid-display-name` at runtime


llamington:
### [READ] Step 1: Are you in the right place?

Yes — bug in the `auth` component of this repo (typing + runtime divergence between `createUser` and `updateUser`).

### [REQUIRED] Step 2: Describe your environment

  * Operating System version: macOS 15.2 (also reproduces on Linux per type definitions alone)
  * Firebase SDK version: `firebase-admin@13.6.0`
  * Firebase Product: auth
  * Node.js version: 22.x
  * NPM version: 10.x

### [REQUIRED] Step 3: Describe the problem

`CreateRequest` extends `UpdateRequest`, so `displayName`, `photoURL`, and `phoneNumber` are typed `string | null | undefined` on the create path. The runtime, however, only handles `null` on the **update** path — `updateExistingAccount` rewrites `displayName: null` into `deleteAttribute: ['DISPLAY_NAME']` before validation (`src/auth/auth-api-request.ts`, the `deletableParams` block). `createNewAccount` has no such transform, so `null` flows straight into `validateCreateEditRequest`, where:

```js
if (typeof request.displayName !== 'undefined' &&
    !validator.isString(request.displayName)) {
    throw new FirebaseAuthError(AuthClientErrorCode.INVALID_DISPLAY_NAME);
}
```

`isString` is `typeof v === 'string'`, so `null` fails. The type permits a value the runtime always rejects.

The same shape applies to `photoURL` and `phoneNumber` on `CreateRequest`.

#### Steps to reproduce:

1. Compile the snippet below — TypeScript accepts it.
2. Run against a real Firebase Auth project (or the emulator).
3. Observe `FirebaseAuthError { code: 'auth/invalid-display-name' }`.

#### Relevant Code:

```ts
import { getAuth } from 'firebase-admin/auth'

await getAuth().createUser({
  uid: 'repro-' + Date.now(),
  displayName: null, // typed as string | null, but runtime rejects null
})
// FirebaseAuthError: The displayName field must be a valid string.
//   code: 'auth/invalid-display-name'

// Contrast — this works because updateUser rewrites null into deleteAttribute:
await getAuth().updateUser('some-existing-uid', { displayName: null })
```

#### Expected

Either:
[...]

Issue #2929, opened 2025-05-27
Incorrect type for CreateRequest.phoneNumber


clintonb:
### [READ] Step 1: Are you in the right place?

  * For issues related to __the code in this repository__ file a Github issue.
  * If the issue pertains to Cloud Firestore, read the instructions in the "Firestore issue"
    template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Operating System version: _____
  * Firebase SDK version: 13.4.0
  * Firebase Product: auth
  * Node.js version: 22.11
  * NPM version: ??? (I use pnpm)

### [REQUIRED] Step 3: Describe the problem
Per https://github.com/firebase/firebase-admin-node/issues/1508, `phoneNumber` is not nullable for create requests.

#### Steps to reproduce:
1. Try to create a user with `phoneNumber = null`.
2. Get an error from the API

clintonb:
This is also somewhat related to https://github.com/firebase/firebase-admin-node/issues/2450.

lahirumaramba:
This is resolved in version 14.0.0. If you continue to encounter issues, please open a new issue with the relevant details.

PR #3165, opened 2026-05-28
fix(auth): Decouple `CreateRequest` interface from `UpdateRequest`

### LINKED ISSUE #2450: auth.createUser() incorrect types: cannot link provider on account creation

  * Operating System version: N/A
  * Firebase SDK version: 12.0.0
  * Firebase Product: auth
  * Node.js version: N/A
  * NPM version: N/A

The types for `auth.createUser()` (i.e. `CreateRequest`) inherit from `UpdateRequest` and therefore inherit types for `providerToLink` and `providersToUnlink`. However these parameters are _not_ supported (they are ignored) by the underlying [`POST /accounts`](https://cloud.google.com/identity-platform/docs/reference/rest/v1/projects/accounts) REST API.

https://github.com/firebase/firebase-admin-node/blob/910460d213f276df955442119ac90083e2836c6b/src/auth/auth-config.ts#L234

This is a misrepresentation. Based on the REST API, provider credentials cannot be linked while creating an account, only on a subsequent `POST /accounts:update` request.

While we're at it the `UpdateRequest` type does not include custom claims, which _is_ supported by the `/accounts:update` REST API. See #1882.

jonathanedey:
By inheriting `UpdateRequest`, `CreateRequest` unintentionally exposes type definitions and optional nullable contracts designed strictly for user update requests. This results in misleading and incompatible type definitions.
 
This PR explicitly defines independent property fields on the `CreateRequest` interface rather than extending `UpdateRequest`. This resolves misleading and incompatible type inheritance between user creation and user update workflows by:
* Removing `providerToLink` and `providersToUnlink` property fields from `CreateRequest`.
* Removing `null` type from the `displayName`, `photoURL`, and `phoneNumber` property fields in `CreateRequest`.

Fixes: #2450, #2929, #3152
Supersedes and closes: #2930

### DIFF
--- a/etc/firebase-admin.auth.api.md
+++ b/etc/firebase-admin.auth.api.md
@@ -232,8 +232,15 @@ export interface CreatePhoneMultiFactorInfoRequest extends BaseCreateMultiFactor
 }
 
 // @public
-export interface CreateRequest extends UpdateRequest {
+export interface CreateRequest {
+    disabled?: boolean;
+    displayName?: string;
+    email?: string;
+    emailVerified?: boolean;
     multiFactor?: MultiFactorCreateSettings;
+    password?: string;
+    phoneNumber?: string;
+    photoURL?: string;
     uid?: string;
 }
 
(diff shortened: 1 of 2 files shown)


lahirumaramba (approved):
Thanks! LGTM!

PR #3166, opened 2026-06-01
change(instance-id): Remove deprecated Instance ID service

jonathanedey:
This pull request completely removes the deprecated Instance ID (IID) service and all its associated exports, namespaces, and test suites. Developers should use the Firebase Installations service (`installations()`) instead for all instance management.


lahirumaramba (approved):
Thanks! LGTM!
Heads up: There will be some merge conflicts once the namespace PR is in

PR #3172, opened 2026-06-08
chore: Upgrade compilation target and lib to ES2021

lahirumaramba:
Upgrade TypeScript compilation target and standard library configurations from es2020 to es2021. It also cleans up a now-redundant runtime workaround for `AggregateError` in the `Messaging` module.


jonathanedey (approved):
LGTM, Thanks!

PR #3180, opened 2026-06-15
fix(dataconnect): fix deserialization of emulator gRPC error codes into detailed errors

stephenarosaj:
## Description
✨ Added gRPC status code mapping to ensure error responses from the local Data Connect emulator are correctly deserialized. Before these changes, when using the emulator, errors are improperly deserialized into `DataConnectError`s with `unknown-error` instead of ones with detailed `code` and `message` fields.

Verified by running the integration tests against the emulator (and production) before and after the changes. What was broken is now fixed!

## Changes
- Added `GRPC_STATUS_CODE_TO_STRING` mapping dictionary
- Updated `toFirebaseError` to handle root-level gRPC error responses
- Added FDC emulator debug log files to `.gitignore`

## Testing
- Currently, integration tests fail against the emulator. Now they pass. 
- Added unit test verifying deserialization of transcoded gRPC numeric error codes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ docgen/markdown/
 
 # Dataconnect integration test artifacts should not be checked in
 test/integration/dataconnect/dataconnect/.dataconnect
+test/integration/dataconnect/dataconnect-debug.log
+test/integration/dataconnect/firebase-debug.log
+test/integration/dataconnect/pglite-debug.log
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,4 @@ docgen/markdown/
 
 # Dataconnect integration test artifacts should not be checked in
 test/integration/dataconnect/dataconnect/.dataconnect
-test/integration/dataconnect/dataconnect-debug.log
-test/integration/dataconnect/firebase-debug.log
-test/integration/dataconnect/pglite-debug.log
+test/integration/dataconnect/*.log
--- a/src/data-connect/data-connect-api-client-internal.ts
+++ b/src/data-connect/data-connect-api-client-internal.ts
@@ -25,7 +25,7 @@ import {
   FirebaseDataConnectError,
   DataConnectErrorCode,
   DATA_CONNECT_ERROR_CODE_MAPPING,
-  GRPC_STATUS_CODE_TO_STRING
+  EMULATOR_GRPC_STATUS_CODE_TO_STRING
 } from './error';
 import * as utils from '../utils/index';
 import * as validator from '../utils/validator';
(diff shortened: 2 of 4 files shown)


src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -409,10 +414,17 @@ export class DataConnectApiClient {
       });
     }
 
-    const error: ServerError = (response.data as ErrorResponse).error || {};
+    const data = response.data as any;
+    const error: ServerError = (validator.isNonNullObject(data) && 'error' in data) ? data.error : data || {};
```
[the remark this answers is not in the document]
stephenarosaj: Done.

dconeybe:
ultranit: looks like there may be some mismatched backticks in the PR description (see attached screenshot)

image

dconeybe (approved):

jonathanedey (approved):
LGTM, Thanks!

.gitignore
```diff
@@ -22,4 +22,7 @@ firebase-admin-*.tgz
 docgen/markdown/
 
 # Dataconnect integration test artifacts should not be checked in
-test/integration/dataconnect/dataconnect/.dataconnect
\ No newline at end of file
+test/integration/dataconnect/dataconnect/.dataconnect
+test/integration/dataconnect/dataconnect-debug.log
+test/integration/dataconnect/firebase-debug.log
+test/integration/dataconnect/pglite-debug.log
```
lahirumaramba: Would it make sense to ignore all log files?
`test/integration/dataconnect/*.log`?

src/data-connect/error.ts
```diff
@@ -69,3 +69,26 @@ export class FirebaseDataConnectError extends FirebaseError {
     });
   }
 }
+
+/**
+ * Mappings from gRPC status codes to their string equivalents.
+ * @internal
+ */
+export const GRPC_STATUS_CODE_TO_STRING: Record<number, string> = {
```
lahirumaramba: This adds a new mapping structure for gRPC status codes. Since this deviates from the existing pattern used elsewhere in the codebase, I want to ensure this is an intentional design choice before we proceed. @jonathanedey, do you have any thoughts on this approach?

.gitignore
stephenarosaj: Good point - that's a better approach than cherry picking. Done.

src/data-connect/error.ts
stephenarosaj: For extra context - our prod backend also uses these, but before they're sent out they are converted from error codes to statuses - but the emulator doesn't do this, so that translation is required in the client (that's why our integration tests all passed until I ran them against the emulators instead of prod)

jonathanedey: From my understanding these gRPC codes are emulator specific since they are not being converted locally like they would by the live backend. The only thing that might be worth considering is expanding the list of server error codes we map to include all the server codes we don't support in `DATA_CONNECT_ERROR_CODE_MAPPING` if these are also coming from the backend.

lahirumaramba: Thanks for clarifying!  Would it help to include that in the comments or change the name of the const to reflect the use-case (`EMULATOR_GRPC_STATUS_CODE_TO_STRING`)?

stephenarosaj: Definitely - will add that in now. Done.

lahirumaramba (approved):
Thank you!

Issue #3041, opened 2025-12-19
[Data Connect] Admin SDK bulk insert/insertMany fails to serialize enum values correctly (GraphQL coercion error)


samtsangbiz:
### [READ] Step 1: Are you in the right place?

  * For issues related to __the code in this repository__ file a Github issue.
  * If the issue pertains to Cloud Firestore, read the instructions in the "Firestore issue"
    template.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Operating System version: **macOS**
  * Firebase SDK version: **firebase-admin 13.6.0**
  * Firebase Product: **Data Connect**
  * Node.js version: **24.12**
  * NPM version: **11.6.2**

#### Steps to reproduce:

1. Use the minimal GraphQL schema with an enum and a table referencing it.

2. Initialize the Firebase Admin app and get a Data Connect instance (optionally connect to the Data Connect emulator).

3. Attempt to insert a row using `dc.insert("ticker", { symbol: "TEST", status: "PRIMARY" })`.

The operation fails with a GraphQL validation error indicating that the enum value is being sent as a quoted string instead of an unquoted enum literal.

Tested only in the Data Connect emulator.

#### Relevant Code:

Minimal schema (`dataconnect/schema/schema.graphql`):

```
enum TickerStatus {
  PRIMARY
  ACTIVE
  HISTORICAL
}

type Ticker @table {
  symbol: String!
  status: TickerStatus!
}
```

Minimal Node.js script to reproduce:

```javascript
import { getDataConnect } from "firebase-admin/data-connect";
import admin from "firebase-admin";

admin.initializeApp();

// For emulator testing:
// process.env.DATA_CONNECT_EMULATOR_HOST = "127.0.0.1:9399";

const dc = getDataConnect({
  location: "us-west2", // or your location
  serviceId: "your-service-id",
});

async function testInsert() {
  try {
    await dc.insert("ticker", {
      symbol: "TEST",
      status: "PRIMARY", // string matching enum literal
    });
    console.log("Success");
  } catch (error) {
    console.error("Error:", error);
  }
}

[...]

PR #3182, opened 2026-06-17
fix(dataconnect): Refactor CRUD helpers to use GraphQL variables and @allow directive

stephenarosaj:
## Description
✨ Refactored Data Connect CRUD operations to execute parameterized GraphQL mutations with query variables and `@allow` directives. This fixes the bug described by Issue #3041 where enums would not be serialized properly by the `[insert,upsert](many)` APIs. 

To verify this fix and harden integration tests, refactored tests so that before checking for equality between the expected and actual input query strings, they are normalized.

## Changes
- Removed `objectToString` inline GraphQL serializer
- Added `getTableNames` and `getFieldsString` helpers to handle 
- Refactored CRUD operations (`insert`, `insertMany`, `upsert`, `upsertMany`) to use GraphQL variables and `@allow` directives, and to not use duplicated code
- Added Data Connect emulator test instructions in `CONTRIBUTING.md`
- Updated ignored emulator debug logs in `.gitignore` so that no matter where they show up, they are properly ignored. This makes it so that even when running integration tests from the root of the SDK repo they are not tracked.

## Testing
- Added `expectNormalizedExecuteGraphqlCall` test helper
- Updated CRUD unit tests to verify parameterized mutations and variables
- Added unit tests for enum field serialization and `@allow` directives, including complex nested input types, and arrays of complex nested input types with different selection sets on each item
- Updated string serialization tests to check equality between normalized expected and actual query strings


src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -438,55 +438,46 @@ export class DataConnectApiClient {
...
-    if (typeof data === 'undefined') {
-      return 'null';
+  private getTableNames(tableName: string): { capitalized: string; formatted: string } {
+    if (!tableName || tableName.length === 0) {
+      return { capitalized: tableName, formatted: tableName };
     }
+    const capitalized = tableName.charAt(0).toUpperCase() + tableName.slice(1);
+    const formatted = tableName.charAt(0).toLowerCase() + tableName.slice(1);
+    return { capitalized, formatted };
+  }
```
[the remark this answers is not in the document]
stephenarosaj: This is an Admin API - the entire point of this is to execute arbitrary mutation GQL (and we even have an execute arbitrary query/mutation GQL API). Also, nothing has changed here anyways - this is just a refactor.

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -438,55 +438,46 @@ export class DataConnectApiClient {
...
-    return String(data);
+  /**
+   * Extracts all defined property keys from an object as a space-separated string.
+   * Used to build the `@allow(fields: ...)` mutation directive for single operations.
+   */
+  private getObjectKeys(data: Record<string, unknown> | object): string {
+    return Object.keys(data)
+      .filter(key => (data as Record<string, unknown>)[key] !== undefined)
+      .join(' ');
   }
```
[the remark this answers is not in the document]
stephenarosaj: Again, there's no need for concern here. This is not a change to the API, and we already literally hand users an "execute your own arbitrary GQL" API.

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -438,55 +438,46 @@ export class DataConnectApiClient {
...
+        Object.keys(record).forEach(key => {
+          if (record[key] !== undefined) {
+            allKeys.add(key);
+          }
+        });
+      }
     }
-    return tableName;
+    return Array.from(allKeys).join(' ');
   }
```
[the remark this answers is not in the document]
stephenarosaj: Not a concern.

stephenarosaj:
From offline review by @mtr002:

By default (admin usage), we build a flat `@allow` fields list containing only the top-level database columns. To explicitly perform nested relational mutations, the developer is required to explicitly set the allowed fields in the `@allow(fields: "...")` to lock down the nested relationship structure 

Also, since this feature is for admin usage, a limit of 100 seems very low, especially given that there is no way to override it. We currently have an upper limit of 10k set

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -636,20 +611,30 @@ export class DataConnectApiClient {
...
     try {
-      tableName = this.formatTableName(tableName);
-      const gqlDataString = this.objectToString(data);
-      const mutation = `mutation { ${tableName}_upsertMany(data: ${gqlDataString}) }`;
-      // Use internal executeGraphql
-      return this.executeGraphql<GraphQlResponse, Variables>(mutation).catch(this.handleBulkImportErrors);
+      const { capitalized, camelCase } = this.getTableNames(tableName);
+      const keys = this.getFieldsString(data);
+      const mutation =
+        `mutation($data: [${capitalized}_Data!]! @allow(fields: "${keys}")) {
```
mtr002: While we correctly validate that the input array doesn't exceed 10,000 items, the generated mutation doesn't specify maxCount inside the @allow directive.

By default, the fsql backend defaults maxCount to 100 if it is omitted from the @allow directive. Therefore, if a user attempts to bulk import 150 items, it will pass the validation but fail at the backend with a payload limit error.

We should hardcode maxCount: 10000 inside the @allow directive

stephenarosaj: Good catch - but quick question - this should only added to the `@allow` directive for bulk inserts (like array variables) right??

mtr002: Yes! MaxCount is only allowed on array variables

mtr002 (commented):
Thank you for the fixes. LGTM after fixing the test assertions

CONTRIBUTING.md
```diff
@@ -150,9 +150,20 @@ And then:
...
 
-Currently, only the Auth, Database, and Firestore test suites work. Some test
-cases will be automatically skipped due to lack of emulator support. The section
-below covers how to run the full test suite against an actual Firebase project.
+Currently, only the Auth, Database, and Firestore test suites work. Some test cases
+will be automatically skipped due to lack of emulator support. 
+
+You can also run the Data Connect test suite against the emulators using the same command,
+but you must run only the dataconnect tests, using a config file specific to Data Connect 
+emulator testing:
```
lahirumaramba, line 158: WDYT of rephrasing this to:
```
Note: You can also run the Data Connect test suite against the emulators using the same command.
However, you must isolate the Data Connect tests by using a configuration file specific to Data Connect
emulator testing.
```

src/data-connect/data-connect-api-client-internal.ts
```diff
@@ -689,3 +631,55 @@ interface ServerError {
...
+ *
+ * @internal
+ */
+export function getFieldsString(data: unknown): string {
+  const root: FieldNode = { children: new Map() };
+  mergeFieldsIntoTree(data, root);
+  return serializeFieldNode(root);
+}
+
+function mergeFieldsIntoTree(data: unknown, node: FieldNode): void {
```
lahirumaramba: This might be an overkill so I will let you decide. Could there be circular references in the data object? If that's the case this could go into an infinite loop.  If you that is a risk, we could prevent it by adding a tracker `visited = new Set<unknown>()`.

```js
mergeFieldsIntoTree(data: unknown, node: FieldNode, visited: Set<unknown>)
....
if (!validator.isNonNullObject(data) || data instanceof Date) {
    return;
  }
if (visited.has(data)) {
    return;
  }
  visited.add(data);
....
```

lahirumaramba (approved):
Thanks! LGTM!

CONTRIBUTING.md
stephenarosaj: That's much more clear. Will update, thx for the suggestion :P

src/data-connect/data-connect-api-client-internal.ts
stephenarosaj: there shouldn't be as it's not valid input for any of our APIs - but also this is user input, so it would be nice to guard against this in the case they're using it improperly. i'll update to include the check!

stephenarosaj: Implemented cycle detection using ancestor visited Set tracking in mergeFieldsIntoTree recursion. It will throw when a cycle is encountered, while correctly allowing duplicate sibling references.

PR #3200, opened 2026-06-26
fix zizmor findings

morganchen12:


lahirumaramba (approved):

.github/workflows/release.yml
```diff
@@ -110,6 +115,8 @@ jobs:
     steps:
     - name: Checkout source for publish
       uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        persist-credentials: false
```
lahirumaramba: Would this break the GITHUB_TOKEN below in `Create release tag`?

morganchen12: Good catch, fixed this

PR #3206, opened 2026-07-08
chore(docs): add custom `@excludeFromDocs` tag to filter generated docs

jonathanedey:
This adds a new docs tag `@excludeFromDocs` used to filter out public APIs that are intended to be hidden in public docs and removes the need for manually omitting these on each release.


lahirumaramba (approved):
Thanks! LGTM!

PR #3210, opened 2026-07-09
feat(functions): Implement TaskQueue Scopes

inlined:
## Description
This PR introduces **Task Queue Scopes** to the Firebase Admin SDK functions module. It decouples the legacy backward-compatible string-to-scope migration logic from the stateless client and isolates it inside the API wrapper layer (`TaskQueue`).

### Key Changes
1. **API Wrapper Retry & Upgrade (`TaskQueue`)**:
   - `TaskQueue.enqueue()` and `delete()` now capture `404` errors for legacy string parameters (`'extensionOrKit'` scope).
   - On a `404` error, they automatically retry using a temporary `kit` scope.
   - If the retry succeeds, the wrapper mutates the stateful scope to `'kit'` and logs a legacy targeting warning.
2. **Immediate Self-Targeting Warning**:
   - Matches targeting parameters against `process.env.EXT_INSTANCE_ID` and `process.env.KIT_INSTANCE_ID` at construction time.
   - If self-targeting is detected, it logs an immediate deprecation/performance warning and mutates the scope immediately to `'extension'` or `'kit'`, bypassing unnecessary `404` fallback call roundtrips.
3. Design style: chose to make the migration logic part of the API type and not the API internals type to more obviously collect what is transitionary/migration logic that can be removed later
4. **Unit Tests**:
   - Added unit tests verifying `scope` resolves to global if no environment variables are set.
   - Added unit tests verifying self-targeting warnings and direct namespaced resolution for both extensions and kits.


src/functions/functions.ts
```diff
@@ -53,11 +53,35 @@ export class Functions {
...
-    return new TaskQueue(functionName, this.client, extensionId);
+  public taskQueue<Args = Record<string, any>>(
+    functionName: string,
+    deprecatedExtensionId: string
+  ): TaskQueue<Args>;
+  public taskQueue<Args = Record<string, any>>(
+    functionName: string,
+    extensionIdOrScope: string | FunctionScope = { scope: 'current' }
+  ): TaskQueue<Args> {
+    const scope: FunctionScope = typeof extensionIdOrScope === 'string'
```
lahirumaramba: If we use `FunctionScope` will that get rid of the `as any` casts?

src/functions/functions-api-client-internal.ts
```diff
@@ -199,6 +212,54 @@ export class FunctionsApiClient {
     }
   }
 
+  private resolveResourceId(
+    resourceId: string,
+    scope?: InternalFunctionScope
+  ): { resourceId: string; extensionOrKitId?: string } {
+    if (typeof scope === 'undefined') {
+      return this.resolveResourceId(resourceId, { scope: 'current' });
```
lahirumaramba: Should we use a default parameter for `scope` and skip this recursive call?

```ts
private resolveResourceId (
  resourceId : string ,
  scope : InternalFunctionScope = { scope: 'current' }
)...
```

src/functions/functions-api-client-internal.ts
```diff
@@ -66,9 +74,13 @@ export class FunctionsApiClient {
    *
    * @param id - The ID of the task to delete.
    * @param functionName - The function name of the queue.
-   * @param extensionId - Optional canonical ID of the extension.
+   * @param scope - Optional FunctionScope configuration.
```
lahirumaramba: Could we get a TW review on these new docstrings, please?

lahirumaramba (approved):
Thank you for submitting this!
LGTM! Added a couple of comments.

src/functions/functions-api-client-internal.ts
egilmorez: This looks like a literal, `FunctionScope`.  Though, would just "function scope" work as a common noun/general term?

See bottom note here: https://docs.google.com/presentation/d/1Ezb4sp4XD6ItHISccnME-EVH_oJj59R5Naq0IskLrAo/edit?slide=id.g1cd7eabb964_0_10#slide=id.g1cd7eabb964_0_10

src/functions/functions-api-client-internal.ts
```diff
@@ -132,10 +140,15 @@ export class FunctionsApiClient {
    *
    * @param data - The data payload of the task.
    * @param functionName - The functionName of the queue.
-   * @param extensionId - Optional canonical ID of the extension.
+   * @param scope - Optional FunctionScope configuration.
```
egilmorez: Same as above; literal?

src/functions/functions-api-client-internal.ts
```diff
@@ -132,10 +140,15 @@ export class FunctionsApiClient {
    *
    * @param data - The data payload of the task.
    * @param functionName - The functionName of the queue.
```
egilmorez: Should be either `functionName` or just "function name."

Better yet, more words might make it all more clear.  Is it the name of the function associated with the queue, or some even better verb than "associate?"

src/functions/functions.ts
```diff
@@ -53,11 +53,35 @@ export class Functions {
    *     `{functionName}`
    * 
    * @param functionName - The name of the function.
-   * @param extensionId - Optional Firebase extension ID.
+   * @param scope - Optional FunctionScope configuration. Only needed if targeting a scope other than the current one.
```
egilmorez: I think this should be `FunctionScope` with code font for the literal.

src/functions/functions.ts
```diff
@@ -66,15 +90,45 @@ export class Functions {
  */
 export class TaskQueue<Args = Record<string, any>> {
 
+  private readonly scope: InternalFunctionScope;
+
   /**
    * @param functionName - The name of the function.
    * @param client - The `FunctionsApiClient` instance.
-   * @param extensionId - Optional canonical ID of the extension.
+   * @param scope - Optional FunctionScope configuration. Only needed if targeting a scope other than the current one.
```
egilmorez: Same as above RE literal.

src/functions/functions.ts
```diff
@@ -66,15 +90,45 @@ export class Functions {
...
+   */
+  constructor(
+    functionName: string,
+    client: FunctionsApiClient,
+    deprecatedExtensionId: string
+  );
+  /**
+   * @param functionName - The name of the function.
+   * @param client - The `FunctionsApiClient` instance.
+   * @param extensionIdOrScope - Optional canonical ID of the extension or FunctionScope.
```
egilmorez: Probably `FunctionScope`

test/unit/functions/functions.spec.ts
```diff
@@ -164,6 +164,60 @@ describe('TaskQueue', () => {
         return new TaskQueue(FUNCTION_NAME, mockClient);
       }).not.to.throw();
     });
+
+    it('should not throw given a valid scope object: current', () => {
```
egilmorez, line 168: LMK if these "it('...." comments are public.  I think they are not.

egilmorez (commented):
Thanks, I do have some nits!  :)

src/functions/functions-api-client-internal.ts
inlined: We're choosing to use `FunctionScope` as a literal because 'a function scope' is an overloaded programming term that can mean something different for lower-case F functions (the lifecycle of variables within a function).

src/functions/functions-api-client-internal.ts
inlined: Resolved. Wrapped `FunctionScope` in backticks.

src/functions/functions-api-client-internal.ts
inlined: Resolved. Updated the description to 'The name of the function associated with the queue.'

src/functions/functions.ts
inlined: Resolved. Wrapped `FunctionScope` in backticks.

src/functions/functions.ts
inlined: Resolved. Wrapped `FunctionScope` in backticks.

src/functions/functions.ts
inlined: Resolved. Wrapped `FunctionScope` in backticks.

test/unit/functions/functions.spec.ts
inlined: These are Mocha unit test blocks in a spec file, so they are not part of the public API documentation.

inlined: They are not

Issue #3059, opened 2026-01-29
Race condition when trying to use more than one cloud tasks instance


amfc:
Environment is irrelevant, but I'm using ubuntu 25.04 with firebase admin 13.6.0, cloud tasks, node 20 and npm 10.8.2.

When using multiple Firebase apps (e.g., one for emulator, one for production), there's a race condition because CLOUD_TASKS_EMULATOR_HOST is read on every operation instead of being cached in the constructor like Auth and Storage.

This makes it impossible to safely use an emulator app and a production app in the same process as when you connect to an emulator all calls go there.

It's trivial to fix, I'll provide a patch.

amfc:
[firebase-admin-cloud-tasks-emulator-fix.patch](https://github.com/user-attachments/files/24933133/firebase-admin-cloud-tasks-emulator-fix.patch)

patch src/functions/functions-api-client-internal.ts < firebase-admin-cloud-tasks-emulator-fix.patch

jonathanedey:
This is resolved in version 14.0.0. If you continue to encounter issues, please open a new issue with the relevant details.

PR #1686, opened 2022-05-04
Added support for calling Eventarc emulator

apascal07:

PR #3223, opened 2026-07-29
fix(eventarc): resolve the emulator host when the client is constructed

kyungseopk1m:
### Discussion

Follows the same fix as #3167 (issue #3059), which moved `CLOUD_TASKS_EMULATOR_HOST` in `FunctionsApiClient` to construction time.

`EventarcApiClient` reads `process.env.CLOUD_EVENTARC_EMULATOR_HOST` on every publish rather than when the client is built, so the routing of an existing channel depends on whatever the environment happens to hold at publish time. Create a production client, set the variable, create an emulator client, then unset it, and both publish to production.

This moves the lookup into the constructor and stores it on a readonly field, matching `FunctionsApiClient` and `AbstractAuthRequestHandler`, which both capture their emulator host up front.

### Behavior change

Code that sets `CLOUD_EVENTARC_EMULATOR_HOST` after obtaining a channel will now keep hitting production. Nothing in the SDK, its tests, or #1686 documents that ordering as supported, and Cloud Tasks and Auth already behave this way, but it is user visible so it is worth calling out.

### Scope

This keeps `??` instead of the `.trim() || undefined` normalization used in `FunctionsApiClient`. There the value is a truthiness flag, checked in four places, so a whitespace only value has to collapse to `undefined`. Here it is a plain URL prefix used once, and normalizing it would change what `''` and `'   '` do today, which is unrelated to the caching fix.

### Testing

Two tests in `test/unit/eventarc/eventarc.spec.ts`, built the same way as the multi client test #3167 added. One covers the reported direction, an emulator channel created while the variable is set keeps using the emulator after it is cleared. The other covers the opposite direction, a production channel created before the variable is set keeps using production. Reverting the source fails both. The full unit suite passes at 6066.

### DIFF
--- a/src/eventarc/eventarc-client-internal.ts
+++ b/src/eventarc/eventarc-client-internal.ts
@@ -42,6 +42,7 @@ const DEFAULT_CHANNEL_REGION = 'us-central1';
  */
 export class EventarcApiClient {
   private readonly httpClient: HttpClient;
+  private readonly eventarcHost: string;
   private projectId?: string;
   private readonly resolvedChannelName: Promise<string>;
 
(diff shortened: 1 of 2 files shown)


kyungseopk1m:
Done in 5e5a1db.

